Issue #1, opened 2015-07-21
Header files installation


cejkato2:
I tried cmake, make and make install but only shared object (with links) was installed.
Is it planed to install header files into standard path? I expect so...
Should I create a patch?

rkrejci:
It is planned, but right now the API is far from being complete or even usable by a stranger. That's why we don't install any header file right now.

PR #2, opened 2015-07-21
installation of header files

cejkato2:
Ok, it was not so difficult... consider merging this patch please, it is related to question #1 
Thanks


rkrejci:
merged into the `devel` branch

PR #3, opened 2015-08-13
fix some corner cases in building the documentation

lperkov:


rkrejci:
merged into the `devel` branch

PR #7, opened 2015-11-10
parser_yin BUGFIX when adding the same module, after free the loading module,  the 'ctx->models.list[i]' is change, not exist module yet

RoyWorkerLuo:


rkrejci:
I don't understand the purpose of this change. What is wrong with the current code?

RoyWorkerLuo:
sorry, I am wrong

rkrejci:
that's ok :)

PR #8, opened 2015-11-24
resolve BUG_FIX when type is leafref, and the 'leafref' is defined in 'typedef' node, the path 'stype->info.lref.path' is null, should use ' stype->der->type.info.lref.path'

RoyWorkerLuo:


michalvasko:
Hi,
your commit fixed just a small bug that was a result of unfinished type resolution. Hopefully, it is all fixed in the devel branch now, please verify that your problem was fixed as well.

Regards,
Michal

PR #9, opened 2015-11-28
resolve BUG_FIX when type is leafref, and the 'leafref' is defined in…

RoyWorkerLuo:
… 'typedef' node, the path 'stype->info.lref.path' is null, should use ' stype->der->type.info.lref.path'


rkrejci:
duplicates issue #8.

PR #30, opened 2016-03-02
tests FEATURE simple unit test of libyang.h functions

mislavn:

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.6)
 
-set(data_tests test_data_initialization test_leafref_remove test_instid_remove)
+set(data_tests test_data_initialization test_leafref_remove test_instid_remove test_libyang)
 set(schema_yin_tests test_ietf test_augment test_print_transform)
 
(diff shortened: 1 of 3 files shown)


mislavn:
Hi Radek,

I am not sure where to put the used xml file in the unit tests. I used a model from schema/yin/files/a.yin but the used xml file is in data/files/a.xml with other xml files and the unit test is in data/test_libyang.c. Do i need to move the a.xml file or the location of the unit test test_libyang.c?

We will expand this unit test, with the rest of the function from libyang.h and add new unit tests for other functions from headers tree_schema.h, tree_data.h, xml.h and dict.h.

Best regards,
Mislav Novakovic

rkrejci:
Hi Mislav,
first, thank you for the tests!

So far, the tests/ directory contained perf/ subdir for performance testing (something different from cmocka tests), schema/ for testing schema parsing (e.g. checking schema correctness) and data/ for testing data validation. At least that was the idea about the directory structure. It seems to me that your tests are something different, so maybe there should be added api/ subdir? Anyway, I'm open to suggestions about the tests, so actually anything (even changes of the current structure) that makes some sense is welcome.

Regards,
Radek

PR #34, opened 2016-03-15
tests FEATURE simple unit test of tree_data.h functions

mislavn:

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.6)
 
-set(api_tests test_libyang test_tree_schema)
+set(api_tests test_libyang test_tree_schema test_tree_data)
 set(data_tests test_data_initialization test_leafref_remove test_instid_remove test_keys)
 set(schema_yin_tests test_ietf test_augment test_print_transform)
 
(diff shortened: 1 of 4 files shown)


mislavn:
Hi Radek,

I did not write unit tests for lyd_new_anyxml, lyd_output_new, lyd_output_new_leaf and lyd_output_new_anyxml because I had trouble writing a xml with rpc's output and input section. Please could you provide with an xml file with rpc section so I can finish the unit tests for tree_data.h header file.

Best regards,
Mislav Novakovic

michalvasko:
Hi Mislav,
as for `lyd_new_anyxml`, its prototype is similar to `lyd_new_leaf` and is used the same way. Just replace leaf value with an arbitrary XML tree, which is technically anything at all (if there are no tags in it, it is interpreted as an XML text node).

The problem with RPC output is that when used in NETCONF, the output parameters are directly under a <rpc-reply> element, which makes standard functions unusable.  I included sample RPC input and output, hope it helps. Also, RPC output is definitely not meant to be parsed from a file (although the options is there), server should create it on its own.

[rpc.zip](https://github.com/CESNET/libyang/files/174143/rpc.zip)

Regards,
Michal

rkrejci:
Hi Mislav, why did you closed this pull request? #35 seems to be something else, so do you plan some other pull request with the functionality from #34?

mislavn:
Hi Radek, I did not yet finished the unit tests for tree_data.h header so I sent another one in the mean time. I will make a new pull request later today for tree_data.h.

PR #37, opened 2016-03-18
tests BUGFIX fix error messages in test_tree_schema.c file

mislavn:


rkrejci:
The file descriptor is non-negative integer, so I'd prefer to initiate `fd` to `-1` and when closing the file descriptor, use condition `fd > -1`. I know that in this case the code works fine (it actually would work correctly even with `fd > 2`), but it is good practise to use `-1` for value of invalid file descriptor.

PR #43, opened 2016-03-23
tests FEATURE simple unit test of tree_data.h functions

mislavn:


michalvasko:
Thank you, we will appreciate if you add tests for the rest of the functions, I just wanted to add tests for `lyd_path_new()` to this file (so you do not need to add tests for that one).

Regards,
Michal

PR #64, opened 2016-05-17
swig javascript: add default value to ly_ctx_destroy in swig for easi…

mislavn:
…er use


mislavn:
There is no easy way to do NULL value in node where it excepts a function pointer and with the
patch I can do easily:

yang.ly_ctx_destroy(ctx)

in node.

PR #81, opened 2016-06-28
hi, I found that 'edit-config' and 'get-config' will fail when there is a 'leafref' leaf, and I try to fix it

fanchanghu:
1, The 'resolve' param in function 'lyp_parse_value' seems to work like 'not resolve' , and two callers(xml_get_value, json_get_value) used it as 'not resolve' too, and the other callers used it as 'resolve'. So I try to fix it.
2, When 'resolve' the value of a leaf, only the value type is set to the leaf, and this may cause 'edit-config' failed for leaf of 'leafref'. and I also try to fix it in this request.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -803,7 +803,7 @@ lyp_parse_value_(struct lyd_node_leaf_list *node, struct lys_type *stype, int re
 
         if (parse_int(dec, __INT64_C(-9223372036854775807) - __INT64_C(1), __INT64_C(9223372036854775807), 10, &num,
                       (struct lyd_node *)node)
-                || validate_length_range(2, 0, 0, ((long double)num) / type->info.dec64.div, stype,
+                || validate_length_range(2, 0, 0, ((long double)num)/(1 << type->info.dec64.dig), stype,
                                          node->value_str, (struct lyd_node *)node)) {
             return EXIT_FAILURE;
         }
--- a/src/parser.c
+++ b/src/parser.c
@@ -655,9 +655,9 @@ lyp_parse_value_(struct lyd_node_leaf_list *node, struct lys_type *stype, int re
     int c, i, j, d;
     int found;
 
-    assert(node && (node->value_type == stype->base));
+    assert(node && ((node->value_type & LY_DATA_TYPE_MASK) == stype->base));
 
-    switch (node->value_type) {
+    switch (node->value_type & LY_DATA_TYPE_MASK) {
     case LY_TYPE_BINARY:
         if (validate_length_range(0, (node->value_str ? strlen(node->value_str) : 0), 0, 0, stype,
                                   node->value_str, (struct lyd_node *)node)) {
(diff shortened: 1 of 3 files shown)


michalvasko:
Hi,
what exactly does not work for you, do you have a use-case?

Regards,
Michal

fanchanghu:
```

const char *mytest_data = "<a xmlns=\"http://www.mytest.com/\"><x>aa</x><y>aa</y></a>\n";
const char *mytest_schema = "module mytest{\
  namespace \"http://www.mytest.com/\";\
  prefix \"mytest\";\
\
  container a{\
    leaf x{\
      type string;\
    }\
    leaf y{\
      type leafref{\
        path \"/a/x\";\
      }\
    }\
  }\
}";

static void
test_mytest_for_leafref(void **state)
{
    int rc;
    struct ly_set *set;
    struct lyd_node *node;
    struct lyd_node_leaf_list *yleaf;
    const struct lys_module *schema;

    char *yang_folder = TESTS_DIR"/api/files";

    ctx = ly_ctx_new(yang_folder);
    assert_non_null(ctx);

    schema = lys_parse_mem(ctx, mytest_schema, LYS_YANG);
    assert_non_null(schema);

    root = lyd_parse_mem(ctx, mytest_data, LYD_XML, LYD_OPT_EDIT);
    assert_non_null(root);

    set = lyd_get_node(root, "/mytest:a/y");
    assert_non_null(set);
    assert_int_equal(1, set->number);

    node = set->set.d[0];
    assert_non_null(node);
    yleaf = (struct lyd_node_leaf_list *)node;
    assert_string_equal("aa", yleaf->value_str);

    assert_int_equal(LY_TYPE_STRING | LY_TYPE_LEAFREF_UNRES, yleaf->value_type);
    assert_string_equal("aa", yleaf->value.string); //it failed here and netopeer2 need this value.

    return 0;
}
```
In fact,  It failed when I do 'edit-config' in netopeer2 with above data and schema.

michalvasko:
Hi,
if you look at the `parseroptions` documentation, you will see that it is mentioned that for `LYD_OPT_EDIT` leafrefs are not resolved. In your example it can be resolved, but what if *leaf x* is already in the datastore? Then the edit-config is valid, but the leafref cannot be resolved. That is the reason why we never resolve it for edit-config configs.

I thought you may have used netopeer2, but in that case I am not sure we can do anything about leafref support now because sysrepo does not seem to support it yet.

Regards,
Michal

fanchanghu:
Hi,
Thanks for your reply, I think I have known why edit-config failed for leafref. But I still has 2 questions.
First, as you see in my test case, this line 
`assert_int_equal(LY_TYPE_STRING | LY_TYPE_LEAFREF_UNRES, yleaf->value_type);` passed, so I think it means leafref is resolved, but only the `yleaf->value_type` is setted, the value of  `yleaf->value.string` is null.
Second, if the  value of `resolve` param in function 'lyp_parse_value_' is 1, it works like 'not resolve', and if the value is 0, it works like 'resolve'.

fanchanghu:
Hi, I found the problem, the line does not passed when I reset the code. maybe this problem need be resolved in netopeer or sysrepo.

fanchanghu:
Thank you, I think I know it now. `yleaf->value_type=LY_TYPE_STRING | LY_TYPE_LEAFREF_UNRES` make me think that the value should be a string. ^_^

michalvasko:
Well, it may be a bit confusing in this case, but generally, `value_type` describes what is stored in `value_str`. Leafref target is `LY_TYPE_STRING`, so naturally `value_str` will be a string as well. But leafref stores the pointer to the leaf as it's normal `value`. You always know what to examine in the union `value` based on the leaf type.

Regards,
Michal

PR #105, opened 2016-08-19
tests: tests for internal xml and tree

antonio-paunovic:
Signed-off-by: Antonio Paunovic <antonio.paunovic@sartura.hr>


antonio-paunovic:
Hi Radek,

this pull requests contains unit tests for two internal headers: tree_internal and xml_internal.
We have checked for memory leaks and 'still reachables'.
There are none, but there are some invalid reads if you want to check that out.
Two dubious functions are test_lys_submodule_parse and test_lys_submodule_read, we weren't able to track down issues in those.

rkrejci:
Hi,
I'm not sure that having standalone tests of internal functions is a good idea. At least they must be written with a deep knowledge of the code - the mentioned invalid reads are caused by invalid use of the functions in the tests. 

In `test_lys_submodule_` tests you read the module and then its submodule, but the submodule is loaded internally when you load the main module, so loading submodule again will cause inconsistencies in the module structures which later lead to double frees/invalid reads. Another issue - assigning the `"freedom"` string at `test_tree_internal.c:287` is very dangerous since libyang uses internal dictionary to handle strings. And there will probably be other issues.

I cannot accept this pull request.

PR #113, opened 2016-09-07
Internal unit tests devel

antonio-paunovic:
Hi Radek,
there are two tests added. Testing lyd_list_equal() is tested for simple cases which don't seem to change.
If the function is considered finished overall more complex cases (unique comparison) can be tested. 

Regards,
Antonio

### DIFF
--- a/tests_internal/CMakeLists.txt
+++ b/tests_internal/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.6)
 
-set(internal_tests test_lys_find_grouping_up)
+set(internal_tests
+  test_lys_find_grouping_up
+  test_lys_node_switch)
 
 include_directories(SYSTEM ${CMOCKA_INCLUDE_DIR})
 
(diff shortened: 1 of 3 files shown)


rkrejci:
There was a bug in `lyd_list_equal()` (see f045d1c), because lists without keys (which is possible in status data) should be considered as different - RFC 7950, sec. 4.2.2.4:
```
A list defines a sequence of list entries.  Each entry is like a
container and is uniquely identified by the values of its key leafs
if it has any key leafs defined.
```
Therefore the test now fails at `test_lyd_list_equal.c:121` because it expects that list instances without keys are supposed to be marked as duplication.

mislavn:
Hi Radek,

we updated the license with a new git commit.

Regards,
Mislav

rkrejci:
ok, but the `test_lyd_list_equal_no_action` is still wrong.

mislavn:
Hi, I will fix it now.

mislavn:
Hi Radek, the test now passed.

PR #123, opened 2016-09-13
modified cmake file for internal tests

mislavn:


rkrejci:
I have probably the better solution in 5708fe6 by using cmake's object library.

PR #127, opened 2016-09-20
add test for function lyd_check_mandatory_tree()

antonio-paunovic:
Hi Radek, 

here is test for mentioned function.
I noticed that in tree_internal.h documentation for that function has four parameters but function takes three. 

Regards,
Antonio

Signed-off-by: Antonio Paunovic <antonio.paunovic@sartura.hr>
Signed-off-by: Mislav Novakovic <mislav.novakovicda@sartura.hr>

### DIFF
--- a/tests_internal/CMakeLists.txt
+++ b/tests_internal/CMakeLists.txt
@@ -4,9 +4,12 @@ set(internal_tests
   test_lys_find_grouping_up
   test_lys_node_switch
   test_lyd_list_equal
-  test_lyd_attr_parent)
+  test_lyd_attr_parent
+  test_lyd_check_mandatory_tree)
 
 include_directories(SYSTEM ${CMOCKA_INCLUDE_DIR})
+
+configure_file("${PROJECT_SOURCE_DIR}/tests_internal/config.h.in" "${PROJECT_SOURCE_DIR}/tests_internal/config.h" ESCAPE_QUOTES @ONLY)
 
 foreach(test_name IN LISTS internal_tests)
     add_executable(${test_name} $<TARGET_OBJECTS:yangobj> tree_internal/${test_name}.c)
--- a/tests_internal/config.h.in
+++ b/tests_internal/config.h.in
@@ -1,0 +1,18 @@
+/**
+ * @file config.h
+ * @author Antonio Paunovic <antonio.paunovic@sartura.hr>
+ * @brief cmocka tests configuration header.
+ *
+ * Copyright (C) 2016 Deutsche Telekom AG.
+ *
+ * Author: Antonio Paunovic <antonio.paunovic@sartura.hr>
+ *
+ * This source code is licensed under BSD 3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://opensource.org/licenses/BSD-3-Clause
+*/
+
+#define TESTS_DIR "@CMAKE_SOURCE_DIR@/tests_internal"
+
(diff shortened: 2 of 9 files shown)


rkrejci:
Hi, few notes:
- I have updated the description of the function, thanks
- the manipulation of `options` in the test does not make sense. The `~0 & (A | B)` is the same as `A | B`, so there is no reason for setting `options` to `~0`. Also setting multiple `LYD_OPT_TYPEMASK` is not a standard way of using `options`.
- creating context with search path `.` does not make sense - working directory is always searched
- NOSIBLING case - it actually depends on modules in context. If there is a module with top-level mandatory node, it should fail. I'm not sure what is this test case testing.
- setting nodetype to `~LYS_NOTIF` is nonsense, there is always one value
- the test does not test checking presence of mandatory nodes at all, but it is the main functionality of the function, it just checks processing of the `options` parameter, but that's too little

antonio-paunovic:
Hi, to clarify:

- I will remove option initialization ~0 and send just single options
- context with "." path was made more as a placeholder with no deeper meaning. Maybe that's confusing, so we'll make dummy context by hand
- nodetype `~LYS_NOTIF` is just testing for a case when nodetype isn't `LYS_NOTIF`
- this test case is about checking expected result depending on `options`. Real mandatory node checking seems to be done in `lyd_check_mandatory_subtree()` and  `lyd_check_mandatory_date()` static functions which aren't in internal headers. I think mocking them with wrapper functions would make much more sense but I had  troubles getting them to work properly from CMake.

To conclude,
- thanks for comments, I'll try to do more descriptive and meaningful test because it obviously isn't enough so a the moment

rkrejci:
Hi,
- ad "." context search path - you can create context using `lyd_ctx_new()`, but use `NULL` parameter, which is completely fine when you do not need searchpath.
- ad static functions - I'm not sure that it is a good idea to test each (static) function, even not available via internal header, in these tests. I would try to focus only to the functions mentioned in the header first.

Actually, looking into [codecov](https://codecov.io/gh/CESNET/libyang/src/887941ae6cb3746445045cc29717af8bfe8abe08/src/tree_data.c#L77), the mentioned functions seem already to be well covered by other (non-internal) tests. That was actually also my feeling at the beginning when we were talking about internal functions tests - some (maybe most) of them seems to be already covered.

antonio-paunovic:
Hi, 
- I wasn't thinking about testing static functions but to avoid them. I thought that they should be mocked like in [this article](https://lwn.net/Articles/558106/). While it looks convenient, I can't find straight-forward way to include the wrapping in current CMake configuration. 
Testing internals is kind of complicated to do right and I guess those internals will be more often changed in comparison to API. Actually, if you have suggestions about which functions from internal headers would be most useful to test and help development cycle, that would be great.

rkrejci:
The note about changes in internal API is good point :( Anyway, the only suggestion is probably to check [codecov](https://codecov.io/gh/CESNET/libyang) to see which functions are not covered well.

antonio-paunovic:
Hi Radek,
we've made total changes for this test because it was untestable using last approach.
This seems simple enough and is based more on RFC paragraph than on code so should be more useful even with future changes.

Regards, 
Antonio

antonio-paunovic:
Hi,
just to note that there is a typo in tests/config.h.in file at first line: 
`@file: conig.h`

rkrejci:
I really like how the test is now written. However, I believe that it just duplicates what is already tested in `tests/data/test_mandatory.c`. If not, I would prefer to extend/fix the `test_mandatory.c` instead of adding a new (very similar) test somewhere else. Your code is definitely better because it is better described (commented) what is wrong in specific check. But again, I would prefer to improve the original test instead of adding a new one.

Also note that there is the `conformance_tests` branch we would like to merge during this week which includes even more tests covering YANG 1.0.

Maybe we should again discuss the idea of internal tests.

PR #134, opened 2016-09-27
rename 'new' to '_new' to make it can be used in c++

fanchanghu:
Hi, 
When I use libyang in my c++ project, there is an error when compiling. 'new' is the keyword of c++, so I hope to change the 'new' to '_new' in the header file.


rkrejci:
I have renamed it to 'repl'

PR #135, opened 2016-09-28
lyd_parse_mem parse rpc's output json will return NULL

yanjunxi:
[issue.txt](https://github.com/CESNET/libyang/files/497088/issue.txt)

### DIFF
--- a/src/parser_json.c
+++ b/src/parser_json.c
@@ -1215,7 +1215,7 @@ lyd_parse_json(struct ly_ctx *ctx, const struct lys_node *parent, const char *da
 
     ly_errno = LY_SUCCESS;
 
-    if (!ctx || !data || (options & LYD_OPT_RPCREPLY)) {
+    if (!ctx || !data) {
         LOGERR(LY_EINVAL, "%s: Invalid parameter.", __func__);
         return NULL;
     }


yanjunxi:
solve lyd_parse_mem can't parse rpc's output data by json format.

michalvasko:
Hi,
the condition was added there on purpose because JSON parser is not currently able to parse RPC reply the way it should (XML parser is). Do you really need to parse RPC replies in JSON? If so, then I will make it fully working, I just thought it is not needed for now.

Regards,
Michal

rkrejci:
This is supposed to have very low priority since JSON-encoded RPCs do not have real-world use case, do they?

yanjunxi:
We do need transmit data between different module using json.
And I find it works well after deleting the code as I did in my case.

Thanks.

rkrejci:
If you just need to transmit data, you do not need RPCs, do you? You can just dump the data tree.

yanjunxi:
We need receive RPC command from one module(A), transmit the RPC's input information to another module(B), then receive the RPC's output info from module(B) and reply to module(A).
Between our module and module B must use json format data because of some other reasons.

Thanks.

Issue #119, opened 2016-09-09
failed to load module with decimal64 type


fanchanghu:
code
```
int main(void)
{
    struct ly_ctx *ly_ctx = NULL;
    const struct lys_module *module;

    ly_ctx = ly_ctx_new(".");

    module = lys_parse_path(ly_ctx, "d.yang", LYS_IN_YANG);
    printf("load d.yang %s\n", module ? "success" : "failure");

    ly_ctx_destroy(ly_ctx, NULL);

    return 0;
}
```
result
```
libyang[1]: Internal error (/root/z-conf/libyang/src/resolve.c:2088).
libyang[1]: Invalid value "0.1..0.6" of "range".
libyang[1]: Module "d" parsing failed.
load d.yang failure
```
d.yang
```
module d {
  namespace "ddd:1";
  prefix d;
  
    leaf f1 {
      type decimal64 {
        fraction-digits 1;
        range "0.1..0.6";
      }
      default 0.2;
    }
}
```

michalvasko:
Hi,
this fix was only in the devel branch, I added it to the master branch, so it should work now.

Regards,
Michal

fanchanghu:
Thanks, I will update from  the master branch later.

PR #140, opened 2016-10-05
Fix range check after decimal64 bugfix

denisvm:
Fix range check `lyp_check_length_range` function after #119 committed at 9efcd5b5aa29b2417d3644ea5338981c8b0be7b0.

I got the following error when tried to use ietf-inet-types schema:
```
libyang[1]: Invalid value "0..63" of "range".
libyang[1]: Module "ietf-inet-types" parsing failed.
```

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -684,13 +684,9 @@ upper:
                 goto error;
             }
             c = tail;
-            if (*c == '.') {
+            if (type->base == LY_TYPE_DEC64 && *c == '.' && isdigit(c[1])) {
                 c++;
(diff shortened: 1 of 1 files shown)


rkrejci:
I believe that this should be fixed in devel branch already, can you confirm? We should merge devel into master within a few hours, so you can also wait for this and check it in master later.

denisvm:
Hello @rkrejci! Thanks for your feedback. I've tried the devel branch and it's fixed there, also it's a lot cleaner and consistent. Sorry, I should have checked the devel branch before.

rkrejci:
ok, thanks for testing!

PR #141, opened 2016-10-05
Fix tests in Mac OS

denisvm:
Fix three tests that were raising exceptions, replaced string with a more meaning one and length with `strlen`.


rkrejci:
already fixed in devel. However, I like your code more, so I have cherry-picked it as e741628f1 into devel, it should be merged into master within a few hours.

PR #163, opened 2016-10-25
lyd_print_mem print string when using boolean or decimal64 type

yanjunxi:
When I run main function, I expect the result is {"Sub:Root":{"a":true,"b":3.14}}.
But the actual result is {"Sub:Root":{"a":"true","b":"3.14"}}.

[issue.txt](https://github.com/CESNET/libyang/files/549467/issue.txt)


michalvasko:
Hi,
you are correct about bool, but not about decimal64 ([RFC7951](https://tools.ietf.org/html/rfc7951#section-6.1)).

Regards,
Michal

PR #167, opened 2016-10-25
Create libyang static library

abhikeshav:
Hi,
Added some lines to CMakeLists.txt to create static `libyang` library as well. This is useful to some applications using `libyang`.

Thanks,
Abhi

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,16 @@ set(headers
 	src/xml.h
 	src/dict.h)
 
+# create static libyang library
+if(STATIC)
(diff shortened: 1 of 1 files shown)


rkrejci:
I have 2 notes:
- IMHO the static library should be compiled without `-fPIC`, so you should move the target definition before the setting `CMAKE_POSITION_INDEPENDENT_CODE` (line 114)
- since it make to compile all the sources twice, I would prefer to have it just optional (let say via cmake -DSTATIC option or something like that)

abhikeshav:
Thanks! Updated per your comments

PR #179, opened 2016-10-31
cmake: Fix manpage installation target directory

jktjkt:
On my cmake 3.6.2, there's no default and therefore this does not even
respect CMAKE_INSTALL_PREFIX.


rkrejci:
Thanks! I have overlooked the issue.

PR #186, opened 2016-11-04
Make use of CMake's GNUInstallDirs

jktjkt:
My main motivation is the first commit, just including this module enables cross-building of Debian packages (along with manually specifying ``$CC`` with a full arch triplet because of $reasons, apparently).

But since this module exists, maybe these semi-standard variables should be used instead of hand-crafted ones -- hence the second commit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.9)
 project(libyang C)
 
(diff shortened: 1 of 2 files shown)


jktjkt:
OOps, seems that I made a mistake -- will push a fix shortly.

jktjkt:
OK, I've verified that commit bffd9e4 works with an entire stack (for me : )).

PR #255, opened 2017-02-02
docs: Explain that Flex and Bison are needed

### LINKED ISSUE #254: Unable to install libyang (required for libnetconf2)
All-

I'm trying to get libenetconf2 installed and as a pre-requisite it looks like
[libyang-install-fails.txt](https://github.com/CESNET/libyang/files/746946/libyang-install-fails.txt)
 I need to install libyang. While attempting that I saw errors with missing 'flex' and 'bison' executables. Is this expected ? I don't see them listed as pre-requisites for libyang (not [here ](https://github.com/CESNET/libyang/blob/master/README.md)at least). Thoughts ?  I've attached the console logs from my attempt

TIA

jktjkt:
fixes #254

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ as well as YANG 1.1 ([RFC 7950](https://tools.ietf.org/html/rfc7950)).
 ## Requirements
 
 * cmake >= 2.8.9
+* flex
+* bison
 * libpcre
 * cmocka >= 1.0.0 (for tests only, see [Tests](#Tests))
 


rkrejci:
hmm, actually the flex/bison generated files should be available in the git, so you shouldn't need them to compile libyang. The `make bison` target is for the case that something has changed in the flex/bison input files.

jktjkt:
Ah, right, sorry for noise (``ENOCOFFEE``).

Issue #281, opened 2017-03-10
lyd_merge does not handle non-keyed lists properly?


dmcpriv:
In the tree_data::lyd_merge_node_equal function, the merge of a list checks for matching elements in the list based on key.  For operational data, they may be no keys defined.  I think we need to treat these as non-matching nodes so they are merged properly.

Of course this opens the door for duplicates if the merge is called repeatedly, but I don't think there is any other way to detect an equivalent element without a key.

```c
    case LYS_LIST:
        child1 = node1->child;
        child2 = node2->child;
        child1_sch = NULL;
        /* the exact data order is guaranteed */
        for (i = 0; i < ((struct lys_node_list *)node1->schema)->keys_size; ++i) {
            if (sch1 && child1) {
                /* we have different contexts */
                if (!child1_sch) {
                    child1_sch =  lyd_get_schema_inctx(child1, node2->schema->module->ctx);
                    if (!child1_sch) {
                        LOGERR(LY_EINVAL, "Target context does not contain schema node for the data node being "
                               "merged (%s:%s).", child1->schema->module->name, child1->schema->name);
                        return 0;
                    }
                } else {
                    /* just move to the next schema node */
                    child1_sch = child1_sch->next;
                }
            }
            if (!child1 || !child2 || ((child1_sch ? child1_sch : child1->schema) != child2->schema)
                    || strcmp(((struct lyd_node_leaf_list *)child1)->value_str, ((struct lyd_node_leaf_list *)child2)->value_str)) {
                break;
            }
            child1 = child1->next;
            child2 = child2->next;
        }
        if (i == ((struct lys_node_list *)node1->schema)->keys_size) {
            return 1;
        }
        break;
```

michalvasko:
Hi,
please fix the formatting and it should be fine, thanks.

Regards,
Michal

PR #282, opened 2017-03-10
Fix lyd_merge for lists without keys

dmcpriv:
Attempts to fix https://github.com/CESNET/libyang/issues/281


dmcpriv:
Formatting fixed.  Thanks!

PR #300, opened 2017-04-01
data tree BUGFIX insert sibling container for lyd_merge

### LINKED ISSUE #299: lyd_merge does not works for sibling containers, in latest devel
Hi,
lyd_merge does not work with following code:
```
int main(void)
{
    struct ly_ctx *ly_ctx = NULL;

    ly_verb(LY_LLDBG);

    ly_ctx = ly_ctx_new(".");
    ly_errno = 0;
    module = lys_parse_path(ly_ctx, "A.yang", LYS_IN_YANG);
    struct lyd_node *node1 = lyd_new_path(NULL, ly_ctx, "/A:A/B/f2", "aaa", 0, LYD_PATH_OPT_UPDATE);
    struct lyd_node *node2 = lyd_new_path(NULL, ly_ctx, "/A:A/C/f3", "bbb", 0, LYD_PATH_OPT_UPDATE);
    int rc = lyd_merge(node1, node2, 0);
    if(rc || ly_errno || !node1 || !node2){
        printf("lyd_merge err.");
        return -1;
    }
    lyd_print_file(stdout, node1, LYD_JSON, 0);
    printf("\n");

    ly_ctx_destroy(ly_ctx, NULL);

    return 0;
}
```
output:
```
{"A:A":{"B":{"f2":"aaa"}}}
```
A.yang
```
module A {
  namespace "aa:A";
  prefix A;

  container A {
  	leaf f1{type string;}
  	container B{
  	  leaf f2{type string;}
  	}
  	container C{
  	  leaf f3{type string;}
  	}
  }
}
```

fanchanghu:
Hi,
I try to fix #299, it's OK in my test case, but I'm not sure it's OK in other cases...

### DIFF
--- a/src/tree_data.c
+++ b/src/tree_data.c
@@ -2009,9 +2009,8 @@ src_skip:
                 /* no children (or the whole subtree will be inserted), try siblings */
                 if (src_elem == src) {
                     /* we are done, src has no children ... */
-                    if ((src_elem->schema->nodetype == LYS_CONTAINER)
-                        || ((src_elem->schema->nodetype & (LYS_LEAF | LYS_ANYDATA)) && trg_child)) {
-                        /* and it's a container (empty one) or an already-updated leaf/anydata, nothing else to do */
+                    if ((src_elem->schema->nodetype & (LYS_LEAF | LYS_ANYDATA)) && trg_child) {
+                        /* and it's an already-updated leaf/anydata, nothing else to do */
                         break;
                     } else {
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
thanks for the pull request, but in these more complex functions it is enough if you just create an issue, we will fix it ourselves.

Regards,
Michal

fanchanghu:
oh, thanks.

PR #323, opened 2017-05-19
Tolerate lack of trailing zeros when comparing decimal64 list keys

jktjkt:
Various callers use different conventions when formatting `decimal64` values. That is a problem when two un-normalized values are compared via a plain `strcmp`.

This commit fixes just that one single instance where I hit that during my testing. I suppose that there might be other instances where this is done. In my opinion, libyang should define a convention on how `decimal64` numbers are formatted, and enforce (or fix) that as these values pass any borders and interfaces.

IMHO, this is an ugly hack. It works if the list index is a `decimal64` or a `leafref` pointing to a `decimal64`. I hope that it is not allowed to have a `leafref` pointing to a `leafref` pointing to a `decimal64`, but it's too late on a Friday evening to verify that, sorry.

A proper fix would be some other comparison than `strcmp` over a string serialization, IMHO.

Cc issue CESNET/Netopeer2#143.

### DIFF
--- a/src/tree_data.c
+++ b/src/tree_data.c
@@ -1155,12 +1155,41 @@ lyd_new_path(struct lyd_node *data_tree, struct ly_ctx *ctx, const char *path, v
                         ly_errno = LY_EINVAL;
                         return NULL;
                     }
-                    if (!value || strcmp(((struct lyd_node_leaf_list *)parent)->value_str, value)) {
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
fixed instead with commit 27b91d0a. We already have functionality for converting to canonical values, it just had to be used properly.

Regards,
Michal

PR #327, opened 2017-05-29
update the javascript bindings

mislavn:
I made some changes to the javascript bindings.

There are now two npm packages, they are automaticly selected by the cmakes build type ("Debug" or "Release")

The cmake file is simplified and does not compile the binary, that is
done entirely with "npm install --unsafe-perm" or "npm publish". Because of that the minimum cmake version for the javascript bindings is 2.8.12.


mislavn:
@rkrejci I will start work on the libyang python bindings this week.

PR #328, opened 2017-06-05
Create package libyang and libyang-experimental

PavolVican:


rkrejci:
merged manually into both devel and master branches

PR #342, opened 2017-08-02
Fix for Decimal64

mohitarora24:
Negative numbers with trailing zeroes were handled incorrectly.
Was working fine for 0.11 or 0.01 but failed for -0.01 or -0.11.


rkrejci:
merged manually to both devel and master branches.

PR #352, opened 2017-08-25
fix the issue that causes the length overwritten.

zyxel-jimmywang:
the length of a base64 symbol chunk (unum) is overwritten by the length of original binary with heading/tailing space chunk (u).

Signed-off-by: Jimmy, Chun-Tsung Wang <jimmy.wang@zyxel.com.tw>


rkrejci:
Thanks for fixing!

PR #359, opened 2017-09-12
Fix for double free

frimpler:
The pull request contains a unit test which stimulates a double free in connection with deviation handling. In order to make this double free visible, I have chosen to make a small change in CMakeLists.txt (test directory) in order to set the environment variable MALLOC_CHECK_=3 for this test case (otherwise the test will NOT fail!). 
Beside, I had to notice, when I enabled it for all schema tests, also "test_extensions" failed - but which I didn't check any closer.
And of course, there is a suggestion how to fix this (see parser_yang.c and tree_schema.c). 

Background:
during the parsing a shallow copy for iffeatures has been created, but when freeing the copied pointer will be freed. I added therefore a "shallow free" for it.


rkrejci:
Hi Frank, thanks for the patch! However, with the test you have written, the valgrind is also able to catch the double free and it seems to me that it provides more useful information. So what would be the reason to use `MALLOC_CHECK_`?

I also tried to set it to all tests, but no one else failed.

frimpler:
Hi Radek, yes, valgrind should be able to reveal this double free in any case. However, a test case is made to trigger a certain behaviour and show the outcome immediately. But if in our case the heap check is not enabled during the test, the test will succeed with or without the double free error. So therefore the test is rather useless w/o the heap check and I guess not all of us will check all things with valgrind . However, I don't really like this single line in the CMakeLists.txt which I have added, but I didn't dare to set MALLOC_CHECK_ for all, especially when I (mistakenly) had to find out, that the test_extentions failed by this setting. In the meantime I know why this happend, (it's the same reason, why my first commit attempt didn't pass Travis) and I'd appreciate, if you could places this set_property state within the test loops. What do you think?

rkrejci:
Agreed, that's make sense. I'll do the change and merge the pull request manually.

frimpler:
Thanks. Question: do I or you close the pull request and when?

jktjkt:
@frimpler , I do not think that adding ad-hoc flags for glibc's implementation of `malloc` is a best possible option here. Have you considered extending the Travis CI setup so that it runs the test suite under valgrind or --even better -- with a compiler that enables ASAN (`-fsanitize=address`)?

rkrejci:
@jktjkt - since it is just an environment variable, I see it just as an extension of our tests (in contrast to valgrind or ASAN, it does not need any changes in compiled binary), especially in case they are run without valgrind. It is merged, but in case there will be some issues with it, it is just a single line in test's CMakelist.txt, so it's easy to revert.

frimpler:
Jan, thanks for your comment. As you can read in Radek's comment, valgrind reveals this double free already in line of the existing CI pipeline. ASAN would be just another / or better(?) means in this CI pipeline. The discussion about this should in my opinion be made with the maintainers of it (I'm quite new here and learn how things currently are...:-)) However, for me as a developer is important to have the tests available in my local environment in order to check whether I have destroyed something...

PR #361, opened 2017-09-20
Resolve augments before leafref features will be checked

frimpler:
Hi,
I have added 2 unit tests, which trigger in two different ways that the yang parser won't be able to resolve features of leafrefs, because they are part of an augment and the feature is part of the augmented tree. Both test cases address different branches in source code and have to be considered both for the possible solution.

In general, this is a problem of the parsing sequence (and might work the other way round). I have  added a suggestion how to solve this problem. Please have a look at the solution. 

Regards,
Frank

### DIFF
--- a/src/resolve.c
+++ b/src/resolve.c
@@ -6157,7 +6157,8 @@ static int
 check_leafref_features(struct lys_type *type)
 {
     struct lys_node *iter;
-    struct ly_set *src_parents, *trg_parents, *features;
+    struct lys_node_augment *aug;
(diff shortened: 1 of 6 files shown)


michalvasko:
Hi,
thanks for the bug report and test case but I fixed the problem itself in a different manner, there was simply a problem with the order of things being resolved.

Regards,
Michal

frimpler:
Thanx, Michal
I expected/hoped that you will find a better place. In fact my "solution" was just a workaround which I really didn't like. I assume the test cases helped to verify the correctness of the implementation :-)

Regards,
Frank

PR #369, opened 2017-10-25
Test and a (possible) solution for a grouping/augment resolution issue

frimpler:
Hi, 

as mentioned in the subject, I have created a test case which triggers a resolution issue. The issue is, that a node-id in a refinement statement cannot be resolved, because some "uses-augments" have not yet been resolved at this point. 
I have further provided a possible solution for it. 
(Btw. instead of calling apply_aug() directly I tried first unres_schema_add_node(), which didn't solve the problem, because the module has not implemented flag set.)

Regards,
Frank


rkrejci:
Thanks for the test case, Frank!

I have merged it manually with the following changes:
- there was printing schema for no reason in test case (with memory leak) so I've removed it
- I'va fixed it in other ways, I believe that the uses's augments should be applied all the time, not only if they are in an instantiated grouping, so it is now resolving the augments if it is in uses or if it is placed in the implemented module (applies to the top level augments)

frimpler:
I'm lucky with your change, because I realized indeed that the root cause might be rather in resolve_augment(), but I didn't change there because I'm still a bit confused about the "implemented" flag handling of a module...

PR #370, opened 2017-10-26
cmake: Do not write to the source directory

jktjkt:
We're including libyang as a git submodule in our downstream project.
Once we perform the build for the first time, `git status` in our
internal integration repository starts showing the libyang submodule as
"dirty" because libyang's build places files into the source directory.

This commit ensures that the original, source directory remains
unmodified.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,14 +48,15 @@ else()
     set(ENABLE_CACHE_MACRO "/**\n * @brief Cache of some temporary information will not be used.\n */")
 endif()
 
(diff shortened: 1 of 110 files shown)


rkrejci:
Hi, is it really so important to keep the source directory clean? From my, developer, point of view, it is quite usefull to have some of the generated files (modified header files) together with the rest of sources since it simplifies searching for definitions placed in those headers.

frimpler:
Hi, 
I would like to support this pull request. It is essential for our product builds, that source directories have to be kept clean. I have noticed the addressed issue, too, but ignored it up to now, because I'm still in an evaluation phase.

rkrejci:
ok, merged.

PR #378, opened 2017-11-08
testcase and solution for an augment resolution issue

frimpler:
Hi,

the pull request contains a test case which triggers an issue in augment resolution and provides a possible solution for it.

Description:
---------------
A grouping has been defined in a module's sub-module. It uses a further grouping from an imported module and augments it by a relative path. This fails, because the resolution does only consider the sub-module as starting point.

Solution:
-----------
As a possible solution, the main module has been set as starting point in resolve_schema_nodeid(), if the target is a relative path.

Regards,
Frank Rimpler


michalvasko:
Hi Frank,
thanks a lot, you are right. Wherever a module is needed, those functions that get the main module should always be used, we missed it here. Please, just address my request and it can be merged.

Regards,
Michal

src/resolve.c
```diff
@@ -1764,7 +1764,8 @@ resolve_schema_nodeid(const char *nodeid, const struct lys_node *start, const st
 
     /* descendant-schema-nodeid */
     if (is_relative) {
-        cur_module = start_mod = start->module;
+        // cur_module = start_mod = start->module;
+        cur_module = start_mod = lys_main_module(start->module);
```
michalvasko: Please rather use `lys_node_module(start)`, it is meant for these cases, thanks.

frimpler: done

frimpler: Unfortunately the imp_aug branch was not clean...
Next attempt.

frimpler: from my point of view, it's through. I don't know what codecov is complaining about, I guess, it's because of the removed test case (which was not intended to be part of the pull request) . Does it compare to the previous build?

michalvasko:
Hi,
nevermind *codecov*, as long as the tests pass, it is fine. Thanks again.

Regards,
Michal

PR #385, opened 2017-11-09
Fixing the empty test string

plewyllie:
Hi,

I'm using the yangre tool as backend of the yangre GUI (https://yangcatalog.org/yangre/). 

The problem is that having an empty string is not supported with the current implementation. So, if a user enters the pattern ([a-zA-Z])? without any input string, we get the error "yangre error: Value " " does not satisfy the constraint "([a-zA-Z])?" (range, length, or pattern).".

I am relying on the file mechanism with the "-f" parameter, so I have added the logic to support an input file with only a pattern and an empty test string.

The test string is considered empty when the file contains:
pattern\n
\n

The test string will be equal to "\n" when the file contains:
pattern\n
\n
\n

As an example, the test file containing:
([a-zA-Z])?\n
\n
will now match.

### DIFF
--- a/tools/re/main.c
+++ b/tools/re/main.c
@@ -150,19 +150,19 @@ main(int argc, char* argv[])
             }
 
             while((l = getline(&str, &len, infile)) != -1) {
-                if (str[0] == '\n') {
+                if (!blankline && str[0] == '\n') {
                     /* blank line */
                     blankline = 1;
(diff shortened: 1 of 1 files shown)


rkrejci:
Merged manually with fixing coding style and adding info into help printing. Pieter, thank for the patch!

PR #397, opened 2017-11-20
API lys_parent should handle augment node correctly

ustcpxy:


michalvasko:
Hi,
`lys_parent` was meant to be used when one does not want to work with augments at all, which is almost always, but I do not see a problem with adding the functionality you propose. However, please use the same coding style as the rest of the code, then I can merge it.

Regards,
Michal

ustcpxy:
Hi Michal,
the caller should guarantee the node is not augement. right?
actually, the issue was found by using sysrepo. 
I also open an issue https://github.com/sysrepo/sysrepo/issues/993 in sysrepo. 

modify sysrepo or libyang? I have no idea. could you give your opinion? Thanks.

PR #423, opened 2018-01-19
add C++ exceptions (and Python) to C++ language bindings

mislavn:
This is the first version of the C++ exception refactoring.

@rkrejci please can you check if this patch works in your use case.

### DIFF
--- a/swig/cpp/src/Libyang.cpp
+++ b/swig/cpp/src/Libyang.cpp
@@ -289,3 +289,8 @@ int Set::rm(S_Schema_Node node) {
 int Set::rm_index(unsigned int index) {
     return ly_set_rm_index(set, index);
 }
+
+/* API for wrapping struct ly_ctx from libnetconf2 python bindings */
+S_Context create_new_Context(struct ly_ctx *new_ctx) {
+    return new_ctx ? std::make_shared<Context>(new_ctx, nullptr) : nullptr;
+}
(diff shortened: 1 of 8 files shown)


swig/cpp/src/Libyang.cpp
```diff
@@ -36,27 +36,44 @@ Context::Context(ly_ctx *ctx, S_Deleter deleter):
...
 Context::~Context() {}
+int Context::set_searchdir(const char *search_dir) {
+    int ret = ly_ctx_set_searchdir(ctx, search_dir);
+    if (ret) {
+        check_libyang_error();
+    }
+    return ret;
+}
+void Context::unset_searchdirs(int idx) {
+    int ret = ly_ctx_unset_searchdirs(ctx, idx);
```
rkrejci: ly_ctx_unset_searchdirs() returns void, it does not fail.

mislavn:
There is still some issues with:

```
>>> import libyang as ly
ImportError: dynamic module does not define module export function (PyInit_libyang)
```

I'll take a look.

mislavn:
I think I know what is the problem. I just need to adjust my cmakefiles for the commit fd7a5c72a7c45750166f1fa5b23d73250b4cf5fa .

rkrejci:
So, should I wait with the merge? I'm not able to reproduce this issue when merged with `swig_change` branch.

mislavn:
It's up to you. I'll probably fix the issue today.

PR #426, opened 2018-01-24
Use absolute paths in includes for generated libyang.h file

jeroen-dhollander:
I work on a project that uses the libyang library,
and to do so I need to include the generated 'libyang.h' file
in my source code.

However, because the includes inside 'libyang.h' use relative paths,
I need to include both the libyang source and build directories:

    include_directories(${PROJECT_BINARY_DIR}/libyang/src ${PROJECT_SOURCE_DIR}/libyang/src)

With this change, only the binary directory must be included,
making it easier to use libyang.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,8 @@ else()
 endif()
 
 include_directories(${PROJECT_BINARY_DIR}/src ${PROJECT_SOURCE_DIR}/src)
+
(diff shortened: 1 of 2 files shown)


rkrejci:
libyang is a library supposed to be included from the system paths:
```
#include <libyang/libyang.h>
```
With your change, the include paths in libyang.h will allways point to some local files in src, that's nonsence.

PR #428, opened 2018-01-30
do not extract namespaces from literals

ustcpxy:


michalvasko:
Hi,
why not? The function is used for NETCONF XPath filter and there could be a foreign identity used (is parsed as literal), the namespace must be extracted from it.

Regards,
Michal

ustcpxy:
ok, I have reverted back the code. 
but I think the ns_count should only be cleared in `_transform_json2xml`. 
please have a check. Thanks.

michalvasko:
Hi,
right, thanks.

Regards,
Michal

PR #432, opened 2018-02-09
Swig vector

mislavn:
This pull request fixes a few C++ compiler warnings, add's a size check in the LY_NEW* macro and creates a vector for Type_Info_Enums::enm().

### DIFF
--- a/swig/cpp/src/Internal.hpp
+++ b/swig/cpp/src/Internal.hpp
@@ -98,6 +98,9 @@
 
 #define LY_NEW_LIST(data, element, size, class)\
     {\
+        if (0 >= data->size) {\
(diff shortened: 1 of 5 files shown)


swig/cpp/src/Internal.hpp
```diff
@@ -98,6 +98,9 @@
 
 #define LY_NEW_LIST(data, element, size, class)\
     {\
+        if (0 >= data->size) {\
+            return NULL;\
```
rkrejci: shouldn't it return 'nullptr'?

mislavn: You are right, I forgot about it.

PR #437, opened 2018-02-13
Bind errors with and store them in a context

michalvasko:
Now all the errors are not just thread but also context-specific. This further allows to store and collect several errors and not just the last one. Lastly, the whole logging mechanism was refactored.


mislavn:
Hi Michal,

I was thinking of creating a C++ vector for the Errors, I could name the function get_ly_errors() or something like that. Also I would hide the Error class constructor from the user so he would have to use the function.
What do you think?

Regards,
Mislav

michalvasko:
Hi Mislav,
would be great, I was just too lazy to implement it properly like that (also, I don't know C++ that well) so I just wanted it to compile. I will merge this and you can then create another pull request (no other way anyway, I think).

Michal

mislavn:
ok

Mislav

mislavn:
Also, you should add a NULL check at:
```
API struct ly_err_item *
ly_err_first(const struct ly_ctx *ctx)
{
    return pthread_getspecific(ctx->errlist_key);
}
```

michalvasko:
Right, thanks.

PR #438, opened 2018-02-14
Cpp log update

mislavn:
I implemented the libyang error handling changes into C++.

Also I changed the code logic so that C++ vectors never return NULL, this removes the need to wrap C++/Python for loop in a if statement.


mislavn:
@rkrejci please can you take a look at the changes.

Issue #443, opened 2018-02-17
Compilation failure on OS X 10.10.5 (darwin 14.5.0)


andaru:
Problem: libyang doesn't build on an OS X 10.10.5 systems (perhaps all versions 10.10 or less, judging by similar bugs in other repositories).

```
$ make 
Scanning dependencies of target yangobj
[  2%] Building C object CMakeFiles/yangobj.dir/src/common.c.o
[  5%] Building C object CMakeFiles/yangobj.dir/src/context.c.o
[  8%] Building C object CMakeFiles/yangobj.dir/src/log.c.o
[ 10%] Building C object CMakeFiles/yangobj.dir/src/dict.c.o
[ 13%] Building C object CMakeFiles/yangobj.dir/src/resolve.c.o
[ 16%] Building C object CMakeFiles/yangobj.dir/src/validation.c.o
[ 18%] Building C object CMakeFiles/yangobj.dir/src/xml.c.o
[ 21%] Building C object CMakeFiles/yangobj.dir/src/parser.c.o
/Users/me/go/src/github.com/CESNET/libyang/src/parser.c:288:74: error: use of undeclared identifier 'MAP_ANONYMOUS'
        addr = mmap(NULL, *length, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
                                                                         ^
1 error generated.
make[2]: *** [CMakeFiles/yangobj.dir/src/parser.c.o] Error 1
make[1]: *** [CMakeFiles/yangobj.dir/all] Error 2
make: *** [all] Error 2
```

Darwin's POSIX headers only define `MAP_ANON`, hence adding this to `src/parser.c` corrects the issue for me.

```
#ifdef __APPLE__
#define MAP_ANONYMOUS MAP_ANON
#endif
```

PR #444, opened 2018-02-17
parser: correct compilation error on Mac OS X <= 10.10

andaru:
On these systems, only the (deprecated) POSIX MAP_ANON definition is
provided; MAP_ANONYMOUS is not provided. Thus this change provides the
alias on Apple systems to correct this compilation error:

  .../libyang/src/parser.c:286:75: error: use of undeclared identifier 'MAP_ANONYMOUS'
        *addr = mmap(NULL, *length, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
                                                                          ^
  1 error generated.

(Corrects #443)

### DIFF
--- a/src/parser.h
+++ b/src/parser.h
@@ -23,7 +23,9 @@
 #include "tree_internal.h"
 
 #ifdef __APPLE__
-#define MAP_ANONYMOUS MAP_ANON
+# ifndef MAP_ANONYMOUS
+#  define MAP_ANONYMOUS MAP_ANON
+# endif
 #endif
 
 /**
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
it makes sense but I would rather be more careful, what if in some newer versions Apple decides to remove `MAP_ANON` and use the correct macro `MAP_ANONYMOUS`? You fill fail to compile again. So, I would condition the redefinition not just on `__APPLE__` being defined but also on `MAP_ANONYMOUS` not being defined.

Regards,
Michal

andaru:
Good idea, will do.

michalvasko:
Seems fine, thanks.

Regards,
Michal

PR #453, opened 2018-03-07
Python: use fixed path for module path search

mislavn:
Fixed the issue where the python module path would be searched based on the CMAKE_INSTALL_PREFIX. Now the PYTHON_MODULE_PATH is set to /usr/lib/python[2|3]/dist-packages.

One issue remains, python does not look for libyang-cpp in /usr/local, so the python bindings will work only when CMAKE_INSTALL_PREFIX is set to "/usr".

I'm trying to fix this issue in swig.

### DIFF
--- a/swig/python/CMakeLists.txt
+++ b/swig/python/CMakeLists.txt
@@ -16,13 +16,9 @@ execute_process(COMMAND ${SWIG_EXECUTABLE} -python -external-runtime ${CMAKE_CUR
 
 file(COPY "examples" DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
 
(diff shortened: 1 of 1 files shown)


katdiam:
Is the issue "python does not look for libyang-cpp in /usr/local" fixed?

Still seen: 
>>> import yang
  File "/usr/lib/python2.7/dist-packages/yang.py", line 18, in <module>
    _yang = swig_import_helper()
  File "/usr/lib/python2.7/dist-packages/yang.py", line 17, in swig_import_helper
    return importlib.import_module('_yang')
  File "/usr/lib/python2.7/importlib/__init__.py", line 38, in import_module
    __import__(name)
ImportError: libyang-cpp.so.1: cannot open shared object file: No such file or directory
>>>

PR #457, opened 2018-03-09
Python: Add unit tests

mamidzic:
Add Python unit tests based on C unit tests, specifically test_libyang.c, test_tree_data.c and test_tree_schema.c. 

Also fixes C++ implementation of 'Set' and implements C++ versions of lys_parse_mem and lys_parse_fd functions.

Generated Python tests:
python_test_libyang
python_test_tree_data
python_test_tree_schema

### DIFF
--- a/swig/cpp/src/Libyang.cpp
+++ b/swig/cpp/src/Libyang.cpp
@@ -304,13 +304,13 @@ LY_LOG_LEVEL set_log_verbosity(LY_LOG_LEVEL level)
 }
 
 Set::Set() {
-    struct ly_set *set = ly_set_new();
-    if (!set) {
+    struct ly_set *set_new = ly_set_new();
+    if (!set_new) {
         check_libyang_error(nullptr);
     }
 
-    set = set;
-    deleter = std::make_shared<Deleter>(set);
+    set = set_new;
+    deleter = std::make_shared<Deleter>(set_new);
 }
 Set::Set(struct ly_set *set, S_Deleter deleter):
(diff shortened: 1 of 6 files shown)


swig/cpp/src/Libyang.cpp
```diff
@@ -206,6 +206,31 @@ S_Data_Node Context::parse_fd(int fd, LYD_FORMAT format, int options) {
     S_Deleter new_deleter = std::make_shared<Deleter>(new_node, deleter);
     return std::make_shared<Data_Node>(new_node, new_deleter);
 }
+
+S_Module Context::parse_module_mem(const char *data, LYS_INFORMAT format) {
```
rkrejci: please unify naming - `lys_parse_path()` was wrapped as `Context::parse_path()`, so similarly here I would expect `Context::parse_fd()` and `Context::parse_mem()`

mislavn: It was not possible to ta name it that because it would get in conflict with:
`S_Data_Node parse_mem(const char *data, LYD_FORMAT format, int options = 0);`

I can name it `parse_mem()` if I remove the default value `options = 0` from `S_Data_Node parse_mem()`.

What would you like to do?

rkrejci: Then, please, rename `parse_path()` to `parse_module_path()` and probably it would be also better to have `parse_data_*()` functions. Currently we mix wrapper names for lys_ and lyd_ functions.

PR #468, opened 2018-03-26
fix crash in lyd_free_value()

frimpler:
Hi,

this pull request contains 
- a fix for crashes in lyd_free_value() in error scenarios when the value_type is set to LY_TYPE_ERR, but it's nothing to be freed. 
- a test case to verify the correction.

Regards,
Frank


michalvasko:
Hi Frank,
I would merge the test case if you create a new pull request but I will not merge the whole pull request although your fix is basically correct. There was a more general problem which I tried to fix in the referenced pull request, which should also fix your problem. Thanks anyway!

Regards,
Michal

PR #469, opened 2018-03-26
data tree BUGFIX return node when updating leaf with same value

olivier-matz-6wind:
According to the documentation of LYD_PATH_OPT_UPDATE:

  If the target node exists and is a leaf, it is updated with the new
  value and returned. If the target node exists and is not a leaf, NULL
  is returned and no error set.


michalvasko:
Hi,
in this case no change occurred so no node is returned. I agree that the documentation is not specific about this, but it is already improved in the `devel` branch so I will not be merging this change.

Regards,
Michal

PR #470, opened 2018-03-26
Split type base and type flags

michalvasko:
Split type base enum and type flags. It is now more clear and easily extensible with new flags.


mislavn:
@michalvasko you are right, the attribute can be ignored. The best approach is to add `#define __attribute__(x)` at the top of cpp_classes.i.

```
%module cpp_classes

#define __attribute__(x)
%include <std_except.i>
%catches(std::runtime_error, std::exception, std::string);
```

michalvasko:
Hi Mislav,
thanks a lot, I did not know where to put it.

Regards,
Michal

PR #480, opened 2018-04-06
Memory Leak Fix - Instance Identifier Value inserted into dictionary twice

mohitarora24:
Just copied the code from one place and pasted at other affected places.
It affected me when parsing a notification xml (having an instance identifier leaf) into libyang node via lyd_parse_mem() and then freeing it via lyd_free(). The value in provided XML was already in correct format and transform_xml2json() produced the same value.


rkrejci:
Thanks.

PR #482, opened 2018-04-11
Identityref fix

### LINKED ISSUE #481: When condition with prefixed value failed
Hi,

with the testcase test_when_value_prefix() in this commit https://github.com/frimpler/libyang/commit/a9451c16ab9c82ae8e64b915f269377365ee0ecc an issue of the XPATH evaluator is triggered.

### Description
When a when condition is evaluated which contains a prefixed identityref value and this identityref comes from an augmentig module the evaluated expression looks like "identityref = prefix:identityref" and this fails of course. The reason is that on the node side of the equality expression the value is assumed to come from the augmented module (resulting in no prefix) instead of the augmenting module.

### Proposed Solution

The proposed solution in xpath.c (https://github.com/frimpler/libyang/commit/9b58e81db7a8f8ffe9b8a703a58842f25d78f466) uses the module of the actual node to determine the prefix for the value.

rkrejci:
Fixes #481

### DIFF
--- a/tests/conformance/sec9_10/mod3.yang
+++ b/tests/conformance/sec9_10/mod3.yang
@@ -34,7 +34,7 @@ module mod3 {
             base sub2;
         }
 
-        when "../test = 'j4'";
+        when "../test = 'abc:j4'";
     }
 
     identity i3 {


frimpler:
Hi Radek,

I had a look at your correction. Just from "reading" it, I'd say it's ok. Unfortunately, Peter has investigated this problem in depth, but he'll be back only on Tuesday. So my suggestion would be, if the test case run through (what it apparently did) and you don't wait anyway for Michal to review, just pull it in.

Regards,
Frank

rkrejci:
I want to wait for Michal's review with mergin, so Peter's feedback is welcome. I just wanted to provide my solution to others to test (use if needed) before it will be merged (or throw away).

There is a possible issue with non-prefixed identityref values in XPath expression as described in #481 and we should agree that the proposed solution with its limitation is fine.

PR #484, opened 2018-04-16
plugins BUGFIX fix dlclose in musl

mislavn:
dlclose should not be used in musl because musl has an internal reference
counter.

fixes https://github.com/CESNET/Netopeer2/issues/138

### DIFF
--- a/src/plugins.c
+++ b/src/plugins.c
@@ -76,10 +76,12 @@ ly_clean_plugins(void)
     type_plugins = NULL;
     type_plugins_count = 0;
 
+#ifdef HAVE_DLCLOSE
     /* close the dl handlers */
     for (u = 0; u < dlhandlers.number; u++) {
         dlclose(dlhandlers.set.g[u]);
     }
+#endif
     free(dlhandlers.set.g);
     dlhandlers.set.g = NULL;
     dlhandlers.size = 0;
@@ -231,6 +233,7 @@ ly_load_plugins_dir(DIR *dir, const char *dir_path, int ext_or_type)
         /* and construct the filepath */
         asprintf(&str, "%s/%s", dir_path, file->d_name);
 
+#ifdef HAVE_DLCLOSE
         /* load the plugin - first, try if it is already loaded... */
         dlhandler = dlopen(str, RTLD_NOW | RTLD_NOLOAD);
         dlerror();    /* Clear any existing error */
@@ -243,6 +246,7 @@ ly_load_plugins_dir(DIR *dir, const char *dir_path, int ext_or_type)
             dlclose(dlhandler);
             continue;
         }
+#endif
 
         /* ... and if not, load it */
         dlhandler = dlopen(str, RTLD_NOW);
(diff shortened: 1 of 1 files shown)


src/plugins.c
```diff
@@ -78,7 +78,9 @@ ly_clean_plugins(void)
 
     /* close the dl handlers */
     for (u = 0; u < dlhandlers.number; u++) {
+#ifdef HAVE_DLCLOSE
```
totszwai: Put the HAVE_DLCLOSE around the for loop?

src/plugins.c
```diff
@@ -261,10 +265,14 @@ ly_load_plugins_dir(DIR *dir, const char *dir_path, int ext_or_type)
             ret = lytype_load_plugin(dlhandler, name);
         }
         if (ret == 1) {
```
totszwai: Perhaps replace this with:

```
#ifdef HAVE_DLCLOSE
        dlclose(dlhandler);
#endif
        if (ret == 1) {
            continue;
        } else if (ret == -1) {
            break;
        }

```

rkrejci:
and by the way - where is the `HAVE_DLCLOSE` defined? Isn't it false always (even with glibc) just because it is not defined?

mislavn:
I wanted to talk with you about this, so the issue is that musl has defined the dlclose function but it's recommended not to use it because of the internal reference counter.

As I understood HAVE_DLCOSE is not defined in musl source code but it is in OpenWrt toolchain (the macro is commonly used in OpenWrt), not sure how is this handled on other environments, need to check it. Also musl does not have a MUSL macro.

In this case the code does not work because of RTLD_NOLOAD and not dlclose, not sure why yet.

I think it would be safer to refactor the code so RTLD_NOLOAD is not used, what do you think about it?

mislavn:
There might be an issue with musl:
https://www.mail-archive.com/digitalmars-d@puremagic.com/msg228787.html

mislavn:
I was thinking of doing manually the reference counting and call dlclose two times or implement some other check for already loaded binaries.

I'm still trying to figure out if the RTLD_NOLOAD issue could be solved with changing the flags on musl which work on glibc also.

rkrejci:
btw, can you share a link to some code using `HAVE_DLCOSE` (the macro is commonly used in OpenWrt) - I guess there is also some configure code to detect `dlclose()` presence, but I may be wrong.

mislavn:
Most programs simply use `HAVE_DLCLOSE` two use something similar to this:
https://code.it4i.cz/lriha/elmer/blob/master/src/post/config.h.cmake

But I'm still not sure how they check it.

Also do you know why netopeer-cli has this issue while sysrepo does not?

rkrejci:
As I thought, the refered code needs the following line in `CMakeLists.txt`:
```
configure_file(${CMAKE_CURRENT_SOURCE_DIR}/config.h.cmake ${CMAKE_CURRENT_BINARY_DIR}/config.h)
```

Anyway, which musl version are you using? I've tried to use its current git and everything seems working for me. I want to investigate it more, but at least I'm able to run `yanglint` which also loads the plugins.

mislavn:
```
root@Inteno:/# ldd --version
musl libc (arm)
Version 1.1.18
Dynamic Program Loader
Usage: ldd [options] [--] pathname
```

rkrejci:
ok, it should be quite new.

Mislav, you are saying that you see this issue only in netopeer2-cli? Can you confirm, that yanglint works (just try to validate some schema) fine? Because I've just tried step yanglint (minimal dependencies, so the most easy way to try app with musl) in debugger and everything seems fine.

rkrejci:
At the end, I was able to reproduce the issue. The problem is actually inside `ly_clean_plugins()` which removes internal list of handlers when it thinks that there is no one using it and plugins will be removed by dlclose(). But with musl, this is never true, which is not detected. I'm working on fix so I'm closing this pull request since it does not solve the issue (or solve it but introduces another one).

PR #485, opened 2018-04-16
tests for strict/non-strict parsing of unknown elements

frimpler:
Hi,

recently we found problems in parsing of unknown elements. We created a correction and a test case. Michal has pushed almost exactly at that time a bigger amount of code in which (beside others) the found problems were solved. Nevertheless, we think it's worth to have this test case permanently on board when running the CI pipeline.

Regards,
Frank


rkrejci:
Definitely agreed, thanks!

PR #488, opened 2018-04-20
data tree BUGFIX update set number in ly_set_merge

olivier-matz-6wind:
The target set should have its number updated after the merge
operation.

Fixes: 563cfeaa32a9 ("data tree CHANGE new function lyd_set_merge")


rkrejci:
Thanks!

PR #493, opened 2018-04-27
Fix month when filling local validation struct tm.

mohitarora24:
tm_mon is from 0 to 11. So subtracted one. Else validation was failing for dates in December.


michalvasko:
Hi,
thanks, you are right.

Regards,
Michal

PR #495, opened 2018-05-01
Fix issue where removing an empty container could lose data in tree

alangefe:
This issue occurs when an empty non-presence container is present as the first node in a data tree that is passed to lyd_validate. In that case the other nodes in the original tree are lost, and only defaults etc. will be populated into the new tree.

This pull request contains a unit test demonstrating the problem and a source code change that fixes the problem.


michalvasko:
Hi,
nice test and fix, much appreciated.

Regards,
Michal

PR #505, opened 2018-05-21
add static compile support

mislavn:
Hi, I'm almost done with the static build. I need to add unit tests but before that I would like to discuss a few things.

1) What do you think about the code generation part in plugin_config.h.in . There are other ways of doing this but with this approach you need to define the lyext_* and lytpe_* plugin's only in one list variable in the cmake file.
```
set(EXTENSIONS_LIST "nacm" "metadata" "yangdata")
set(USER_TYPE_LIST "user_date_and_time")
```
The list applies for shared libraries also.

2) Now there is one new requirement, the C file must have the same name as the struct:
```
struct lyext_plugin_list nacm[] = {
    {"ietf-netconf-acm", "2012-02-22", "default-deny-write", &nacm_deny_write},
    {"ietf-netconf-acm", "2012-02-22", "default-deny-all", &nacm_deny_all},
    {NULL, NULL, NULL, NULL} /* terminating item */
};

```
or
```
struct lytype_plugin_list user_date_and_time[] = {
    {"ietf-yang-types", "2013-07-15", "date-and-time", date_and_time_store_clb, NULL},
    {NULL, NULL, NULL, NULL, NULL} /* terminating item */
};
```
That's because of the list's EXTENSIONS_LIST and USER_TYPE_LIST.

3) There is a compiler warning for yanglint:
CMakeFiles/yanglint.dir/tools/lint/configuration.c.o: In function `get_yanglint_dir':
/opt/fork/libyang/tools/lint/configuration.c:36: warning: Using 'getpwuid' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking

https://github.com/golang/go/issues/13470

Which means yanglint will work only if it's compiled with the same version as it's run on the host. We can disable yanglint by default in a static build.

4) Because of simplicity, for now, in `plugins.c` I add:
```
#ifdef STATIC
    return;
#endif /* STATIC */
```
If you want we can have two separate files for shared and static build.

5) What should I do with the ext_test extension on a static build. I was thinking of adding it if the tests are enabled, otherwise not.


mislavn:
I'll add support for static swig bindings after I'm done with the unit tests.

michalvasko:
Hi Mislav,

 1. This is the best solution, I think (why wouldn't it be?). Side note, why did you not include `user_ipv4`?
 2. Only the `*.so` had to match before but the comment in user types already said that the source file should be named the same as the structure to keep it simpler, so no problem.
 3. Not sure about this one.
 4. Seems fine to me.
 5. I think that is reasonable.

Regards,
Michal

mislavn:
Hi Michal

1. Wasn't sure what are your opinions on generating code from cmake.
Also, I didn't add `user_ipv4` because it was not added in `src/user_types/CMakeLists.txt`. I'll add it now then
2. Great.
3. I'll skip this part for now.
4. OK.
5. OK

Probably I'll be done with unit tests and swig bindings in 2-3 days.

Regards,
Mislav

mislavn:
Hi Radek,

3. I just checked, you're right with musl. I'll print a message.

4. That's reasonable.

Regards,
Mislav

mislavn:
Hi,

a quick status update:
I made all of the changes, except for 3. 

Now in static mode the valgrind tests are disabled because it's not recommended to run valgrind on a static binary. Also I disabled python bindings in static mode (it's not possible to have static python module).

I updated the javascript bindings so they use the static generated code and C++.

Also there was a bug in the code where only the first plugin was loaded, that's fixed now.

mislavn:
Hi Michal,

yes, I'm working on it also.

Regards,
Mislav

mislavn:
Hi Michal,

I fixed some issues, the tests should pass.

There is one issue, if someone builds libyang and later sets `-DENABLE_STATIC=ON` the build process will fail, even  if `make clean` was used.
Should I update the `make clean` command or write a warning when static build is selected?

Regards,
Mislav

michalvasko:
Hi Mislav,
I see no problem in changeing `make clean` commands if that helps, what additional files need to be deleted?

Regards,
Michal

mislavn:
Hi Michal,

it enough to delete `CMakeCache.txt`, I tried to add it myself but the `make clean` does not remove the file. Do you know how to add it and run the command?

Regards,
Mislav

rkrejci:
Hi, please analyse the cause of error in more details. This is usually caused by relaying on default values which are actually replaced by the values from cache. Example:

```
VAR_A=ON
VAR_B=ON
...
if (VAR_A) 
    VAR_B=OFF
endif
```
first, you run cmake with no parameter, so `VAR_B` is stored in cache with `OFF` value. Second, you run cmake with `-DVAR_A=OFF` and expect that `VAR_B` is `ON`, but it is not - it gets value `OFF` from the cache. The solution is to add 
```
else ()
    VAR_B=ON
endif
```

mislavn:
Hi Radek,

the issue is in `find_package(PCRE REQUIRED)`, for an example if you build a shared library and switch to static `PCRE` will still hold `libpcre.so` and not `libpcre.a` and I don't know how to force a new search.

Do you have any idea how to solve this?

Regards,
Mislav

rkrejci:
I'm fine with forcing to run `find_package()` always by unsetting `PCRE_FOUND` before calling `find_package()` routine. I believe this should work.

michalvasko:
Hi,
could you please look at commit 3b887ec004c4adfe178dc755a5488c3f89a6e02d and make changes so that the new function works even for static build? We can finally merge this afterwards, I think.

Regards,
Michal

mislavn:
Hi,

currently I'm not at the office, I'll finish it on monday.

Regards,
Mislav

mislavn:
Hi, I'm done.

mislavn:
Hi Michal,

I'm a bit confused, does `ly_get_loaded_plugins` also have the user_type plugins?

Regards,
Mislav

michalvasko:
Hi,
yes, it is supposed to be a list of all loaded plugins.

Regards,
Michal

mislavn:
Hi,

I fixed all of the issues, it should pass now with your updated tests.

Regards,
Mislav

PR #507, opened 2018-05-22
Fix strncpy invocation in tests

jktjkt:
`strncpy()` should be used to ensure that we do not overwrite the destination buffer. Calculating `strlen()` of the source does not make sense as it duplicates what `strncpy` is doing internally.

GCC 8.1 warns about this:
```
libyang/tests/api/test_xml.c:247:5: warning: ‘strncpy’ output truncated before terminating nul copying 19 bytes from a string of the same length [-Wstringop-truncation]
     strncpy(file_name, TMP_TEMPLATE, strlen(TMP_TEMPLATE));
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/tests/api/test_tree_data.c
+++ b/tests/api/test_tree_data.c
@@ -1138,7 +1138,7 @@ test_lyd_print_fd_xml(void **state)
     int fd;
 
     memset(file_name, 0, sizeof(file_name));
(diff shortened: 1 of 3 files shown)


michalvasko:
Hi,
you are correct, thanks.

Regards,
Michal

PR #509, opened 2018-05-23
Fix issue when defaults aren't stored in canonical form

alangefe:
Sometimes the `dflt` value on a type isn't stored in canonical form (for example, storing '0xa' instead of '10'). This isn't a problem by itself, but it makes it important that functions that create default values
properly pass-on the `dflt` flag.

This commit also addresses a secondary issue that stems from the primary fix. Because the `dflt` flag is now being propagated properly in more circumstances, an issue crops up when a default value for an
identityref is given in a typedef from a separate module. The fix is to also search the typedef-defining module to find the module that defines the specific identity instance.


michalvasko:
Hi,
you are right, I am not sure why the flag was not passed properly, thanks!

Regards,
Michal

PR #517, opened 2018-05-30
Resolving stack corruption issue when prefixes are longer than 32 characters

aweast:
There are two functions in libyang that have c-string arrays on the stack for XML prefixes. This approach makes the memory allocation dynamic.


michalvasko:
Seems fine, thanks.

Regards,
Michal

PR #522, opened 2018-06-04
fix build on OmniOS/Solaris

rwestphal:
1 - OmniOS doesn't have vdprintf(), use fdopen()+vfprintf() to
    workaround this problem;
2 - Remove the _XOPEN_SOURCE definition in src/user_types/user_date_and_time.c
    to fix this compiler error: "Compiler or options invalid for pre-UNIX 03
    X/Open applications and pre-2001 POSIX applications". This change doesn't
    affect libyang build on Debian 9.3 and OpenBSD 6.3.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 2.8.12)
 project(libyang C)
 
 include(GNUInstallDirs)
+include(CheckSymbolExists)
 
 set(LIBYANG_DESCRIPTION "libyang is YANG data modelling language parser and toolkit written (and providing API) in C.")
 
@@ -231,6 +232,11 @@ set(headers
     src/user_types.h
     src/xml.h
     src/dict.h)
+
+check_symbol_exists(vdprintf stdio.h HAVE_VDPRINTF)
+if(HAVE_VDPRINTF)
+    add_definitions(-DHAVE_VDPRINTF)
+endif(HAVE_VDPRINTF)
 
 # create static libyang library
 if(ENABLE_STATIC)
--- a/src/printer.c
+++ b/src/printer.c
@@ -79,12 +79,23 @@ ly_print(struct lyout *out, const char *format, ...)
     int count = 0;
     char *msg = NULL, *aux;
     va_list ap;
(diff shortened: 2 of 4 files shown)


src/printer.c
```diff
@@ -84,7 +84,15 @@ ly_print(struct lyout *out, const char *format, ...)
...
     case LYOUT_FD:
+#ifdef HAVE_VDPRINTF
         count = vdprintf(out->method.fd, format, ap);
+#else
+	{
+            FILE *stream = fdopen(out->method.fd, "r");
+            count = vfprintf(stream, format, ap);
+            fclose(stream);
+        }
+#endif
```
michalvasko: Why the curly brackets? We try to keep case code without redundant indents so please remove them. Also, `fdopen` error checking would be nice.

CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 2.8.12)
 project(libyang C)
 
 include(GNUInstallDirs)
+INCLUDE(CheckSymbolExists)
```
michalvasko: Just a detail, please keep `INCLUDE` lowercase so that we do not mix it.

rwestphal: Indeed, will update.

src/printer.c
rwestphal: I created a new block because I needed to declare a new variable (`FILE *stream`). But I agree it's ugly, I'll add another `#ifdef` and declare the variable in the top of the function.

michalvasko: Right, forgot about the variable declaration but it will probably be better to put it separately anyway.

rwestphal:
Looks like my changes in `src/printer.c` caused several problems. I believe I'll need to include a new header file to have access to `HAVE_VDPRINTF`. Will take a look in a few hours...

rwestphal:
Done :)

Added a `dup()` inside `fdopen()` to prevent `fclose()` from closing the underlying file descriptor. I also added a `add_definitions()` line in _CMakeLists.txt_ to make the `HAVE_VDPRINTF` constant visible in the source code.

michalvasko:
Hi,
seems fine to me now, merging.

Regards,
Michal

PR #525, opened 2018-06-13
Preventing apply_augment from being called on already applied augments in certain cases

aweast:
@michalvasko, proposed solution for the problem detailed in https://github.com/sysrepo/sysrepo/issues/1173

### DIFF
--- a/src/tree_schema.c
+++ b/src/tree_schema.c
@@ -4454,7 +4454,7 @@ lys_make_implemented_r(struct lys_module *module, struct unres_schema *unres)
 
     for (i = 0; i < module->augment_size; i++) {
         /* apply augment */
-        if ((module->augment[i].flags & LYS_NOTAPPLIED && module->augment[i].target) && apply_aug(&module->augment[i], unres)) {
+        if ((module->augment[i].flags & LYS_NOTAPPLIED) && apply_aug(&module->augment[i], unres)) {
             return EXIT_FAILURE;
         }
     }
@@ -4472,7 +4472,7 @@ lys_make_implemented_r(struct lys_module *module, struct unres_schema *unres)
 
         for (j = 0; j < module->inc[i].submodule->augment_size; j++) {
             /* apply augment */
-            if ((module->inc[i].submodule->augment[j].flags & LYS_NOTAPPLIED && module->inc[i].submodule->augment[j].target) && apply_aug(&module->inc[i].submodule->augment[j], unres)) {
+            if ((module->inc[i].submodule->augment[j].flags & LYS_NOTAPPLIED) && apply_aug(&module->inc[i].submodule->augment[j], unres)) {
                 return EXIT_FAILURE;
             }
         }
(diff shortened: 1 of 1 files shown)


src/tree_schema.c
```diff
@@ -4005,7 +4005,7 @@ apply_aug(struct lys_node_augment *augment, struct unres_schema *unres)
     unsigned int u;
     uint8_t *v;
     struct lys_ext_instance *ext;
-
+
```
michalvasko: Please no useless whitespaces.

src/tree_schema.c
```diff
@@ -4454,11 +4454,10 @@ lys_make_implemented_r(struct lys_module *module, struct unres_schema *unres)
 
     for (i = 0; i < module->augment_size; i++) {
         /* apply augment */
-        if (apply_aug(&module->augment[i], unres)) {
+        if ((module->augment[i].flags & LYS_NOTAPPLIED && module->augment[i].target) && apply_aug(&module->augment[i], unres)) {
```
michalvasko: Adding LYS_NOTAPPLIED check is the correct fix but why check the augment target? It must be resolved at this point and if it is not, we want to fail the assert in apply_aug(), so please remove it.

src/tree_schema.c
```diff
@@ -4472,7 +4471,7 @@ lys_make_implemented_r(struct lys_module *module, struct unres_schema *unres)
 
         for (j = 0; j < module->inc[i].submodule->augment_size; j++) {
             /* apply augment */
-            if (apply_aug(&module->inc[i].submodule->augment[j], unres)) {
+            if ((module->inc[i].submodule->augment[j].flags & LYS_NOTAPPLIED && module->inc[i].submodule->augment[j].target) && apply_aug(&module->inc[i].submodule->augment[j], unres)) {
```
michalvasko: The same, please remove the target check.

src/tree_schema.c
```diff
@@ -4454,11 +4454,10 @@ lys_make_implemented_r(struct lys_module *module, struct unres_schema *unres)
 
     for (i = 0; i < module->augment_size; i++) {
         /* apply augment */
-        if (apply_aug(&module->augment[i], unres)) {
+        if ((module->augment[i].flags & LYS_NOTAPPLIED && module->augment[i].target) && apply_aug(&module->augment[i], unres)) {
             return EXIT_FAILURE;
         }
     }
-
```
michalvasko: No reason to remove the empty line so please keep it there (I would let this one slide but since you will be correcting other things... :) ).

src/tree_schema.c
aweast: @michalvasko, I added the target check too because of another edge case we are hitting on certain models; however, we are still trying to truly identify and understand what is going wrong.  I'll remove the target check for now and open a new issue / pull request once we know more about the particular case.

michalvasko: Thank you, that would be best for now. Also, if it can safely be reproduced, I will gladly take a look at your use-case myself.

Issue #528, opened 2018-06-19
Module xxx in another revision already implemented


pfeige:
Hi,

with the modified testcase in this commit https://github.com/frimpler/libyang/commit/8deb9a07ad86e45ecf35c928b8409648f2b3cb4a another issue resolving identityrefs is shown.

### Description

With our model we get a parsing error in libyang that says "Module xxx in another revision already implemented" even though there is only one revision of the module. The issue happens while resolving an identity when libyang invokes the dm_module_clb() of sysrepo (via ctx->data_clb) to obtain a module not yet in context. The issue has been broken down into a test-model that comprises augments, identy-refs and a must condition to one identity. 

The error log in a release build is as follows:

libyang[0]: Module "ident-aug-must-issue-aps" in another revision already implemented.
libyang[0]: Making the augment target module "ident-aug-must-issue-aps" implemented failed.
libyang[0]: Module "ident-aug-must-issue-apst" parsing failed.

In a debug build with assertions enabled the issue shows up earlier as

libyang/src/parser.c:3827: lyp_ctx_add_module: Assertion `!lyp_ctx_check_module(module)' failed.

The problem seems to occur whenever 
1.	two module struct instances with same module name are pointing to the same ctx instance, so that lyp_ctx_add_module() adds two lys_module* pointers with same name but different address to ctx.list[]
2.	the first added module in the list is implemented. 

Then lys_set_implemented() detects “Module xxx in another revision already implemented.” and returns EXIT_FAILURE, which eventually lets lys_parse_path() return a NULL pointer.

### Used workaround

Currently we are using a workaround (https://github.com/frimpler/libyang/commit/ecc5af345c0b84b375555f578662af6092ef49a4) in order to have a succesful parsing for this part of our model. The workaround is just fixing the obvious symptoms and is by no means expected to be the final solution. 

Regards,
Peter

PR #529, opened 2018-06-21
Test case for issue #528

frimpler:
Test case for testing the issue #528, as has been discussed there.

Regards,
Frank (doing it for Peter)


michalvasko:
Hi Frank & Peter,
thanks for the test case.

Regards,
Michal

PR #530, opened 2018-06-22
Issue data_clb when loading schema node not in context

alangefe:
This commit fixes a case where libyang is attempting to merge a data tree, but the target context does not have the schema for the source data being merged into the target.

The fix is to issue the data callback for the missing schema so that the required information can be found and the data can be successfully merged into the target context.

We have included a new unit test that demonstrates the problem and validates the solution.


michalvasko:
Hi Andrew,
nice, thank you.

Regards,
Michal

PR #532, opened 2018-06-26
Fix when dumping Leaf Elements

NoobAruna:
We were accessing child info for a lyd_node_leaf_list type node and it was resulting in a crash. Added a check and it's working fine now.


michalvasko:
Hi,
I thought I have fixed all instances of this problem, seems not. Thanks.

Regards,
Michal

PR #547, opened 2018-07-26
json: print all numeric types as json numbers

rjarry:
Hi,

I noticed that when printing a lyd_node structure in JSON format, `uint64` and `int64` leafs are printed as JSON strings.

Here is an example:

```c
/* example.c */
#include <libyang/libyang.h>

static const char *yang =
        "module example {"
        "       namespace 'urn:example';"
        "       prefix 'ex';"
        "       container c {"
        "               leaf u32 { type uint32; }"
        "               leaf u64 { type uint64; }"
        "       }"
        "}";

int main(void) {
        struct ly_ctx *ctx;
        struct lyd_node *container;

        ctx = ly_ctx_new(NULL, 0);
        module = lys_parse_mem(ctx, yang, LYS_IN_YANG);

        container = lyd_new(NULL, module, "c");
        lyd_new_leaf(container, module, "u32", "123456789");
        lyd_new_leaf(container, module, "u64", "123456789");

        lyd_print_file(stdout, container, LYD_JSON, LYP_FORMAT | LYP_WD_ALL);
}
```

```
localhost:~$ gcc example.c -o example -lyang
localhost:~$ ./example 
{
  "example:c": {
    "u32": 123456789,
    "u64": "123456789"
  }
}
```

Looking at the code I saw this patch that explicitly print 64 bit integers as strings, but it is not mentioned why.

c27114c87f87 ("json printer BUGFIX format of the printed boolean, decimal64, int64 and uint64 values")

The JSON standard says that there is no range limit for numbers (source http://json.org/):

image

I think that it would be better to print all numeric types as JSON numbers. But I could be missing something.


fanchanghu:
Hi,
This is exactly required by [rfc7951](https://tools.ietf.org/html/rfc7951#section-6.1) , so I think current implement is just fine ~
```
A value of the "int64", "uint64", or "decimal64" type is represented
   as a JSON string whose content is the lexical representation of the
   corresponding YANG type as specified in Sections 9.2.1 and 9.3.1 of
   [RFC7950].
```

rjarry:
Hello,

Oh, my bad... I was not aware of RFC 7951.

However, there is no explanation neither as of why this choice was made.

Do you know why it was decided to represent 64bit numbers as JSON strings?

michalvasko:
Hi,
the reason will be that the encoding used is I-JSON, not JSON ([RFC](https://tools.ietf.org/html/rfc7493#section-2.2)).

Regards,
Michal

rjarry:
Ok, I see. Thanks Michal.

Issue #538, opened 2018-07-11
Libyang returns error as "Too many instances ..." while giving same element as selection and match in request


Prasann1506:
Hello,
When we send rpc request for query operation in which "**name**" field is given as both selection and content-match then reply  output is coming as **"Too many instances of "name" in "rule""**

This is netconf module <get> operation 
When netconf receive this request, we call libyang API **lyd_parse_xml()**  to parse the request message.
But libyang is giving the error **LYVE_TOOMANY i.e. "Too many instances"**
**We think this is a valid scenario and libyang should not return as error.**
Please help to check this issue.

**Test Report:**
![image](https://user-images.githubusercontent.com/35322401/42555212-1f4f230e-8505-11e8-8f03-700498ad496e.png)

Thanks,
Prasann

Prasann1506:
LYD_OPT_GET | LYD_OPT_STRICT

Prasann1506:
This is

Prasann1506:
Yes, this is our own NETCONF server.
Below is our option for API lyd_parse_xml()
Option value : **LYD_OPT_GET | LYD_OPT_STRICT**
is there any issue in this ? Please suggest.

michalvasko:
Hi,
could you put some printer before this `lyd_parse_xml()` call and send also the data tree that is being parsed?

Regards,
Michal

PR #549, opened 2018-07-27
resolve BUGFIX "Too many instances .." is reported for rpc output

fanchanghu:
Hi,
I met the same issue with #538 , and I'm trying to fix it, can you see if my pull request is adequate for this? 

@michalvasko

### DIFF
--- a/src/resolve.c
+++ b/src/resolve.c
@@ -2611,7 +2611,7 @@ resolve_partial_json_data_nodeid(const char *nodeid, const char *llist_value, st
     struct lyd_node_leaf_list *llist;
     const struct lys_module *prev_mod;
     struct ly_ctx *ctx;
-    const struct lys_node *ssibling;
+    const struct lys_node *ssibling, *sparent;
     struct lys_node_list *slist;
     struct parsed_pred pp;
 
@@ -2659,6 +2659,19 @@ resolve_partial_json_data_nodeid(const char *nodeid, const char *llist_value, st
     goto parse_predicates;
 
     while (1) {
+        /* find the correct parent schema node for rpc */
+        sparent = (start && start->parent) ? start->parent->schema : NULL;
+        if (sparent && (sparent->nodetype == LYS_RPC || sparent->nodetype == LYS_ACTION)) {
+            if (options & LYD_PATH_OPT_OUTPUT) {
+                for (sparent = sparent->child; sparent && sparent->nodetype != LYS_OUTPUT; sparent = sparent->next);
+            } else {
+                for (sparent = sparent->child; sparent && sparent->nodetype != LYS_INPUT; sparent = sparent->next);
+            }
+            if (!sparent) {
+                free(pp.pred);
+                return last_match;
+            }
+        }
         /* find the correct schema node first */
         ssibling = NULL;
         while ((ssibling = lys_getnext(ssibling, (start && start->parent) ? start->parent->schema : NULL, prev_mod, 0))) {


fanchanghu:
In addition, This issue is produced  when there are nodes with same name in rpc input and output, and node in input is wrongly used for building output data tree.

src/resolve.c
```diff
@@ -2659,6 +2659,19 @@ resolve_partial_json_data_nodeid(const char *nodeid, const char *llist_value, st
     goto parse_predicates;
 
     while (1) {
+        /* find the correct parent schema node for rpc */
+        sparent = (start && start->parent) ? start->parent->schema : NULL;
```
Prasann1506: sparent : This pointer is not declared

fanchanghu: oh... I forgot it when I copy the code. It should be fixed now.

fanchanghu: I just checked your issue again, it seems that is's not same with mine, and may not be fixed by this request. Sorry for that...

michalvasko:
Hi,
firstly, I do not think mixing `input` and `output` nodes in one data tree is a good idea. But I agree that it should be handled some way.

Could you maybe provide the steps that you are creating the data tree with? Did this pull request solve your issue?

Regards,
Michal

fanchanghu:
Hi,
the module:
[container-list-leaf.yang.txt](https://github.com/CESNET/libyang/files/2248833/container-list-leaf.yang.txt)

the code:
```
#include <stdlib.h>
#include <stdio.h>
#include <libyang/libyang.h>

#define assert(x) if(!(x)){printf("assert fail: " #x "\n");return 1;}else{printf("assert success: " #x "\n");}

int main()
{
    struct ly_ctx *ctx = NULL;
    const struct lys_module *m = NULL;
    struct lyd_node *req = NULL, *root = NULL, *d = NULL;
    int rc = 0;

    ctx = ly_ctx_new(NULL, LY_CTX_NOYANGLIBRARY);
    m = lys_parse_path(ctx, "container-list-leaf.yang", LYS_IN_YANG);
    assert(m)
    req = lyd_parse_path(ctx, "container-list-leaf.xml", LYD_XML, LYD_OPT_RPC, NULL);
    assert(req)

    root = lyd_new_path(NULL, ctx, "/ListLeaf:setdata", NULL, LYD_ANYDATA_CONSTSTRING, 0);
    assert(root)
    d = lyd_new_path(root, ctx, "/ListLeaf:setdata/interfaces", NULL, LYD_ANYDATA_CONSTSTRING, LYD_PATH_OPT_OUTPUT);
    assert(d)
    d = lyd_new_path(root, ctx, "/ListLeaf:setdata/interfaces/ifEntry[ifDescr='thisisone']", NULL, LYD_ANYDATA_CONSTSTRING, LYD_PATH_OPT_OUTPUT);
    assert(d)
    d = lyd_new_path(root, ctx, "/ListLeaf:setdata/interfaces/ifEntry[ifDescr='thisisone']/ifMtu", (void*)"2", LYD_ANYDATA_CONSTSTRING, LYD_PATH_OPT_OUTPUT);
    assert(d)
    d = lyd_new_path(root, ctx, "/ListLeaf:setdata/interfaces/ifEntry[ifDescr='thisisone']/ifIndex", (void*)"2", LYD_ANYDATA_CONSTSTRING, LYD_PATH_OPT_OUTPUT);
    assert(d)

    rc = lyd_validate(&root, LYD_OPT_RPCREPLY, NULL);
    assert(!rc)

    lyd_free(root);
    ly_ctx_destroy(ctx, NULL);

    printf("finish\n");

    return 0;
}

```
Then libyang will report something like:
```
libyang[0]: Too many instances of "interfaces" in "output". (path: /ListLeaf:setdata/interfaces)
```

michalvasko:
Hi,
I think it should be fixed now.

Regards,
Michal

PR #556, opened 2018-08-02
Fix warnings in C++ bindings

jktjkt:
The extra semicolons are not compatible with -pedantic, and the missing
`override` keyword breaks `-Wsuggest-override` on GCC 8.


michalvasko:
Hi,
thanks, merged.

Regards,
Michal

PR #569, opened 2018-08-08
Fix dangling parsing_sub_modules pointer

alangefe:
This change fixes a dangling pointer and memory leak introduced by the refactor in 78ef03a.

In 78ef03a, the call to `lyp_check_circmod_pop()` (under the `error` label) was moved to after the early `if (!module)` return. This meant that if `module` had been freed above due to an error (see line 2637) and we entered the early return, `lyp_check_circmod_pop()` would not be called. `ctx->models->parsing_sub_modules` would now contain a danging pointer to a module whose memory had been freed. Subsequent attempts to use the list could result in undefined behavior, including a segmentation fault.

The fix is to pop the module off the list before we free it.


michalvasko:
Hi Andrew,
you are right, I have missed this. Just wanted to unify that code for YIN and YANG parser. Thanks.

Regards,
Michal

PR #585, opened 2018-08-22
C++: allocate std::vector on stack instead of heap

mislavn:
To simplify the C++ bindings, std::vector's are allocated on the stack.

C++ and Python examples are updated.


michalvasko:
Hi,
is some other pull request needed for it to pass? If so, I have merged the other one, so you can restart the tests. If not, then please fix the compilation issues.

Regards,
Michal

mislavn:
Forgot to update the unit tests, will do it now.

PR #586, opened 2018-08-23
C++: Wrap all symbols in the `libyang` namespace

jktjkt:
Using generic names such as `Context` or `Deleter` has a huge potential for collisions. If two libraries in one stack (such as libyang and sysrepo)  use conflicting names, then one has an interesting debugging session ahead.

This patch wraps all C++-specific symbols in the libyang namespace. This also required making the `S_Foo` `#define`s into something scopeable, and C++'s way of doing this is via the `using` keyword.

While changing this, I made the tests and examples follow a more modern pattern of calling std::make_shared instead of a std::shared_ptr(new X).


jktjkt:
I assume that this will require some matching changes on the Python side of these bindings -- help is appreciated. I only care about the C++ side here.

mislavn:
@jktjkt I intended to the the namespace changes next week, guess I'll do it for Sysrepo then.

I'll take a look at the Python bindings.

michalvasko:
Hi,
so this can be merged now or it will break Python bindings and I should wait for @mislavn to finish those changes, too?

Regards,
Michal

mislavn:
SWIG complains a lot, this will take some time.

mislavn:
I pushed the commits to the forked git repository: https://github.com/sartura/libyang/tree/cpp-namespace

You can cherry pick the commits.

mislavn:
https://github.com/CESNET/libyang/pull/586/commits/a0ce0e4ab56f77cd0798cea7e6a4ec7a7f2c30b1 will generate SWIG errors, I was going to fix this next week.

jktjkt:
Thanks for your work on this. I gave it a quick look, but I have no clue on how to fix it properly :(. I noticed that a [random stackoverflow post](https://stackoverflow.com/a/42780893/2245623) suggests to put the `%shared_ptr foo::bar` stanzas prior to including the corresponding C++ header, but doing that quickly exploded.

I think that I'll try to continue my SWIG-ignorant life for a bit longer. I'm looking forward to your patches!

mislavn:
@jktjkt I'm afraid `std::make_shared` can't be used in the Python SWIG interface, because `SWIG_NewPointerObj` needs a heap allocated object.

jktjkt:
OK, feel free to skip my last commit. I just saw a pattern which can be made less verbose in its then-current form. If it cannot be done, then it cannot be done. And thanks for looking into this!

jktjkt:
@mislavn, did you have a chance on fixing the Python SWIG bits here?

mislavn:
To fix this simply remove the commit https://github.com/CESNET/libyang/pull/586/commits/a0ce0e4ab56f77cd0798cea7e6a4ec7a7f2c30b1 from the pull request.

jktjkt:
I think that I tested that before and it didn't work (the `auto` will still deduce a `std::shared_ptr<T>`, the `std__make_shared` is just a detail there), but I'll give the CI a run here and post back. Thanks.

jktjkt:
Aaargh, now I see the problem -- I fat-fingered some vim macros. Doh, sorry.

jktjkt:
Triggering the CI now.

mislavn:
I don't get `Warning 341: The 'using' keyword in type aliasing is not fully supported yet.` on my local build.

Is it a problem if we switch back to "#define"?

jktjkt:
I can confirm that my attempt at using `std::make_shared` was obviously a complete nonsense, sorry for that -- I kinda missed that the bindings need the resulting `shared_ptr` itself on the heap. My bad.

The build succeeds on my env with SWIG 3.0.12. I see that the CI env is limited to Ubuntu 14.04 which has SWIG imported from an external `.deb` file by @michalvasko since aad3998d67dde444d489453d4e183de2596aab23. That version does not match any release in Ubuntu repositories, and the commit doesn't say where did it come from. Ubuntu has a newer LTS release with a newer SWIG, but we cannot use that because of Travis limitations.

We *cannot* use `#define S_Deleter` because C preprocessor puts everything into the global scope. It would improve the situation a bit (stuff would not result in undefined behavior due to ODR violations in the linker), but any user's translation unit which attempts to `#include` both sysrepo-cpp and libyang-cpp would explode.

I'll try adding some `using namespace libyang;` into SWIG's global scope so that the C++ compiler can at least produce these bindings. A real fix is a newer SWIG, though.

mislavn:
If you want we can set a higher SWIG version requirement.

jktjkt:
> If you want we can set a higher SWIG version requirement.

Yes please. I would prefer if you/Michal took care of this because I have no idea how the existing .deb was built.

The building now succeeds, but the Python tests fail on Travis. They work locally.

jktjkt:
@mislavn , @michalvasko : can we get an updated SWIG, please?

mislavn:
I'm working on it.

jktjkt:
This is now ready to be merged. Many thanks to @mislavn !

Issue #580, opened 2018-08-21
lyb: warnings when building for ARM 32bit


jktjkt:
```
/home/jkt/work/prog/_build/br-cfb/build/libyang-custom/src/parser_lyb.c: In function ‘lyb_parse_val_2’:
/home/jkt/work/prog/_build/br-cfb/build/libyang-custom/src/parser_lyb.c:569:29: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 3 has type ‘int64_t {aka long long int}’ [-Wformat=]
         sprintf(num_str, "%ld", value->int64);
                           ~~^   ~~~~~~~~~~~~
                           %lld
/home/jkt/work/prog/_build/br-cfb/build/libyang-custom/src/parser_lyb.c:573:29: warning: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 3 has type ‘uint64_t {aka long long unsigned int}’ [-Wformat=]
         sprintf(num_str, "%lu", value->uint64);
                           ~~^   ~~~~~~~~~~~~~
                           %llu
/home/jkt/work/prog/_build/br-cfb/build/libyang-custom/src/parser_lyb.c:585:29: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 3 has type ‘long long int’ [-Wformat=]
         sprintf(num_str, "%ld.%.*ld", value->dec64 / (int64_t)type->info.dec64.div, dig, frac);
                           ~~^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                           %lld
/home/jkt/work/prog/_build/br-cfb/build/libyang-custom/src/parser_lyb.c:585:35: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 5 has type ‘int64_t {aka long long int}’ [-Wformat=]
         sprintf(num_str, "%ld.%.*ld", value->dec64 / (int64_t)type->info.dec64.div, dig, frac);
                               ~~~~^
                               %.*lld
```

michalvasko:
Hi,
should be fixed now.

Regards,
Michal

jktjkt:
Thanks, these are OK now. What remains are some smaller warnings which (I suppose) are tricky to fix in a portable manner):
```
src/parser_json.c: In function ‘lyjson_parse_text’:
src/parser_json.c:153:32: warning: comparison is always true due to limited range of data type [-Wtype-limits]
[...]

PR #588, opened 2018-08-24
JSON: Improve readability of ASCII checks

jktjkt:
- JSON: remove redundant test. We've already checked for backslash in the branch above, so there's no need to check for 0x5c again here.

- JSON: more readable fix of char/signed char/unsigned char
    
@michalvasko  correctly suggested that a mere cast is not terribly readable. This patch ensures that all comparisons are done on unsigned chars which match one's understanding of ASCII codes. Now that everything is positive, the condition can be simplified to check just for values `< 0x20`.

I am not introducing the `const unsigned char ascii` into `parser_json.c` because that code can consume more than one character in one loop iteration. If there was another variable, then one would have to be extra careful not to skip over its re-assignment after bumping the index.

refs #580

### DIFF
--- a/src/parser_json.c
+++ b/src/parser_json.c
@@ -150,7 +150,7 @@ lyjson_parse_text(struct ly_ctx *ctx, const char *data, unsigned int *len)
             }
             o += r - 1; /* o is ++ in for loop */
             (*len) += i; /* number of read characters */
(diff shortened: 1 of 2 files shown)


michalvasko:
Okay, seems nice, thanks.

Regards,
Michal

PR #589, opened 2018-08-25
Fix lyb bit-handling bugs

alangefe:
This commit fixes a couple of problems with the handling of the `bit` type for the `lyb` file format. Both bugs were only exposed when there were more than 8 values defined for the bit type.

The first bug was an off-by-one error. If the user attempted to set the value at position 0, no values would be set. If the user attempted to set the value at position 1, position 0 would be erroneously set instead, and so forth. I fixed this by correcting the bit shifts in `lyb_print_value()`.

The second bug was due to the fact that `lyb_parse_val_1()` continued to read the encoded data byte for positions 0 through 7 for all subsequent positions. Thus if the user had set the value at position 0, it would appear that the values at position 8, position 16, and so forth were also set. I fixed this by taking an offset into account when reading from the encoded data in the loop.

This commit also contains a test modification to verify the fixes. Without the source code patches, a set of `position-0 position-5` will result in a data tree with `position-0` not set at all, with `position-4` set instead of `position-5`, and `position-4` erroneously mirrored into `position-12`.

### DIFF
--- a/src/parser_lyb.c
+++ b/src/parser_lyb.c
@@ -382,11 +382,10 @@ lyb_parse_val_1(struct ly_ctx *ctx, struct lys_type *type, LY_DATA_TYPE value_ty
         for (i = 0; i < type->info.bits.count; ++i) {
             if (i % 8 == 0) {
                 /* read another byte */
-                ret += (r = lyb_read(data, &byte, sizeof byte, lybs));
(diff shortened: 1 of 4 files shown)


michalvasko:
Hi Andrew,
thanks for the fixes and a test!

Regards,
Michal

PR #590, opened 2018-08-28
python: in examples use LYD_LYB format

mislavn:

### DIFF
--- a/swig/python/examples/process_tree.py
+++ b/swig/python/examples/process_tree.py
@@ -23,10 +23,22 @@ if module is not None:
 else:
     module = ctx.load_module("turing-machine")
 
+node = None
 try:
-    node = ctx.parse_data_path("/etc/sysrepo/data/turing-machine.startup", ly.LYD_XML, ly.LYD_OPT_CONFIG)
+    if node is None : node = ctx.parse_data_path("/etc/sysrepo/data/turing-machine.startup", ly.LYD_LYB, ly.LYD_OPT_CONFIG)
 except Exception as e:
     print(e)
+try:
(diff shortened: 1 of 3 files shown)


michalvasko:
Hi Mislav,
I do not use these examples much myself so am fine with this change but I think it would be best to read the format from the `cmake` variable so that they work even if you change it to something other than the default.

Regards,
Michal

mislavn:
Hi Michal,

in that case I would like to make something like this:
```
node = None
try:
    if node is None : node = ctx.parse_data_path("/etc/sysrepo/data/turing-machine.startup", ly.LYD_JSON, ly.LYD_OPT_CONFIG)
except Exception as e:
    print(e)
try:
    if node is None : node = ctx.parse_data_path("/etc/sysrepo/data/turing-machine.startup", ly.LYD_XML, ly.LYD_OPT_CONFIG)
except Exception as e:
    print(e)
try:
    if node is None : node = ctx.parse_data_path("/etc/sysrepo/data/turing-machine.startup", ly.LYD_LYB, ly.LYD_OPT_CONFIG)
except Exception as e:
    print(e)
```

But `lyd_parse_path` never return NULL or set's the `ly_vecode`. Do you know what parseroptions I need to use so that libyang fails parsing the wrong format?

michalvasko:
HI Mislav,
LYB should not be possible to parse as XML with the most recent *libyang*, so please try again.

Regards,
Michal

mislavn:
Thanks, it works now.

michalvasko:
Okay, thanks!

Regards,
Michal

PR #597, opened 2018-08-31
python: implemented set_module_imp_clb

mislavn:

### DIFF
--- a/swig/python/yang.i
+++ b/swig/python/yang.i
@@ -4,11 +4,14 @@
     extern "C" {
         #include "libyang.h"
         #include "tree_data.h"
+        #include "tree_schema.h"
     }
 %}
 
 %include exception.i
 %include <std_except.i>
+%include <std_pair.i>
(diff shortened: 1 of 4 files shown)


swig/python/yang.i
```diff
@@ -21,8 +24,78 @@
...
+    (void)free_module_data;
+    auto pair = ctx->ly_module_imp_clb(mod_name, mod_rev, submod_name, sub_rev, ctx->private_ctx);
+    *format = pair.second;
+    return pair.first;
+}
 %}
 
+%extend Context {
+
+    void set_module_imp_clb(PyObject *clb, PyObject *user_data) {
```
rkrejci: The function tries to handle NULL `user_data`, but when I was testing the python version of this function, it requires 2 parameters, so the caller has to set `user_data` as `None` explicitely. Is it possible to cover it and make the default value `None`?

mislavn: yes, give me a minute.

PR #599, opened 2018-09-04
swig: set min version to 3.0.12

mislavn:
Because of the keyword "using" in C++ which will replace #define SWIG
requires min version 3.0.12


mislavn:
This PR is needed for https://github.com/CESNET/libyang/pull/586

PR #600, opened 2018-09-04
Module filepath on BSD

snar:
To record filename module was loaded from, libyang uses OS-dependent methods
to recover filename from open file descriptor. As these methods are defined for Linux
and Apple builds only, file path on BSD systems is either not recorded at all (when 
lys_parse_fd called from context.c) or recorded as relative path (when called from 
tree_schema.c). 
This patch is an attempt to implement OS-independent way of recording absolute
path name of module file if os-dependent methods failed or were unavailable.

PS: CMakeList and tests.h patches required to cover the case when build working
directory has different realpath, f.e., you build under /home/user when /home is a symlink
to /usr/home.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -884,8 +884,11 @@ ly_ctx_load_localfile(struct ly_ctx *ctx, struct lys_module *module, const char 
 
     if (!result->filepath) {
         char rpath[PATH_MAX];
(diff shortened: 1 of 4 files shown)


michalvasko:
Hi,
okay, I have no problem with changing it this way.

Regards,
Michal

PR #602, opened 2018-09-06
Implement callbacks in the C++ API

jktjkt:
## Lambdas and std::function for schema callbacks

This patch series makes it possible to use C++-style API with lambdas or `std::function` as callbacks for requesting module implementations/imports. It changes the public API (trivially), so I had to adapt libnetconf2 and Netopeer2-server.

There are three commits:

###  C++-ish way of reading module data via callbacks
    
This was not possible to do when using C++ bindings because the `ly_ctx` was private. Also, while C-style freestanding functions certainly work, this API can be used more easily I think.
    
The Python bindings already contained support for something like this. I do not know if it's possible to reimplement Python's usual semantics on top of the API introduced in this patch.

### Change constness of the data returned by module implementation callbacks
    
libyang itself won't modify these data, so a `const char *` is a better match. That will allow returning, e.g., string literals or C++'s `std::string::c_str` without any extra deleter.
    
The cost of this patch is an extra cast which is needed when calling a deleter. If a deleter is used, then quite obviously that data cannot be const.

### Pass the original user_data to the deleter function

The callback which is used for feeding libyang with module source code accepts a generic void* for contextualization. The callback itself can, when called, specify another function which can be used for deleting the result. The function signature matches that of `free`. However, when this custom deleter is called, the user_data context is not passed.
    
This means that the first callback cannot really provide any contextualization. It could still swap a generic `free` for something else, but that "something else" again lacks any context because it can only be a free-standing function.
    
In the C++ API for this, I will need a real context even for the custom deleter, otherwise it is not possible to use basic features such as a `std::function`.
    
The Python bindings appear to ignore the custom deleter. I have no clue if that behavior is correct or if it leaks.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -904,7 +904,7 @@ ly_ctx_load_sub_module_clb(struct ly_ctx *ctx, struct lys_module *module, const 
     struct lys_module *mod = NULL;
     char *module_data = NULL;
     LYS_INFORMAT format = LYS_IN_UNKNOWN;
-    void (*module_data_free)(void *module_data) = NULL;
+    void (*module_data_free)(void *module_data, void *user_data) = NULL;
 
     ly_errno = LY_SUCCESS;
     if (module) {
@@ -928,7 +928,7 @@ ly_ctx_load_sub_module_clb(struct ly_ctx *ctx, struct lys_module *module, const 
         }
 
         if (module_data_free) {
-            module_data_free(module_data);
+            module_data_free(module_data, ctx->imp_clb_data);
         }
     }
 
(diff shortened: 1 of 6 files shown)


mislavn:
Python bindings can use virtual functions for callbacks or a custom function in the sysrepo.i interface file, so you don't need to be concerned about Python bindings in this pull request.

michalvasko:
Hi,
is this ready to be merged?

Regards,
Michal

jktjkt:
Yes, this is ready. All three pull requests (this one, libnetconf2 and netopeer2) should be merged together.

PR #607, opened 2018-09-08
fix bugs and add java support

9606:
### 1.fix segmentation error caused by methods such as parse_data_path in some cases
When I run the process_notif_tree.cpp program, the following error message is displayed:

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff6e9d113 in lyd_parse_data_ (ctx=0x620c40,
    data=0x40a230 "<busy xmlns=\"cpu-notif\">\n    <usage>80</usage>\n</busy>", format=LYD_XML, options=64, ap=0x7fffffffe050)
    at /root/libyang-0.15-r1/src/tree_data.c:966
966                     if (iter->parent) {

In some cases, you need to transfer the fifth parameter to the methods such as lyd_parse_mem. Otherwise, the fifth parameter may be an invalid pointer.

lyd_parse_mem method defined in tree_data.h as following:

\/\*\*
 \* @ param[in] ctx Context to connect with the data tree being built here.
 \* @ param[in] data Serialized data in the specified format.
 \* @ param[in] format Format of the input data to be parsed.
 \* @ param[in] options Parser options, see @ref parseroptions.
 \* @ param[in] ... Variable arguments depend on \p options. If they include:
 \*　　　　　　　\- \#LYD_OPT_DATA:
 \*　　　　　　　\- \#LYD_OPT_CONFIG:
 \*　　　　　　　\- \#LYD_OPT_GET:
 \*　　　　　　　\- \#LYD_OPT_GETCONFIG:
 \*　　　　　　　\- \#LYD_OPT_EDIT:
 \*　　　　　　　　 \- no variable arguments expected.
 \*　　　　　　　\- \#LYD_OPT_RPC:
 \*　　　　　　　\- \#LYD_OPT_NOTIF:
 \*　　　　　　　　\- struct lyd_node *data_tree - additional data tree that will be used
 \*　　　　　　　　when checking any "when" or "must" conditions in the parsed tree that require
 \*　　　　　　　　some nodes outside their subtree. It must be a list of top-level elements!
 \*　　　　　　　\- \#LYD_OPT_RPCREPLY:
 \*　　　　　　　　\- const struct ::lyd_node *rpc_act - pointer to the whole RPC or action operation data
 \*　　　　　　　　tree (the request) of the reply.
 \*　　　　　　　　\- const struct ::lyd_node *data_tree - additional data tree that will be used
 \*　　　　　　　　when checking any "when" or "must" conditions in the parsed tree that require
 \*　　　　　　　　some nodes outside their subtree. It must be a list of top-level elements!
 \* \@ return Pointer to the built data tree or NULL in case of empty \p data. To free the returned structure,
 \*　　　　use lyd_free(). In these cases, the function sets #ly_errno to LY_SUCCESS. In case of error,
 \*　　　　#ly_errno contains appropriate error code (see #LY_ERR).
 \*/
struct lyd_node *lyd_parse_mem(struct ly_ctx *ctx, const char *data, LYD_FORMAT format, int options, ...);

### 2.add method to get definition of the instantiated extension
In some cases, definition of the instantiated extension need to be obtained.
### 3.add swig java support and add java test and README.md
Add swig java support and add some test cases.All test cases have been passed.


swig/java/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
+set(JAVA_SWIG_BINDING libyangJava)
```
mislavn: Please rename the library to "libyang", for that you will also need to remove the line:
`set_target_properties(${SWIG_MODULE_${JAVA_SWIG_BINDING}_REAL_NAME} PROPERTIES PREFIX "")`,
rename the interface `libyangJava.i` to `libyang.i` and the load library name in examples and tests.

9606: but if I modify the first line like this:
`set(JAVA_SWIG_BINDING libyang)`
and remove the line:
`set_target_properties(${SWIG_MODULE_${JAVA_SWIG_BINDING}_REAL_NAME} PROPERTIES PREFIX "")`
and rename the interface `libyangJava.i` to `libyang.i`
the target java so file will be named `liblibyang.so`, it sounds strange 

if I don't remove the line:
`set_target_properties(${SWIG_MODULE_${JAVA_SWIG_BINDING}_REAL_NAME} PROPERTIES PREFIX "")`
the target java so file will be named `libyang.so`, it will conflict with the libyang.so in the src/build directory.

mislavn: You are right, I forgot to run "make install". Then can you change it so that `System.loadLibrary` uses libyangJava name?

9606: The name of the generated .so file now is libyangJava.so, the code for invoking the dynamic link library libxxx.so in Java is as follows:
`System.loadLibrary(xxx);`

swig/java/CMakeLists.txt
```diff
@@ -19,6 +19,19 @@ set_target_properties(${SWIG_MODULE_${JAVA_SWIG_BINDING}_REAL_NAME} PROPERTIES P
 swig_link_libraries(${JAVA_SWIG_BINDING} ${JNI_LIBRARIES} libyang-cpp)
 
 add_custom_command(TARGET ${JAVA_SWIG_BINDING} POST_BUILD COMMAND "${Java_JAVAC_EXECUTABLE}" *.java -d . COMMAND "${Java_JAR_EXECUTABLE}" -cvf libyang.jar com)
```
9606: I think it would be better to change the package name to org.cesnet. What do you think?

swig/java/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+message(STATUS "JNI_INCLUDE_DIRS: ${JNI_INCLUDE_DIRS}")
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+
+set(CMAKE_SWIG_FLAGS "-c++")
+set(CMAKE_SWIG_FLAGS "-I${PROJECT_SOURCE_DIR}")
+set(CMAKE_SWIG_OUTDIR ${CMAKE_CURRENT_BINARY_DIR})
+
+file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/com/cesnet")
+
+set_source_files_properties(${JAVA_SWIG_BINDING}.i PROPERTIES CPLUSPLUS ON PREFIX "" SWIG_FLAGS "-package;com.cesnet")
```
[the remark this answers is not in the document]
mislavn: I'm fine with the change.

swig/java/CMakeLists.txt
mislavn: OK.

michalvasko:
Hi,
@9606 i am a bit confused here. Do you want to rename the package in this pull request or in another? In other words, can this be merged or are there some changes you want to add first?

Regards,
Michal

PR #615, opened 2018-09-13
Libyang shall use standard fileno(3) instead of unportable _fileno.

snar:
... else it breaks swig build on at least freebsd.


michalvasko:
Hi,
right, seems reasonable.

Regards,
Michal

PR #617, opened 2018-09-15
json parser BUGFIX wrong use input for action

fanchanghu:
Hi,
When parsing json data of action output, if there is a node with same name in input and output, we need use output schema node not input, like rpc.


michalvasko:
Hi,
right, thanks.

Regards,
Michal

PR #618, opened 2018-09-17
Fix multiple must deviations to same node

alangefe:
A segmentation fault can occur when multiple deviations add must deviations to the same target node. The sequence of events is as follows:

1. Some number of must deviations have already been parsed, and the module's deviation array and the target node are populated to point an array of `struct lys_restr`.
2. An additional must deviation is parsed, deviating the same target node, and the `ly_realloc()` in `yang_check_deviate()` (for YANG parsing) or `yin_fill_deviation()` (for YIN parsing) cannot extend the existing buffer any further, and so it frees the original buffer and allocates a new, larger buffer.
3. Any existing must deviations on the module that point to that same target node now point to memory freed by `ly_realloc()`. Attempts to use that memory will result in undefined behavior, including a segmentation fault.

I've added some additional leafs and must deviations to those leafs in the test models that demonstrate this problem. `test_parse_print` and `test_deviation` will crash with a segmentation fault given these models.

(Unrelated to these changes, I also updated leafs 9-10 of all.yang to match all.yin. These nodes were changed in all.yin in c27114c, but all.yang was never updated to match.)

I've also implemented a possible fix. If the deviation is an `add`, the module's deviation now contains a shallow copy of the `struct lys_restr` array pointed to by the target node, instead of them pointing to the same memory. (This meant I had to add a shallow free of this copy when the module was cleaned up. See `lys_sub_module_remove_devs_augs()`.)

Another possible solution, which I did not implement, might be for the parser to revisit all siblings of the module's deviations when `ly_realloc()` freed the original memory. We would also have to keep track of what offset each module deviation pointed to within the larger target node array.

### DIFF
--- a/src/parser_yang.c
+++ b/src/parser_yang.c
@@ -1878,8 +1878,6 @@ yang_check_deviate_must(struct lys_module *module, struct unres_schema *unres,
         LY_CHECK_ERR_GOTO(!must, LOGMEM(ctx), error);
         *trg_must = must;
         memcpy(&(*trg_must)[*trg_must_size], deviate->must, deviate->must_size * sizeof *must);
-        free(deviate->must);
-        deviate->must = &must[*trg_must_size];
         *trg_must_size = *trg_must_size + deviate->must_size;
         erase_must = 0;
     } else if (deviate->mod == LY_DEVIATE_DEL) {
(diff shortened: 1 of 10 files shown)


michalvasko:
Hi,
seems fine to me, thanks a lot.

Regards,
Michal

Issue #608, opened 2018-09-11
About the order of nodes in rpc/action output/input


fanchanghu:
Hi,
It seems libyang requires all nodes in rpc/action output/input are in same order as defined in schema, this may be right for XML encoding, but may not be right for JSON encoding, the RFC 7951 has some descriptions about the order:
[section-5.4](https://tools.ietf.org/html/rfc7951#section-5.4):
```
   Unlike the XML encoding, where list keys are required to precede any
   other siblings within a list entry and appear in the order specified
   by the data model, the order of members in a JSON-encoded list entry
   is arbitrary because JSON objects are fundamentally unordered
   collections of members.
```
[section-7](https://tools.ietf.org/html/rfc7951#section-7):
```
   o  The order of JSON object members is never relied upon.
```
So, I think the order for output/input data tree could be arbitrary if it is JSON-encoded.

michalvasko:
Hi,
the order should no longer be required and your example program should work now but let me just note that, strictly speaking, there is no standardized way of encoding RPCs/actions/rpc-replies into JSON (point me to some text if you know about one).

Regards,
Michal

fanchanghu:
Hi,
Yes, you are right, there is no explicit specification about how to encode rpc/action as JSON, 
but the RFC 7951 says its encoding rules are identical for all types of data trees, 
we could naturally assume that there is no order restriction for rpc/action. :)
```
   ...
   The encoding rules are identical for all types of data trees, i.e.,
   configuration data, state data, parameters of RPC operations,
   actions, and notifications.
```

fanchanghu:
And, maybe a problem for your change: if I print the data tree which is loaded from a JSON text as a XML text, the output should be sorted. (but it is not now)
```
lyd_print_file(stdout, dt2, LYD_XML, LYD_OPT_RPCREPLY);
```

fanchanghu:
Hi,
It seems it still not work for action case, can you check following code, thanks.
```
#include <stdio.h>
#include <stdlib.h>
#include <assert.h>

#include "libyang/libyang.h"

static const char *yang =
        R"(
module ListLeaf {
    namespace "http://example.net/zte/list/leaf";
    prefix ListLeaf;
    yang-version "1.1";

    container ListLeaf{
       leaf id{type uint32;}
    action setdata{
        input{
            container interfaces {
                list ifEntry {
                    key ifIndex;
                    leaf ifIndex {
                        type uint32;
                    }
                    leaf ifDescr {
[...]

PR #620, opened 2018-09-17
json parser BUGFIX action input/output need to be ordered

fanchanghu:
Hi,
I have tried to fix the #608 myself, the request could work for me.


michalvasko:
Hi,
okay, this is also a valid approach.

Regards,
Michal

PR #621, opened 2018-09-18
json parser BUGFIX list/leaf-list in input/output do not have to be ordered

fanchanghu:
Hi,
It seems #608 is still not fixed fully, multi instances of list or leaf-list in input/output will cause an error when parsing json data tree, hope it could be fully fixed this time...


michalvasko:
Hi,
okay, makes sense. Thanks.

Regards,
Michal

PR #629, opened 2018-10-05
Fix big endian issues writing and reading .LYB files

gcampbell512:
The print and parse routines which handle .LYB files assume a little endian
CPU. This means libyang is failing when it is compiled for a big endian
target.

For example consider the code:
 lyb_write(out, (uint8_t *)&mod_count, 2, lybs);

mod_count is an integer and on a typical 32 bit big endian target its
format in memory is (byte0(MSB) byte1 byte2 byte3(LSB)) meaning
if mod_count has a value of 1 the first three bytes of mod_count
are zero, and as the routine only write the first two bytes, it writes
out 0. This effectively corrupts the .LYB file.

This patch modifies the routines to write and read .LYB files to use
endian agnostic methods to access data in a .LYB file.

### DIFF
--- a/src/parser_lyb.c
+++ b/src/parser_lyb.c
@@ -17,6 +17,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <inttypes.h>
+#include <endian.h>
 
 #include "libyang.h"
 #include "common.h"
@@ -32,7 +33,7 @@ lyb_read(const char *data, uint8_t *buf, size_t count, struct lyb_state *lybs)
 {
     int ret = 0, i, empty_chunk_i;
     size_t to_read;
-    LYB_META meta;
+    uint8_t meta_buf[LYB_META_BYTES];
 
     assert(data && lybs);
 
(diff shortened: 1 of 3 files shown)


src/parser_lyb.c
```diff
@@ -107,13 +108,31 @@ lyb_read_number(uint64_t *num, size_t bytes, const char *data, struct lyb_state
...
+        else if (bytes == 4)
+        {
+            *(uint32_t *)num = le32toh(*(uint32_t *)num);
+        }
+        else if (bytes == 8)
+        {
+            *(uint64_t *)num = le64toh(*(uint64_t *)num);
+        }
+    }
+
```
michalvasko: Please put { on the same lines as if / else if, thanks.

michalvasko:
Hi,
thanks a lot for looking into this, we do not really have access to any big-endian machines. Just fix the small formatting problem and I will be happy to merge it.

Regards,
Michal

gcampbell512:
Hi,

I have simplified the code so that the code with the bad bracing is no longer present.

My company actually builds software for embedded products with both big and little endian CPU's. It took me a few days to figure out how to run the tests actually on our products to verify the code changes.

Regards Graeme

gcampbell512:
Hi,

Did my last patch fail to build or did my closing the pull request by mistake cause this issue?

Regards Graeme

michalvasko:
It passed the tests just fine, is it ready to be merged? 

Regards,
Michal

gcampbell512:
Hi Michal,

Yes.

Regards Graeme

PR #641, opened 2018-10-23
Remove dependency on a C11 capable compiler

rwestphal:
Since commit 404aea2a441, libyang uses the _Generic statement which is
only available on C11 capable compilers. This dependency however is
problematic since lots of systems are stuck with old compilers which
don't support the C11 standard (e.g. CentOS 6, some embedded systems,
etc). Further, the fact that _Generic is used on a public header means
that the dependency on a C11 capable compiler also applies to all software
using libyang.

Solve this problem by using the __builtin_types_compatible_p()
built-in function instead of _Generic when the compiler is not C11
capable. While __builtin_types_compatible_p() is a GNU GCC extension,
it's also implemented by other widely used compilers like clang.


michalvasko:
Okay, no problem.

Regards,
Michal

PR #650, opened 2018-10-30
fix bug when parse action output msg first node is list,issue:643

ZhaoGuangYue1986:
fix bug when parse action output msg first node is list,issue:643

### DIFF
--- a/src/parser_json.c
+++ b/src/parser_json.c
@@ -1398,7 +1398,8 @@ empty:
         if (!result) {
             if (reply_parent) {
                 result = next->child;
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
it should suffice to use the `prev` pointer and the result should be resolved, can you check please?

Regards,
Michal

ZhaoGuangYue1986:
yes if we use the prev pointer this issue can be resolved also , but  i think we should check next->child null first

michalvasko:
Fair enough, it is there now.

Regards,
Michal

PR #652, opened 2018-11-03
Extend the TYPES_COMPATIBLE macro to understand C++

rwestphal:
We had two versions of the LY_TREE_DFS_END macro, one for C and the
other for C++. This commit removes the duplication by extending the
TYPES_COMPATIBLE macro in a way that it can also understand C++, allowing
us to have a single LY_TREE_DFS_END macro for both C and C++.

Also, change the TYPES_COMPATIBLE macro in a way that it can also work
with primitive types and not only structures.


michalvasko:
Alright, thanks!

Regards,
Michal

PR #658, opened 2018-11-08
cmake use only LIBYANG_MAJOR_VERSION for SOVERSION

eqvinox:
SOVERSION determines the SONAME in the final library file, and SONAME
must be in the form "libyang.so.0" without the minor version number
added.  Otherwise, the dynamic linker & ldconfig mis-process the
version.

(This fixes a packaging problems on Debian.)


eqvinox:
=> will see what the Debian people say about the `0.16` SOVERSION, if they refuse it I'll reopen the PR

PR #661, opened 2018-11-09
add yangre.1 to Debian/RedHat packages

eqvinox:
For Debian we can just list the man directory, for RedHat we list the
gzipped file.

Also fix a spelling error.  Should've run spellcheck on it before
submitting...

(oops :disappointed:)


rkrejci:
Thanks!

PR #666, opened 2018-11-12
plugins NEW add extension loading by pointer

eqvinox:
When libyang is used in some application (e.g. FRRouting), FRR brings
its own type handlers.  Currently, these need to be split off into a
separate module which libyang loads from its module directory.

Allowing the application to directly register types by pointer addresses
some problems:

* a separate module file needs to be correctly installed and found by
  libyang for the entire thing to work.  This makes it hard to run the
  application without installing it, and doesn't really work at all with
  static linking.

* depending on what the application does with the extension types, the
  code in the module will likely need to match a specific version of the
  application because it will have references to application-specific
  data structures and code.  Not only would this need additional
  safeguarding by the application to prevent crashes resulting from
  changed data structures, it also makes it very difficult to work with
  2 application versions at the same time, as the module name would
  probably collide.

While these issues can be worked around by the application by changing
LIBYANG_USER_TYPES_PLUGINS_DIR, it's far more simple and robust to just
have the application hand the proper ly*_plugin_list array to libyang.

### DIFF
--- a/src/plugins.c
+++ b/src/plugins.c
@@ -138,8 +138,7 @@ cleanup:
 static int
 lytype_load_plugin(void *dlhandler, const char *file_name)
 {
-    struct lytype_plugin_list *plugin, *p;
-    uint32_t u, v;
+    struct lytype_plugin_list *plugin;
     char *str;
 
 #ifdef STATIC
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi,
seems just fine, thanks.

Regards,
Michal

Issue #679, opened 2018-12-05
unit test failure (test_yang_data) on big endian systems (0.16.52)


eqvinox:
Something seems to be going wrong on `sparc64`, `ppc64` and `hppa` (all big-endian 64-bit systems) specifically in this unit test.  Logfiles:
https://buildd.debian.org/status/fetch.php?pkg=libyang&arch=sparc64&ver=0.16.52-1&stamp=1544025575&raw=0
https://buildd.debian.org/status/fetch.php?pkg=libyang&arch=ppc64&ver=0.16.52-1&stamp=1544026374&raw=0
https://buildd.debian.org/status/fetch.php?pkg=libyang&arch=hppa&ver=0.16.52-1&stamp=1544027739&raw=0

Unfortunately I don't have any further details on this.  I'm adding the `-V` option to the test run to get more info but that will take a few hours.  But considering that it passes tests on `ppc64el` (little-endian mode) it seems extremely likely to be somehow endianness-related.

eqvinox:
Add: also fails on `powerpc` (32bit big-endian)
https://buildd.debian.org/status/fetch.php?pkg=libyang&arch=powerpc&ver=0.16.52-1&stamp=1544031030&raw=0

eqvinox:
on `powerpc`:
```
28: Test command: /home/equinox/libyang/build/tests/test_yang_data
28: Environment variables: 
28:  LIBYANG_EXTENSIONS_PLUGINS_DIR=/home/equinox/libyang/build/src/extensions
28:  LIBYANG_USER_TYPES_PLUGINS_DIR=/home/equinox/libyang/build/src/user_types
28:  MALLOC_CHECK_=3
28: Test timeout computed to be: 9.99988e+06
28: [==========] Running 5 test(s).
28: [ RUN      ] test_new_path
28: libyang[0]: Invalid value "test" in "first" element. (path: /yang-data:#node/test1/first)
28: libyang[0]: Failed to create node "first" as a child of "test1". (path: /yang-data:#node/test1/first)
28: libyang[0]: Schema node not found. (path: /yang-data:test1)
28: [       OK ] test_new_path
28: [ RUN      ] test_new_yangdata
28: [       OK ] test_new_yangdata
28: [ RUN      ] test_path_yangdata
28: [       OK ] test_path_yangdata
28: [ RUN      ] test_validate_yangdata
28: [  ERROR   ] --- Test failed with exception: Segmentation fault(11)
[...]

PR #680, opened 2018-12-05
unittest crash fixes for test_yang_data

eqvinox:
These should fix the crash seen on Debian big-endian builds (#679)

(The realloc() does not cause the crash, I just found it while looking for the crash)

### DIFF
--- a/src/resolve.c
+++ b/src/resolve.c
@@ -5038,9 +5038,10 @@ resolve_extension(struct unres_ext *info, struct lys_ext_instance **ext, struct 
             /* nothing change */
             break;
         case LYEXT_COMPLEX:
-            tmp_ext = realloc(*ext, sizeof **ext + ((struct lyext_plugin_complex*)e->plugin)->instance_size);
+            tmp_ext = realloc(*ext, ((struct lyext_plugin_complex*)e->plugin)->instance_size);
             LY_CHECK_ERR_GOTO(!tmp_ext, LOGMEM(ctx), error);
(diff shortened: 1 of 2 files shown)


src/resolve.c
```diff
@@ -5038,7 +5038,7 @@ resolve_extension(struct unres_ext *info, struct lys_ext_instance **ext, struct
             /* nothing change */
             break;
         case LYEXT_COMPLEX:
-            tmp_ext = realloc(*ext, ((struct lyext_plugin_complex*)e->plugin)->instance_size);
+            tmp_ext = realloc(*ext, sizeof **ext + ((struct lyext_plugin_complex*)e->plugin)->instance_size);
```
rkrejci: Actually it seems correct to me - the `instance_size` is supposed to cover the whole size including the `lys_ext_instance_complex` structure size. But I see a problem in the following memset. A few months ago, there was a change in `lys_ext_instance_complex` structure which doesn't seem to be reflected here. I'll push the fix into the PR branch.

rkrejci:
Hi David, please could you check it now? I mean - I'm not sure what did you see on the realloc call, so I don't want to overlook something.

eqvinox:
> Hi David, please could you check it now? I mean - I'm not sure what did you see on the realloc call, so I don't want to overlook something.

I thought the `->parent` pointer was wrong in `lys_is_disabled` (but it isn't) so I looked at where it comes from, and i saw the realloc and thought the memset is correct but the realloc is not :)

I'll run your change on the powerpc box just to make sure.

eqvinox:
```
100% tests passed, 0 tests failed out of 216
```

All good!

PR #682, opened 2018-12-07
Update resolve.c

NoobAruna:
when strchr returns 0, above code might crash, so added a safety code check when strchr returns 0. I am getting crash in case of invalid identityref.


michalvasko:
Hi,
after some formatting problems (on every single line, actually...) it was good to be merged, thanks.

Regards,
Michal

PR #685, opened 2018-12-14
Optimize check_leaf_list_backlinks and ly_set

olivier-matz-6wind:
Hi,

This is more a request for comments instead of a pull request. In two words, the patchset adds the support for hash table in ly_set and optimize check_leaf_list_backlinks().

But let's first rewind to my initial problem: I have an application that basically copies a big data tree (~50K nodes), do some modification in it, do some xpath requests, then frees it. The copy and the free takes some time (several seconds), and it appears that most of the time is spent in check_leaf_list_backlinks().

So I managed to reduce its complexity. During this, I realized that ly_set is not what I expected at first glance (it is more a vector than a set). After the patches, the performance gain is less than what I expected (only 2x faster in my use case), but I suspect it will scale better with more nodes.

- The first 2 patches of series can probably be merged as is.
- The 3rd one is just a check to ensure that once a ly_set is used as a list or dict, it stays in this mode, which is a condition to implement the hash table for the dict mode. It can probably be removed.
- The patches 4-6 implement the optimization, but I have many questions :)

About ly_set, I wonder if it's really a good idea to add the hash_table support? Maybe there is already a structure doing that job in libyang. Moreover, modifying ly_set has some impacts in the code that access in the structure internals (see the first patch in libyang, but there is also one in sysrepo and many in netopeer & sysrepo unit tests).

I'm also a bit confused by the "duality" of the ly_set which is either a vector or a set, depending on a parameter passed to ly_set_add(). If it does not already exist in libyang, having a new ly_set_htable type is perhaps a better approach than what I did.

My performance issue is only partially solved. Even if it's faster, I see that check_leaf_list_backlinks() is called many times during the lyd_dup and during the lyd_free (I suppose it's in O(nb nodes)). Is it really required? Especially for a dup or a free of the whole tree?

Advices or guidelines would be appreciated.
Thanks !

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -158,7 +158,7 @@ ly_ctx_new_yl_legacy(struct ly_ctx *ctx, struct lyd_node *yltree)
     struct lyd_node *module, *node;
     struct ly_set *set;
     const char *name, *revision;
-    struct ly_set features = {0, 0, {NULL}};
+    struct ly_set features = {0};
     const struct lys_module *mod;
 
     set = lyd_find_path(yltree, "/ietf-yang-library:yang-library/modules-state/module");
@@ -215,7 +215,7 @@ ly_ctx_new_yl_common(const char *search_dir, const char *input, LYD_FORMAT forma
     unsigned int i, u;
     struct lyd_node *module, *node;
     const char *name, *revision;
-    struct ly_set features = {0, 0, {NULL}};
+    struct ly_set features = {0};
     const struct lys_module *mod;
     struct lyd_node *yltree = NULL;
     struct ly_ctx *ctx = NULL;
--- a/src/libyang.h.in
+++ b/src/libyang.h.in
@@ -1704,6 +1704,10 @@ union ly_set_set {
 struct ly_set {
     unsigned int size;               /**< allocated size of the set array */
     unsigned int number;             /**< number of elements in (used size of) the set array */
+#define LY_SET_NONE 0x00
+#define LY_SET_LIST 0x01
+#define LY_SET_DICT 0x02
+    unsigned int type;               /**< none, list, or dict */
     union ly_set_set set;            /**< set array - union to keep ::ly_set generic for data as well as schema trees */
 };
 
--- a/src/libyang.h.in
+++ b/src/libyang.h.in
@@ -15,6 +15,7 @@
 #ifndef LY_LIBYANG_H_
 #define LY_LIBYANG_H_
 
+#include <sys/queue.h>
 #include <stdio.h>
 
 /**
@@ -1691,6 +1692,16 @@ union ly_set_set {
 };
 
 /**
+ * An element of the hash table
+ */
+struct ly_set_elt {
+    unsigned int index;
+    void *node;
+    LIST_ENTRY(ly_set_elt) next;
+};
+LIST_HEAD(ly_set_elt_list, ly_set_elt);
+
+/**
  * @brief Structure to hold a set of (not necessary somehow connected) ::lyd_node or ::lys_node objects.
  * Caller is supposed to not mix the type of objects added to the set and according to its knowledge about
  * the set content, it is supposed to access the set via the sset, dset or set members of the structure.
@@ -1709,6 +1720,8 @@ struct ly_set {
 #define LY_SET_DICT 0x02
     unsigned int type;               /**< none, list, or dict */
     union ly_set_set set;            /**< set array - union to keep ::ly_set generic for data as well as schema trees */
+    unsigned int htable_size;        /**< size of the hash table (always a power of 2) */
+    struct ly_set_elt_list *htable;  /**< hash table to store elements */
 };
 
 /**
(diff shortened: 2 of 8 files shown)


olivier-matz-6wind:
> thank you for your effort, but my first question is why have you written
> your own hash tables when there is already an implementation in libyang
> together with a hashing function, was it unsuitable for some reason?
> Maybe you have missed it but there is an internal header called
> hash_table.h so I hope not :)

hum well... probably for bad reasons.

My initial thinking was that storing a set of keys (which
are nodes memory addresses) without values would not require the use of
the hash table which associates a key of any type to a value.

Do you think that adding a hash table in ly_set is something useful?
If yes, I can try to do the patch again with the libyang hash table.
Else, let's just drop that part of the patchset.

> As for "duality" of ly_set, I would not call it that way. If you use the
> flag, the only difference is that it is not checked whether the
> particular item is already in the set. It is assumed it is not on the
> callers responsibility so that inserting the item can be more
> efficient. The flag name is perhaps misleading.

Ok. I'm maybe biased because of the python set() type. If I understand
properly, ly_set is more a vector than a set, and it should not be used
if we want for store + lookup with many nodes. For that, we should use
a hash table, right?

Anyway, I think the 2 patches "do not bypass ly_set api" and "add a test
for ly_set" could be merged. What do you think?

> Regarding your original issue, you are right that lyd_dup() and
> lyd_free() are not particular effective. I have made some changes
> recently (f41c25e) that cause lyd_free_withsiblings() to skip all
> redundant work when you are freeing a whole data tree (it is called on a
> node without a parent), so you may want to use that. Similar improvement
> for lyd_dup_withsiblings() would make sense. The way I understand it
> this way your issue would be solved with little effort.

That's great, I'll test my use-case with it. Do you want me to check if
I can do the same for the dup ?

> It is also important to mention that @rkrejci is currently working on
> some significant (but mostly internal) changes to libyang. These changes
> concern mainly schemas but there will be some improvements when handling
> data, too. This is relevant for 2 reasons because, firstly, we do not
> want to make much changes to the current libyang anymore and, secondly,
> these backlinks are among the things we were thinking of handling
> differently so this issue may no longer be even relevant.

Ok, got it.

Thanks!
Olivier

michalvasko:
> Do you think that adding a hash table in ly_set is something useful?
If yes, I can try to do the patch again with the libyang hash table.
Else, let's just drop that part of the patchset.

I do not think we would want to change `ly_set` as it is fully in the public API. What would be possible is to use hash table for this particular (internal) use-case, leafref backlinks. However, like I said, there may be no backlinks after the changes to *libyang* so I think the most reasonable thing to do would be to keep it the way it is for now.

> Ok. I'm maybe biased because of the python set() type. If I understand
properly, ly_set is more a vector than a set, and it should not be used
if we want for store + lookup with many nodes. For that, we should use
a hash table, right?

Yeah, our set is meant as simply a nicer structure than a simple array to return. Hash table is optimized for lookup.

> Anyway, I think the 2 patches "do not bypass ly_set api" and "add a test
for ly_set" could be merged. What do you think?

Right, I believe so.

> That's great, I'll test my use-case with it. Do you want me to check if
I can do the same for the dup ?

Yes, that would definitely be a useful patch and should solve your problem. Although, probably some serious refactoring will be needed for the code to not be a mess. You can look at my changes for `lyd_free_withsiblings()` in the commit I referenced but I am afraid `lyd_dup_withsiblings()` will be somewhat more difficult.

Regards,
Michal

olivier-matz-6wind:
> Yes, that would definitely be a useful patch and should solve your
> problem. Although, probably some serious refactoring will be needed
> for the code to not be a mess. You can look at my changes for
> lyd_free_withsiblings() in the commit I referenced but I am afraid
> lyd_dup_withsiblings() will be somewhat more difficult.

I did the following change which solves my issue (from several seconds
of latency to ~immediate):
https://github.com/6WIND/libyang/commit/7f6eee00766afb6c2c23226dc11612aecceffb46

It passes libyang tests and our own cli tests, but I must confess that
I don't master this code very well, so I may be missing something
important.

Please let me know if it's what you expected. BTW, no urgency on my side.

Regards,
Olivier

michalvasko:
Hi Olivier,
I had something a bit different in mind than what you did.

Firstly, we were talking about `lyd_dup_withsiblings()`, not `lyd_dup()`. The difference is that the former one duplicates all the siblings (preceding as well as following) while the latter one does not so it may not be safe to assume it will remain valid in case it has some siblings.

Secondly, while this may be the most urgent problem (the repeated validation), I think there are other tasks performed that can be skipped in case of duplicating a whole data tree.

Lastly, I do not think there is a need for an internal public flag which I would avoid wherever possible.

So, I will implement an efficient `lyd_dup_withsiblings()` once I finish other things I am currently working on.

Regards,
Michal

olivier-matz-6wind:
Ok, thanks! Please let me know if I can help, at least for testing.

michalvasko:
Hi,
you can try it, should be even faster than your optimization.

Regards,
Michal

olivier-matz-6wind:
It works great, thanks!

PR #702, opened 2019-01-25
Add basic tests to fuzz the libyang xml and YANG/YIN parsers with AFL.

jvijtiuk:
Hi,

I've recently been fuzzing the libyang YANG parser with the american fuzzy lop (AFL) fuzzer, and in the process i had to write test programs that were used by the fuzzer.

This PR would add the test programs i used to fuzz the parser, the required cmake changes, and some notes about fuzzing in the README.

I decided to use the value 100 in the __AFL_LOOP(100) statements after experimenting with the fuzzer on my machine, so the value could differ for other fuzzer runs, especially if the code that was fuzzed was different.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,7 @@ option(ENABLE_CALLGRIND_TESTS "Build performance tests to be run with callgrind"
 option(ENABLE_CACHE "Enable data caching for schemas and hash tables for data (time-efficient at the cost of increased space-complexity)" ON)
 option(ENABLE_LATEST_REVISIONS "Enable reusing of latest revisions of schemas" ON)
 option(ENABLE_LYD_PRIV "Add a private pointer also to struct lyd_node (data node structure), just like in struct lys_node, for arbitrary user data" OFF)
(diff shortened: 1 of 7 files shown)


michalvasko:
Hi,
we are a bit confused by your pull request, which may be connected to us never trying fuzz testing before. Anyway, strictly by looking at the actual code, it is trivial. So if someone wanted to test what you tested I do not think it would be a problem to write these small programs themselves. Also, from what I understand you still need to provide your own input. Could you maybe elaborate more on the exact benefit of the proposed changes?

Regards,
Michal

michalvasko:
Hi,
Mislav has explained what this is all about so just say when the pull request is ready and it can be merged since it does not change default build in any way now (also, you probably want to remove `tests/fuzz/xmlfuzz.c` and `tests/fuzz/yangfuzz.c` as they are obsolete now). Thanks.

Regards,
Michal

jvijtiuk:
Hi,

after additional discussion with Mislav, the tests are now in tests/fuzz instead of the fuzz subdirectory.
I've removed some trailing whitespace and updated the commit today, so the pull request can now be merged.

Regards,
Juraj

michalvasko:
Hi Juraj,
everything seems fine, thanks.

Regards,
Michal

PR #707, opened 2019-01-31
OS and PKG fixes

mwinter-osr:
These are a few fixes for the Libyang
- FreeBSD build fix (uses sys/endian.h instead of endian.h)
- Solaris/OmniOS fix (missing le16toh() etc)
- CentOS/RedHat 6/7 Pkg fix (both can't build the language bindings, so disable this in package as the swig is too old and RH6 has Valgrind and pcre-devel too old as well. Disabled valgrind checks and caching to work with old pcre-devel)

The CentOS/RedHat limitations are detected on build automatically. Unfortunately, this gives limits the functionality (no language bindings), but for FRRouting we still support and depend on libyang on these platforms.

### DIFF
--- a/packages/libyang.spec.in
+++ b/packages/libyang.spec.in
@@ -8,20 +8,44 @@ Source1: libyang.rpmlintrc
 License: BSD-3-Clause
 BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}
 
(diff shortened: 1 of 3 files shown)


michalvasko:
Hi,
all the changes seems OK, thanks.

Regards,
Michal

PR #709, opened 2019-02-05
More rpm fixes

mwinter-osr:
Fix for previous package fix. Broke building on newer systems as RPM keywords can't be indented.

This PR fixes it and adds support for Fedora as well (Fedora < 27 without language bindings and full features on Fedora >= 27)


michalvasko:
Okay, no problem, thank you.

Regards,
Michal

PR #721, opened 2019-03-05
C++: Add value getter to Data_Node_Anydata

syyyr:
A way to retrieve the value field from `lyd_node_anydata` was missing, so I added a getter. To avoid making the return type too long, I added a typedef for the value union.

cc @jktjkt

### DIFF
--- a/src/tree_data.h
+++ b/src/tree_data.h
@@ -259,6 +259,17 @@ struct lyd_node_leaf_list {
 /* 0x80 is reserved for internal use */
 
 /**
+ * @brief Anydata value union
+ */
+typedef union {
+    const char *str;             /**< string value, in case of printing as XML, characters like '<' or '&' are escaped */
+    char *mem;                   /**< raw memory (used for LYB format) */
+    struct lyxml_elem *xml;      /**< xml tree */
+    struct lyd_node *tree;       /**< libyang data tree, does not change the root's parent, so it is not possible
+                                      to get from the data tree into the anydata/anyxml */
+} lyd_anydata_value;
+
+/**
  * @brief Structure for data nodes defined as #LYS_ANYDATA or #LYS_ANYXML.
  *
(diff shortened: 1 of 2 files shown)


swig/cpp/src/Tree_Data.hpp
```diff
@@ -268,6 +268,8 @@ class Data_Node_Anydata : public Data_Node
     ~Data_Node_Anydata();
     /** get value_type variable from [lyd_node_anydata](@ref lyd_node_anydata)*/
     LYD_ANYDATA_VALUETYPE value_type() {return ((struct lyd_node_anydata *) node)->value_type;};
+    /** get value variable from [lyd_node_anydata](@ref lyd_node_anydata)*/
+    LYD_ANYDATA_VALUE value() {return ((struct lyd_node_anydata *) node)->value;};
     //union value
```
rkrejci: this is probably kind of TODO comment you are fixing, so it should be removed

src/tree_data.h
```diff
@@ -258,6 +258,17 @@ struct lyd_node_leaf_list {
...
+/**
+ * @brief Anydata value union
+ */
+typedef union {
+    const char *str;             /**< string value, in case of printing as XML, characters like '<' or '&' are escaped */
+    char *mem;                   /**< raw memory (used for LYB format) */
+    struct lyxml_elem *xml;      /**< xml tree */
+    struct lyd_node *tree;       /**< libyang data tree, does not change the root's parent, so it is not possible
+                                      to get from the data tree into the anydata/anyxml */
+} LYD_ANYDATA_VALUE;
```
rkrejci: Upper case names are so far used only for macros and enums, so please name the union in lower case

swig/cpp/src/Tree_Data.hpp
syyyr: done

src/tree_data.h
syyyr: done

PR #725, opened 2019-03-08
C++: extend unit tests

mamidzic:
Extend unit tests coverage in test_tree_data.cpp and test_tree_schema.cpp

Also fix a memory leak in Schema_Node xpath_atomize.

### DIFF
--- a/swig/cpp/src/Tree_Schema.cpp
+++ b/swig/cpp/src/Tree_Schema.cpp
@@ -281,7 +281,8 @@ S_Set Schema_Node::xpath_atomize(enum lyxp_node_type ctx_node_type, const char *
         return nullptr;
     }
 
-    return std::make_shared<Set>(set, deleter);
+    S_Deleter new_deleter = std::make_shared<Deleter>(set, deleter);
+    return std::make_shared<Set>(set, new_deleter);
 }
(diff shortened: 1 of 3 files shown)


swig/cpp/src/Tree_Schema.cpp
```diff
@@ -281,6 +281,7 @@ S_Set Schema_Node::xpath_atomize(enum lyxp_node_type ctx_node_type, const char *
         return nullptr;
     }
 
+    S_Deleter new_deleter = std::make_shared<Deleter>(set, deleter);
     return std::make_shared<Set>(set, deleter);
```
jktjkt: this should likely return a shared_ptr with `new_deleter` instead of the old deleter, right?

swig/cpp/src/Tree_Schema.cpp
```diff
@@ -290,6 +291,7 @@ S_Set Schema_Node::xpath_atomize(int options) {
...
```
jktjkt: same here

swig/cpp/src/Tree_Schema.cpp
mamidzic: Yes, you are right, I'll fix it.

PR #731, opened 2019-03-12
c++: Virtual methods require a virtual destructor

jktjkt:
This might lead to calling a wrong destructor when destroying an instance of a child class. clang fortunately warns about that:
```
warning: destructor called on non-final 'libyang::Data_Node' that has virtual functions but non-virtual destructor [-Wdelete-non-virtual-dtor]
```


michalvasko:
Hi,
I know little about this, but the change seems reasonable so I am going to trust you :)

Regards,
Michal

PR #732, opened 2019-03-13
Remove unused members

jktjkt:
Looking at the source, both of these flags appear to be used just internally within libyang. They are only set and never read, and therefore clang warns about their use:
```
swig/cpp/src/Tree_Data.hpp:89:13: warning: private field 'flags' is not used [-Wunused-private-field]
swig/cpp/src/Tree_Schema.hpp:434:13: warning: private field 'flags' is not used [-Wunused-private-field]
```
If they are useful for users of libyang and if they should be made available, then an accessor and presumably also some consumers of this variable should be added.

### DIFF
--- a/swig/cpp/src/Tree_Data.cpp
+++ b/swig/cpp/src/Tree_Data.cpp
@@ -30,10 +30,9 @@ extern "C" {
 
 namespace libyang {
 
-Value::Value(lyd_val value, LY_DATA_TYPE* value_type, uint8_t value_flags, S_Deleter deleter):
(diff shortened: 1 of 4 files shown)


michalvasko:
Hi,
they really are used mostly internally, but in some cases they may be needed. There can be `LYD_VALUE_UNRES` and `LYD_VALUE_USER` flags set which affect what is stored in the union `value` (something undefined or a user-encoded value, respectively) and how to access it.

I think this information may be relevant but I do not understand the bindings enough to decide what changes should be done (but simply removing them is not likely to be the correct solution). Can you maybe help? Thanks.

Regards,
Michal

PR #748, opened 2019-04-08
Extension plugins validation enhancements

### LINKED ISSUE #744: extension data validation
Hello,

I am trying to implement a custom extension backed by a libyang plugin to do some data validation.

Basically, I need to perform the validation of unicity amongst a set of nodes represented by an xpath. To make sure the condition is only evaluated once, I want to place the extension on a container and the xpath will reference children nodes of that container.

Here is an example:

```yang
module unicity {
  namespace "urn:test:unicity";
  prefix u;
  revision 2019-03-28 {
    description "";
    reference "";
  }

  extension unique-values {
    argument "xpath";
    description "All node values referenced by the xpath must be unique.";
  }

  container data {
    unique-values "*/name";
    list first {
      key name;
      leaf name {
        type string;
      }
    }
    list second {
      key name;
      leaf name {
        type string;
      }
    }
  }
}
```

I implemented a minimalist plugin:

```c
struct lyext_plugin unique_values = {
	.type = LYEXT_FLAG,
	.flags = LYEXT_OPT_VALID,
	.check_position = &unique_values_check_position,
	.check_result = &unique_values_check_result,
	.valid_data = &unique_values_valid_data,
};

struct lyext_plugin_list unicity[] = {
	{"unicity", "2019-03-28", "unique-values", &unique_values},
	{NULL, NULL, NULL, NULL} /* terminating item */
};
```

I copied the .so file into the "extensions" folder and I checked with gdb that my plugin is loaded. The `unique_values_check_position` and `unique_values_check_result` are indeed called when the yang schema is parsed.

However, the `unique_values_valid_data` callback is never called upon data validation with a data tree containing list elements that have identical "name" values in both lists. I'm not sure why.

Did I miss something? Could you help?

rjarry:
This should fix #744 and allow plugins to report proper validation errors.

Please let me know what you think.

### DIFF
--- a/src/tree_schema.h
+++ b/src/tree_schema.h
@@ -419,6 +419,10 @@ typedef enum {
 #define LYEXT_OPT_CONTENT    0x04    /**< content of lys_ext_instance_complex is copied from source (not dup, just memcpy). */
 /** @endcond */
 #define LYEXT_OPT_VALID      0x08    /**< needed to call calback for validation */
+#define LYEXT_OPT_VALID_SUBTREE 0x10 /**< The plugin needs to do validation on nodes in the subtree of the extended node
+                                          (i.e. not only the extended node nor its direct children). valid_data callback
+                                          will be called when any descendant node in the subtree of the extended node is
+                                          modified. */
 #define LYEXT_OPT_PLUGIN1    0x0100  /**< reserved flag for plugin-specific use */
 #define LYEXT_OPT_PLUGIN2    0x0200  /**< reserved flag for plugin-specific use */
 #define LYEXT_OPT_PLUGIN3    0x0400  /**< reserved flag for plugin-specific use */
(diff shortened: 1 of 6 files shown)


src/tree_schema.h
```diff
@@ -419,6 +419,10 @@ typedef enum {
 #define LYEXT_OPT_CONTENT    0x04    /**< content of lys_ext_instance_complex is copied from source (not dup, just memcpy). */
 /** @endcond */
 #define LYEXT_OPT_VALID      0x08    /**< needed to call calback for validation */
+#define LYEXT_OPT_SUBTREE    0x10    /**< The plugin needs to do validation on nodes in the subtree of the extended node
```
michalvasko: This flag meaning is that the whole subtree should be validated but the name does not convey this. Maybe rename it to `LYEXT_OPT_VAL_SUBTREE` or something better?

michalvasko:
Hi,
we have decided it can be merged once you address the comment despite me being against it. You have written an extension for something that is already supported by standard YANG and implemented efficiently in *libyang* just because a model is not written well. In my opinion that is a very poor reason to add options to our extensions API, hence my objection.

Regards,
Michal

src/tree_schema.h
rjarry: You're right.

At first, I wanted to name it `LYEXT_OPT_VALID_SUBTREE` (and its related `LYS_VALID_EXT_SUBTREE` schema flag) so they have the same prefix that  `LYEXT_OPT_VALID` and `LYS_VALID_EXT`.

However these names are quite long and it breaks source alignment. Would that be OK anyway?

michalvasko: Always better than having vague names.

rjarry: I pushed the rename, please let me know if there are other changes required.

michalvasko:
Okay, now the names are fine.

Regards,
Michal

rjarry:
Thanks

PR #757, opened 2019-04-16
extensions BUGFIX fix link error when ENABLE_STATIC=ON

rjarry:
Fix the following error when ENABLE_STATIC is "ON".

```
[100%] Linking C executable yanglint
libmetadata.a(metadata.c.o):(.data+0xa8): multiple definition of `lyext_api_version'
libnacm.a(nacm.c.o):(.data+0x0): first defined here
libyangdata.a(yangdata.c.o):(.data+0x60): multiple definition of `lyext_api_version'
libnacm.a(nacm.c.o):(.data+0x0): first defined here
libuser_inet_types.a(user_inet_types.c.o):(.data+0x0): multiple definition of `lytype_api_version'
libuser_yang_types.a(user_yang_types.c.o):(.data+0x0): first defined here
```

Since we are linking all extensions statically in the same binary, we must not have duplicate public symbols. When building with ENABLE_STATIC dynamic plugins loading is disabled anyway, we do not need version check.

### DIFF
--- a/src/extensions.h
+++ b/src/extensions.h
@@ -35,7 +35,11 @@ extern "C" {
  * @brief Macro to store version of extension plugins API in the plugins.
  * It is matched when the plugin is being loaded by libyang.
  */
+#ifdef STATIC
+#define LYEXT_VERSION_CHECK
+#else
 #define LYEXT_VERSION_CHECK int lyext_api_version = LYEXT_API_VERSION;
+#endif
 
 /**
  * @brief Extension instance structure parent enumeration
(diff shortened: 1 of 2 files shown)


rjarry:
By the way, could it be possible to add a check in `.travis.yml` so that static build is not broken with new changes? I tried to do it myself but I'm not sure how to do it properly.

michalvasko:
Hi,
I have merged it and added a static build to *travis*, thanks.

Regards,
Michal

rjarry:
Thanks

PR #758, opened 2019-04-18
swig/cpp: add functions enable/disable/state for features

YohanPipereau:


michalvasko:
Hi,
no problem, merging.

Regards,
Michal

PR #779, opened 2019-05-13
very special test case for hash table.

fanchanghu:
Hi,
I have met an error like '/libyang/src/xpath.c:642: set_remove_node_hash: Assertion `!r' failed.' 

It is very hard to reproduce this error with libyang API, so I try to reproduce it  by a test case of hash table.

this test should be passed, but not. 

Could you check it, thanks.

### DIFF
--- a/src/hash_table.c
+++ b/src/hash_table.c
@@ -588,8 +588,9 @@ int
 lyht_insert_with_resize_cb(struct hash_table *ht, void *val_p, uint32_t hash,
                            values_equal_cb resize_val_equal, void **match_p)
 {
-    struct ht_rec *rec, *crec = NULL;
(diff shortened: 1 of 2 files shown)


fanchanghu:
Hi,
I have tryed to fix it like that: If the inserted element is the first element, and any other element has hold the right position, move it to any other available position and insert the element to the right position.

@michalvasko

michalvasko:
Hi,
I have decided to fix this in another way. I have basically reverted also the previous fix and instead the records are moved when possible, which should speed up a bit later look-ups but more importantly prevent these errors caused by invalid records making too much of a mess of the record order. Thanks for the effort.

Regards,
Michal

PR #809, opened 2019-06-22
Fix two issues of yang2yin

brchiu:
Two issues of yang2yin are fixed in this pull reuqest :

1. Fix "when" tag erroneously closed by "condition".
2. Fix the problem that there is string enclosed by quote sign, e.g. "...", inside comment can not be handled properly.


michalvasko:
Hi,
you are right but this utility is not really meant for actual use, we developed it primarily to be used in our tests. For converting schemas from YANG to YIN, you should use `yanglint`. Thanks anyway.

Regards,
Michal

PR #833, opened 2019-07-30
Avoid coverage test failure

### LINKED ISSUE #832: libyang2: conflict between valgrind and coverage
When compiling libyang2, `cmake` with `ENABLE_VALGRIND_TESTS=ON` and `ENABLE_COVERAGE=ON`, the valgrind test and the coverage statistics will fail.
```
$ cmake  -DENABLE_COVERAGE=ON ..
```
```
$ make test
```

image
```
$ make coverage
```
image

so if we want to make coverage, we should turn off the ENABLE_VALGRIND_TESTS at first.
```
$ cmake -DENABLE_VALGRIND_TESTS=OFF  -DENABLE_COVERAGE=ON ..
```

Alanscut:
Fixes #832

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,7 @@ $ make test
 Based on the tests run, it is possible to generate code coverage report via the
 make's `coverage` target:
 ```
+$ cmake -DENABLE_VALGRIND_TESTS=OFF -DENABLE_COVERAGE=ON ..
 $ make coverage
 ```
 


Alanscut:
> Why is it necessary to disable Valgrind here?

Please see #832 , if you don't disable the Valgrind, it will fail when you `make coverage` in gcc 4.8.

jktjkt:
> > Why is it necessary to disable Valgrind here?
> 
> Please see #832 , if you don't disable the Valgrind, it will fail when you `make coverage` in gcc 4.8.

Thanks. I disagree with this; if there's a bug in GCC 4.8, then we should not be patching it around, but encourage people to use a newer toolchain.

Historically, the biggest reasons people used GCC 4.8 was RHEL/CentOS, but they have been supporting newer toolchains even on old RHEL via Software Collections. What is your reason for sticking with GCC 4.8?

Alanscut:
> > > Why is it necessary to disable Valgrind here?
> > 
> > 
> > Please see #832 , if you don't disable the Valgrind, it will fail when you `make coverage` in gcc 4.8.
> 
> Thanks. I disagree with this; if there's a bug in GCC 4.8, then we should not be patching it around, but encourage people to use a newer toolchain.
> 
> Historically, the biggest reasons people used GCC 4.8 was RHEL/CentOS, but they have been supporting newer toolchains even on old RHEL via Software Collections. What is your reason for sticking with GCC 4.8?

ok, I will upgrade my gcc, and close this PR.

PR #839, opened 2019-08-12
bugfix: use unicode character(treated as utf-8 encoding) count for length constraint

fanchanghu:
Hi,
We are using a wide charater language with libyang, but the length constraint does not work well, so I try to fix it.
Can you check this pr or give some suggestions ? 
thanks.


rkrejci:
Hi, seems fine to me - I've just slightly modified code formatting and added `ly_` prefix to the function.

Thanks!

PR #841, opened 2019-08-14
add yang2yin help messages

open520:
yang2yin is an important tool.  I noticed that more information can be added to the  input_file and output_file params for users.  So that user can use it more smoothly.


rkrejci:
Thanks!

PR #845, opened 2019-08-17
remove unused variable char *cwd in ly_ctx_new()

dota17:
When I read the function ly_ctx_new(), I find that the variable cwd is unused.  And I test the code, It's OK.

So I removed the variable. Tell me if I'm wrong.


michalvasko:
Hi,
seems fine to me, thanks.

Regards,
Michal

PR #848, opened 2019-08-20
Fix the potential dead loop in ly_ctx_new()

dota17:
As I know, there can be 1024 pthread_key_t at most in a process.
Every ly_ctx struct hold one pthread_key_t.  So, if there are more than 1024 ly_ctx structs, the statements bellow in ly_ctx_new() function will run in dead loop.

```
    /* initialize thread-specific key */
    while ((i = pthread_key_create(&ctx->errlist_key, ly_err_free)) == EAGAIN);
```

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -99,7 +99,10 @@ ly_ctx_new(const char *search_dir, int options)
     ly_load_plugins();
 
     /* initialize thread-specific key */
-    while ((i = pthread_key_create(&ctx->errlist_key, ly_err_free)) == EAGAIN);
+    if (pthread_key_create(&ctx->errlist_key, ly_err_free) == EAGAIN) {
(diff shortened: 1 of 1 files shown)


fredgan:
Yes, if there are more than 1024 ly_ctx, it will go to dead loop.

src/context.c
```diff
@@ -99,7 +99,10 @@ ly_ctx_new(const char *search_dir, int options)
     ly_load_plugins();
 
     /* initialize thread-specific key */
-    while ((i = pthread_key_create(&ctx->errlist_key, ly_err_free)) == EAGAIN);
+    if (pthread_key_create(&ctx->errlist_key, ly_err_free) == EAGAIN) {
```
rkrejci: ok, you are right, but then the condition should test any error of `pthread_key_create`, not just the `EAGAIN`

rkrejci:
Merged manually, thanks!

PR #871, opened 2019-09-06
Add nested grouping testcase for sec5_5

fredgan:


rkrejci:
Hi, does it really test something new? Codecov does not detect any new line covered - I know that it doesn't mean  the a specific test case is already tested, I'm just asking what is special about this particular test case.

Issue #877, opened 2019-09-14
yanglint looping/high CPU on some models


evyncke:
yanglint 0.14.81 (an old one... will try with a fresh version) cannot process one specific YANG model and appears to run in a loop with high CPU.

```
yanglint -V -i -p vendor/cisco/nx/7.0-3-I7-4/ vendor/cisco/nx/7.0-3-I7-4/Cisco-NX
-OS-device.yang
err : Invalid keyword "cpu history duration name, it can be last60seconds, \
                          last60minutes, last60hours"".
```

Model is from [https://raw.githubusercontent.com/YangModels/yang/master/vendor/cisco/nx/7.0-3-I7-4/Cisco-NX-OS-device.yang]

fredgan:
Yes, you find a bug. Thank you very much.
The reason is this as follow, I will fix it.
```
static void
yang_tpdf_free(struct ly_ctx *ctx, struct lys_tpdf *tpdf, uint16_t start, uint16_t size)
{
    uint8_t i;  // max is 255,  and your module has more than 2000 typedefs, so this will run in dead roop.

    assert(ctx);
```

evyncke:
Thank you, now, let's hope for a quick merge ;-)

PR #878, opened 2019-09-16
fix issue #877

fredgan:
Fix issue #877


rkrejci:
Thanks! Marged manually to bump versions.

PR #879, opened 2019-09-17
libyang: Cosmetic change to pkgconfig

neheb:
For consistency between different packages.


michalvasko:
Hi,
okay, if you say so, no harm in merging.

Regards,
Michal

PR #882, opened 2019-09-25
Add action/anydata/modifier stmt for yang2yin

fredgan:
When I am trying to add some testcases, I find that some keywords are missing in yang2yin tool. So I add this commit first.


michalvasko:
Hi,
no problem, merging.

Regards,
Michal

Issue #889, opened 2019-09-27
libyang segmentation fault after failing to insert node due to different parents


chris-staley:
Hello, 

I'm running into an issue where it says a node has different parents when I don't believe that's the case. The node is not created and I receive a segfault when it attempts to check its validity.

The issue can be reproduced in yanglint using this JSON rpc request:

`{"Cisco-IOS-XE-rpc:test":{"Cisco-IOS-XE-cable-diag-rpc:cable-diagnostics":{"tdr":{"interface":{"GigabitEthernet":"1\/0\/1"}}}}}`

and with these YANG models: 

https://github.com/YangModels/yang/blob/master/vendor/cisco/xe/16121/Cisco-IOS-XE-cable-diag-rpc.yang
https://github.com/YangModels/yang/blob/master/vendor/cisco/xe/16121/Cisco-IOS-XE-rpc.yang

The output is as follows (with bt):

```
(gdb) run
Starting program: /home/chris/Downloads/libyang-1.0-r3/build/yanglint 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
> add ../../../yang/Cisco-IOS-XE-rpc.yang 
> add ../../../yang/Cisco-IOS-XE-cable-diag-rpc.yang 
> verb 3
> data -t rpc -f json ../../../start_test.json 
libyang[3]: DICT: inserting "1/0/1"
libyang[3]: DICT: inserting "1/0/1"
libyang[0]: Cannot insert, different parents ("test" and "cable-diagnostics").

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7f4625d in lyd_wd_add_subtree (root=0x7fffffffddb0, last_parent=0x55555559a5b0, subroot=0x0, schema=0x55555559b0f0, toplevel=0, options=16, unres=0x55555557ad90) at /home/chris/Downloads/libyang-1.0-r3/src/tree_data.c:7331
7331                subroot->validity &= ~LYD_VAL_MAND;
(gdb) bt
#0  0x00007ffff7f4625d in lyd_wd_add_subtree (root=0x7fffffffddb0, last_parent=0x55555559a5b0, subroot=0x0, schema=0x55555559b0f0, toplevel=0, options=16, unres=0x55555557ad90) at /home/chris/Downloads/libyang-1.0-r3/src/tree_data.c:7331
[...]

PR #893, opened 2019-10-08
fix the segment fault for #889

fredgan:


michalvasko:
Hi,
your pull request does not really solve the problem, this function should always succeed so checking the return value is nice but ultimately redundant. @rkrejci will take a look and fix the problem as soon as he can.

Regards,
Michal

fredgan:
OK. I also think so and I commented it in #889 
Thanks @michalvasko

PR #903, opened 2019-10-15
printer: ignore EAGAIN and EINTR errno

rjarry:
When using libnetconf2 via the UNIX or FD transport and sending large configs, we can get the following error from libyang:

```
Print error (Resource temporarily unavailable).
```

Do not fail when errno holds `EAGAIN` or `EINTR`. These are not errors. This can occur when using ly_print_clb with a callback that does I/O on a non-blocking file or if the application received a signal during the write syscall.

Ideally, we should check the return value of every `ly_print` call but it would be rather tedious and would make the code clobbered.

Fixes: 609d708ca567 ("printer CHANGE use errno to check for printer errors")

### DIFF
--- a/src/printer.c
+++ b/src/printer.c
@@ -127,6 +127,15 @@ ly_print(struct lyout *out, const char *format, ...)
     case LYOUT_CALLBACK:
         count = vasprintf(&msg, format, ap);
         count = out->method.clb.f(out->method.clb.arg, msg, count);
+        if (count >= 0) {
+            /*
+             * Depending on what the callback function does, errno might
+             * contain non-zero values that are not real "errors" (EAGAIN or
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
I do not particularly like your fix. I was thinking that it could be fixed instead in the way that if the user callback returns 0 or a positive number, `errno` gets zeroed. What do you think?

Regards,
Michal

michalvasko:
Hi,
seems fine now, merging.

Regards,
Michal

brchiu:
hi, @michalvasko , will this commit apply to devel branch too ? otherwise, these two branches will diverge.

michalvasko:
Yes, it was already merged.

rjarry:
Thanks @michalvasko

PR #907, opened 2019-10-18
libyang2: fix the include statement assertion failed problem

fredgan:
@rkrejci 
Hi Radek,
I find a problem when writing the printer_yin part.

If the module contains "include" statement, the following error happened:
```
yanglint: /data/libyang2/src/tree_schema_helpers.c:708: lysp_load_module_check: Assertion `info->submoduleof' failed.
```
I try to fix this problem by adding one parameter to the function `lys_module_localfile`, I don't know if it is the proper way.

   Thank you!

### DIFF
--- a/src/tree_schema_helpers.c
+++ b/src/tree_schema_helpers.c
@@ -748,8 +748,8 @@ lysp_load_module_check(struct ly_ctx *ctx, struct lysp_module *mod, struct lysp_
 }
 
 LY_ERR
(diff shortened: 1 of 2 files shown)


rkrejci:
Maybe the link to the module itself could be even part of the `lys_parser_ctx` so available whereever in the parser. It would remove the module parameter from many others function, but I think that it is not necessary and since your fix solves the issue, I'm going to merge it.

rkrejci:
Merged as cd485b8 with rewording the commit message.

PR #909, opened 2019-10-21
libyang2: fix issue: printer_yang's deviate statements are missing 

fredgan:
Since deviation is LIST structure, the iterator should be LY_LIST_FOR rather than LY_ARRAY_FOR


fredgan:
@rkrejci  I'm comparing the YANG printer output and YIN printer output, then I found this issue

rkrejci:
Good catch! I'll merge it.

rkrejci:
Merged as 2b11ddb

PR #910, opened 2019-10-21
libyang2: fix read_qstring escape issue

fredgan:
@rkrejci Hi, Radek, I find another issue in the parser when checking my `printer_yin` part .
The issue is as this:
I write an example module3
```
module module3{
  namespace "h\na";
  prefix m3;
}
```
then use  the yanglint print the module
`./yanglint -f yang ./module3.yang`

following is the output:
```
module module3 {
  namespace "h\\n";
  prefix m3;
}
```

As you see, the `\n` is incorrect and the last character `a` is missing!
Then I try to fix this issue. I find that when read `\n` or `\t`,  `need_buf` is also need to be set 1 in the function `read_qstring`.

Thank you very much.

### DIFF
--- a/src/parser_yang.c
+++ b/src/parser_yang.c
@@ -401,9 +401,11 @@ read_qstring(struct lys_parser_ctx *ctx, const char **data, enum yang_arg arg, c
             switch (**data) {
             case 'n':
                 c = "\n";
+                need_buf = 1;
                 break;
             case 't':
                 c = "\t";
+                need_buf = 1;
                 break;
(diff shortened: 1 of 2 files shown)


rkrejci:
Good catch! I'm just going to modify the commit message and merge it.

rkrejci:
Merged as d49fe11

fredgan:
@rkrejci Would you tell me the commit message format for libyang? So that you need not to recommit the PRs for saving time. Thanks

Issue #859, opened 2019-09-02
libyang2: YIN printer


rkrejci:
Print YANG schema (only the parsed structures) in YIN format as defined in [RFC 7950](https://tools.ietf.org/html/rfc7950#section-13)

fredgan:
Almost finished except two cases of extension instances
The PR is  #919

PR #919, opened 2019-10-29
libyang2: yin printer NEW initial yin printer except compiled extension instances

fredgan:
@rkrejci 
Hi Radek,
Yin printer is finished except:
1 The extension instance substatement in extension instance statement is unable to print currently because there are no compiled info in `lysp_stmt`  for them. Currently they are ignored.
2 The compiled printer for extension instance.

### DIFF
--- a/src/printer_internal.h
+++ b/src/printer_internal.h
@@ -101,6 +101,28 @@ LY_ERR yang_print_parsed(struct lyout *out, const struct lys_module *module);
 LY_ERR yang_print_compiled(struct lyout *out, const struct lys_module *module);
 
 /**
+ * @brief YIN printer of the parsed schemas. Full YIN printer.
+ *
+ * @param[in] out Output specification.
+ * @param[in] module Schema to be printed (the parsed member is used).
+ * @return LY_ERR value, number of the printed bytes is updated in lyout::printed.
+ */
+LY_ERR yin_print_parsed(struct lyout *out, const struct lys_module *module);
+
+/**
+ * @brief YIN printer of the compiled schemas.
+ *
+ * This printer provides information about modules how they are understood by libyang.
+ * Despite the format is inspired by YIN, it is not fully compatible and should not be
+ * used as a standard YIN format.
+ *
+ * @param[in] out Output specification.
+ * @param[in] module Schema to be printed (the compiled member is used).
+ * @return LY_ERR value, number of the printed bytes is updated in lyout::printed.
+ */
+LY_ERR yin_print_compiled(struct lyout *out, const struct lys_module *module);
+
+/**
  * @brief XML printer of the YANG data.
  *
  * @param[in] out Output specification.
--- a/src/printer_internal.h
+++ b/src/printer_internal.h
@@ -30,6 +30,14 @@ typedef enum LYOUT_TYPE {
 } LYOUT_TYPE;
 
 /**
+ * @brief Types of the YANG printers
+ */
+enum schema_type {
+    YPR_PARSED,   /**< YANG printer of the parsed schema */
+    YPR_COMPILED  /**< YANG printer of the compiled schema */
+};
+
+/**
  * @brief Printer output structure specifying where the data are printed.
  */
 struct lyout {
--- a/src/printer_internal.h
+++ b/src/printer_internal.h
@@ -118,19 +118,6 @@ LY_ERR yang_print_compiled(struct lyout *out, const struct lys_module *module);
 LY_ERR yin_print_parsed(struct lyout *out, const struct lys_module *module);
 
 /**
- * @brief YIN printer of the compiled schemas.
- *
- * This printer provides information about modules how they are understood by libyang.
- * Despite the format is inspired by YIN, it is not fully compatible and should not be
- * used as a standard YIN format.
- *
- * @param[in] out Output specification.
- * @param[in] module Schema to be printed (the compiled member is used).
- * @return LY_ERR value, number of the printed bytes is updated in lyout::printed.
- */
-LY_ERR yin_print_compiled(struct lyout *out, const struct lys_module *module);
-
-/**
  * @brief XML printer of the YANG data.
  *
  * @param[in] out Output specification.
--- a/src/printer_internal.h
+++ b/src/printer_internal.h
@@ -30,14 +30,6 @@ typedef enum LYOUT_TYPE {
 } LYOUT_TYPE;
 
 /**
- * @brief Types of the YANG printers
- */
-enum schema_type {
-    YPR_PARSED,   /**< YANG printer of the parsed schema */
-    YPR_COMPILED  /**< YANG printer of the compiled schema */
-};
-
-/**
  * @brief Printer output structure specifying where the data are printed.
  */
 struct lyout {
--- a/src/printer_yin.c
+++ b/src/printer_yin.c
@@ -48,8 +48,6 @@ struct ypr_ctx {
 
 static void yprp_extension_instances(struct ypr_ctx *ctx, LYEXT_SUBSTMT substmt, uint8_t substmt_index,
                                struct lysp_ext_instance *ext, int *flag, unsigned int count);
-static void yprc_extension_instances(struct ypr_ctx *ctx, LYEXT_SUBSTMT substmt, uint8_t substmt_index,
-                                 struct lysc_ext_instance *ext, int *flag, unsigned int count);
 
 static void
 ypr_open(struct ypr_ctx *ctx, const char *elem_name, const char *attr_name, const char *attr_value,  int flag)
@@ -122,11 +120,7 @@ ypr_substmt(struct ypr_ctx *ctx, LYEXT_SUBSTMT substmt, uint8_t substmt_index, c
         if (((struct lysp_ext_instance*)ext)[u].insubstmt != substmt || ((struct lysp_ext_instance*)ext)[u].insubstmt_index != substmt_index) {
             continue;
         }
-        if (ctx->schema == YPR_PARSED) {
-            yprp_extension_instances(ctx, substmt, substmt_index, &((struct lysp_ext_instance*)ext)[u], &extflag, 1);
-        } else {
-            yprc_extension_instances(ctx, substmt, substmt_index, &((struct lysc_ext_instance*)ext)[u], &extflag, 1);
-        }
+        yprp_extension_instances(ctx, substmt, substmt_index, &((struct lysp_ext_instance*)ext)[u], &extflag, 1);
     }
 
     /* argument as yin-element */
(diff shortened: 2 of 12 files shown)


src/printer_yin.c
```diff
@@ -0,0 +1,2439 @@
...
+    {"contact",          "text",        SUBST_FLAG_YIN},/**< LY_STMT_CONTACT */
+    {"container",        "name",        SUBST_FLAG_ID}, /**< LY_STMT_CONTAINER */
+    {"description",      "text",        SUBST_FLAG_YIN},/**< LY_STMT_DESCRIPTION */
+    {"deviate",          "value",       SUBST_FLAG_ID}, /**< LY_STMT_DEVIATE */
+    {"deviation",        "target-node", SUBST_FLAG_ID}, /**< LY_STMT_DEVIATION */
+    {"enum",             "name",        SUBST_FLAG_ID}, /**< LY_STMT_ENUM */
+    {"error-app-tag",    "value",       SUBST_FLAG_ID}, /**< LY_STMT_ERROR_APP_TAG */
+    {"error-message",    "value",       SUBST_FLAG_YIN},/**< LY_STMT_ERROR_MESSAGE */
+    {"extension",        "name",        SUBST_FLAG_ID}, /**< LY_STMT_EXTENSION */
+    {"feature",
```
rkrejci: But this is only the cas if the input data format was YANG, right? So the same problem as YIN input for YANG printer.

rkrejci: This is supposed to be XML element and XML comment.

rkrejci:
Thanks for your work. Check the comments, but mianly we should talk about printing the compiled schema tree in YIN format. It was not proposed by the #859 and I'm not sure about its purpose.

src/printer_yin.c
```diff
@@ -0,0 +1,2439 @@
...
+    yprc_extension_instances(ctx, LYEXT_SUBSTMT_SELF, 0, ident->exts, &flag, 0);
+    yprc_iffeatures(ctx, ident->iffeatures, ident->exts, &flag);
+
+    LY_ARRAY_FOR(ident->derived, u) {
+        ypr_close_parent(ctx, &flag);
+        if (ctx->module != ident->derived[u]->module) {
+            ly_print(ctx->out, "%*sderived %s:%s;\n", INDENT, ident->derived[u]->module->prefix, ident->derived[u]->name);
+        } else {
+            ly_print(ctx->out, "%*sderived %s;\n", INDENT, ident->derived[u]->name);
+        }
```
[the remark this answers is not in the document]
fredgan: Sorry for my carelessness.
Yes, I think so.
Compile mode is good for validating data, but it is hard for users(just common users) to understand.

src/printer_yin.c
fredgan: Sorry, I don't understant what you mean.
I compared the yin output with pyang and also use the output as input, I don't find any issue.
Can you figure it out more clearly?

fredgan: Also, I don't understand this part.

src/printer_yin.c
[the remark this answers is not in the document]
rkrejci: I thing that it can be important e.g. for module authors to see how groupings or typedefs apply and to better see e.g. paths for when/must. Therefore I don't thing, that we need to print the compiled schema in YIN format, do you agree?

src/printer_yin.c
rkrejci: You print "extension_name; // Model comes from different input..." which is not XML. The `extension_name` is supposed to be XML element and that comment is supposed to be XML comment (`<!-- ... -->`)

rkrejci: The code has outdated, but it was about the TODO comment in `yprp_stmt()`

rkrejci: Note that the code has changed, so github does not show the code snippet correctly in Conversation tab, open  the `Files changed" tab.

fredgan: Oh...I just read in the conversation tab so I don't understand! Now I get it. Thanks

fredgan: Oh, I see. Yes it is. So maybe the data structure should be modified in some way.

fredgan:
I have handled all the comments refered.
Maybe we can reserve the compiled YIN output just now since we don't know if it is needed.

rkrejci:
> Maybe we can reserve the compiled YIN output just now since we don't know if it is needed.

I would remove it. 

Otherwise, we have to solve at least how to print those non-YIN elements like `<derived>` in identity. They must be placed into a different namespace. Actually I would prefer to put the compiled output completely into a different namespace, because there can be some other differences and we should avoid possibility of parsing this output as a standard YIN.

It seems to me as a big thing which can confuse users. Even the compiled YANG should be better explained to users. libyang 1.x (yanglint) has its `info` format to print information about the tree used by libyang. In libyang 2, I was planning actually to replace this `info` format by printing compiled YANG, so maybe we just need to stop talking about it as about YANG. It could be even in a separated printer, but since the output is very similar to YANG, it was useful to share the code. I don't see a reason to have another format for the same purpose. It is just more code that must be maintained.

fredgan:
Yes, exactly.  I agree with you.
I will remove the compile mode code in the YIN printer.

Now that we want to print the compiled info, maybe we can design it in another way besides the different name for distinguish the YANG. We shall define a complete description system for this. `derived` is just one of them.

From this sense, it's like the intermediate code in the compiler.

fredgan:
Now I have remove the compiled mode code.
But yanglint command line `-c` is still reserved for YANG format.
If you like, we could print the compiled mode using `compiled` as the parameter for `-f`.  I'm not sure it's an appropriate way.

src/printer_internal.h
```diff
@@ -29,6 +29,14 @@ typedef enum LYOUT_TYPE {
...
 } LYOUT_TYPE;
 
+/**
+ * @brief Types of the YANG printers
+ */
+enum schema_type {
+    YPR_PARSED,   /**< YANG printer of the parsed schema */
+    YPR_COMPILED  /**< YANG printer of the compiled schema */
+};
+
```
rkrejci: I think that now, when the compiled YIN printer was removed, this change should be reverted and the enum should be moved back into printer_yang.c as it was.

src/printer_yin.c
```diff
@@ -0,0 +1,1557 @@
...
+#define YIN_NS_URI "urn:ietf:params:xml:ns:yang:yin:1"
+
+/**
+ * @brief YIN printer context.
+ */
+struct ypr_ctx {
+    struct lyout *out;               /**< output specification */
+    unsigned int level;              /**< current indentation level: 0 - no formatting, >= 1 indentation levels */
+    const struct lys_module *module; /**< schema to print */
+    enum schema_type schema;         /**< type of the schema to print */
```
rkrejci: The type of the schema is no more needed, so remove it from the printer's context.

src/printer_yin.c
```diff
@@ -0,0 +1,1557 @@
...
+    LEVEL--;
+    ypr_close(ctx, "deviation", 1);
+}
+
+/**
+ * @brief Minimal print of a schema.
+ *
+ * To print
+ * a) compiled schema when it is not compiled or
+ * b) parsed when the parsed form was already removed
```
rkrejci: The comment should be modified since the compiled schema is not printed by this printer.

src/printer_yin.c
```diff
@@ -0,0 +1,1557 @@
...
+    ly_print_flush(ctx->out);
+
+    return LY_SUCCESS;
+}
+
+static void
+ypr_xmlns(struct ypr_ctx *ctx, const struct lys_module *module)
+{
+    unsigned int u;
+    char *space = "        ";
```
rkrejci: I don't thing this is a good idea to indent namespaces. This function will be used also for printing submodules, so the indentation can change.

src/printer_yin.c
```diff
@@ -0,0 +1,1557 @@
...
+
+    struct lysp_module *modp = module->parsed;
+
+    LY_ARRAY_FOR(modp->imports, u){
+        ly_print(ctx->out, "\n%s%*sxmlns:%s=\"%s\"", space, INDENT, modp->imports[u].prefix, modp->imports[u].module->ns);
+    }
+}
+
+
+struct ext_substmt_info_s stmt_attr_info[] = {
```
rkrejci: Not sure, @Dajvid isn't this information already somewhere in the YIN parser? Or is it just hard-coded in the YIN parser's functions?

Dajvid: In YIN parser, this information is hardcoded in functions that are responsible for parsing of particular elements. At that time I didn't find it useful to have it separated, but now, when it's also needed in yin printer, it would be better to have it all at one place in format that is suitable for both printer and parser. Later I definitely could make some changes in YIN parser to have this information shared, but for that purpose, it would be much better to use enum yin_argument (currently defined in parser_yin.h) for identification of argument instead of string (for example YIN_ARG_NAME instead of "name") otherwise, I would either have to again, slightly modify this table and thus duplicate this information in YIN parser or unnecessarily compare strings when processing each argument multiple times.

So I would suggest to slightly change this table to use mentioned enumeration (there is also macro yin_attr2str that maps enum to string (yin_attr2str(YIN_ARG_NAME) == "name") and when this pull request will be merged I will change YIN parser to also use information from this table.

fredgan: Yes, I see some in parser. But they are only half of them. I don't want to modify them in this PR in case of making too many changes in yin parser. Should I modify them here?

rkrejci: I would actually let it as it is for now. Looking more into it (checking where and for what the structure is used) make me feel that the change will be more complex and it would be better to make it separated. It seems that actually the name of the structure itself (`ext_substmt_info_s`) should be changed to reflect its changed purpose (because its original purpose was slightly different).

So I agree with the proposed changes, I would just postopne the changes to a separated task.

src/printer_yin.c
fredgan: I agree. But actually I don't find any API or access to print submodules. Unlike master, there are a `type` property in `lys_module`, `lys_module` here seems won't be submodule.

src/printer_yin.c
fredgan: Get it.

src/printer_yin.c
fredgan: Get it

src/printer_internal.h
fredgan: OK, I will revert this.

src/printer_yin.c
rkrejci: It is not yet supported even in YANG printer. And there are several other things regarding submodules.

fredgan: Then how should I modify  here from your point? Modify it like master branch or just make all the namespace in one line ?

rkrejci: All namespaces in one line are fine for me. However, I agree that putting them on a standalone line is better - what about an additional parameter of the function with the number of additional indentation spaces (caller will just count the length of the element name)? In the printing function, this value can be sum with the `INDENT`.

src/printer_yin.c
```diff
@@ -0,0 +1,1553 @@
...
+#include "printer_internal.h"
+#include "tree.h"
+#include "tree_schema.h"
+#include "tree_schema_internal.h"
+#include "plugins_types.h"
+#include "xpath.h"
+#include "xml.h"
+
+
+#define YIN_NS_URI "urn:ietf:params:xml:ns:yang:yin:1"
```
Dajvid: Same define is also in parser_yin.h, you could define it once in shared header file (maybe tree_schema_internal.h?) to eliminate duplication.

fredgan: Get it.

src/printer_yin.c
```diff
@@ -1229,10 +1226,15 @@ ypr_missing_format(struct ypr_ctx *ctx, const struct lys_module *module)
...
+ypr_xmlns(struct ypr_ctx *ctx, const struct lys_module *module, unsigned int indent)
 {
     unsigned int u;
-    char *space = "        ";
+    char *space;
+    if(asprintf(&space, "%*s", indent, "") == -1) {
+        LOGMEM(ctx->module->ctx);
+        ctx->out->status = LY_EMEM;
+        return;
+    }
```
rkrejci: Is this necessary? Isn't `indent + INDENT` in the `ly_print` calls sufficient (while the `space` parameter is removed)?

src/tree_schema_internal.h
```diff
@@ -22,6 +22,7 @@
 #include "tree_schema.h"
 #include "xml.h"
 
+#define YIN_NS_URI "urn:ietf:params:xml:ns:yang:yin:1"
```
rkrejci: Please add an indentation blank line between not connected macros

fredgan: OK

src/printer_yin.c
fredgan: Exactly, you are right.

Issue #918, opened 2019-10-25
unable to refine a case statement


kwatsen:
Using a `grouping` that contains a `choice` and trying to `refine` an `if-feature` statement into it fails to validate with a message like:

```
err : Invalid value "<snip2>" of "refine". (<snip2/>)
err : Refine substatements not applicable to the target-node. (<snip2, again)
```
Modifying the refine-path to target an arbitrary `container` node inside the `case` statement validates fine.

Note that [7.13.2](https://tools.ietf.org/html/rfc7950#section-7.13.2) says:
```
   o  A leaf, leaf-list, list, container, choice, case, anydata, or
      anyxml node may get additional "if-feature" expressions.
```

So, refining a `case` statement should be okay, right?

fredgan:
@kwatsen Thank you for your report. Yes, it is as what you said.

@rkrejci Hi Radek, I find that this issue is caused by that the parse part is the implement of YANG1.0(RFC6020), But after Aug 2016, YANG1.1 (RFC7950) is published. The `refine` ability is strenghened. `if-feature` function is added to `refine` in YANG1.1 and some other functions.

kwatsen:
Fix confirmed.  Thanks!

PR #920, opened 2019-10-29
refine BUGFIX refine support if-feature statement

fredgan:
fix issue #918


rkrejci:
Hi, could you please also add a test case? I§m wondering if the same issue is also in YIN parser or not.

fredgan:
> Hi, could you please also add a test case? I'm wondering if the same issue is also in YIN parser or not.

@rkrejci  You are right. YIN parser also have some part of this issue. It is fixed now.

rkrejci:
Merged manually.

PR #943, opened 2019-11-14
lyb parser BUGFIX when model count is zero don't call malloc()

barbarost:
malloc(0) can return either NULL or a unique pointer value, so
it is not an error when malloc(0) returns NULL


michalvasko:
Hi,
you are right, an empty file does not have any models. Merging.

Regards,
Michal

PR #948, opened 2019-11-17
Print info

rkrejci:
Print compiled YANG as (instead of) info format. There is also a possibility to print information about a single node or a subtree. What I'm not sure about is format of the schema path - the initial implementation requires full schema module names to be used as prefixes. As an alternative, it would be passible to use import prefixes refering to a specific schema module.

### DIFF
--- a/src/printer_schema.h
+++ b/src/printer_schema.h
@@ -27,7 +27,7 @@
  *
  * @{
  */
-#define LYS_OUTPUT_NO_SUBST          0x10 /**< Print only top-level/referede node information,
+#define LYS_OUTPUT_NO_SUBSTMT        0x10 /**< Print only top-level/referede node information,
                                                do not print information from the substatements */
 //#define LYS_OUTOPT_TREE_RFC        0x01 /**< Conform to the RFC TODO tree output (only for tree format) */
 //#define LYS_OUTOPT_TREE_GROUPING   0x02 /**< Print groupings separately (only for tree format) */
--- a/src/printer_schema.h
+++ b/src/printer_schema.h
@@ -18,6 +18,7 @@
 #include <stdio.h>
 #include <unistd.h>
 
+#include "printer.h"
 #include "tree_schema.h"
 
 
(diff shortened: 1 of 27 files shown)


src/printer_schema.h
```diff
@@ -20,6 +20,25 @@
...
 #include "tree_schema.h"
 
+
+/**
+ * @defgroup schemaprinterflags Schema output options
+ * @ingroup schema
+ *
+ * @{
+ */
+#define LYS_OUTPUT_NO_SUBST          0x10 /**< Print only top-level/referede node information,
```
fredgan: I suggest  using  `LYS_OUTPUT_NO_SUBSTMT` in order to be consistent with other place.

src/tree_schema.c
```diff
@@ -184,6 +184,73 @@ lys_getnext(const struct lysc_node *last, const struct lysc_node *parent, const
...
+                        struct lys_module *m = (struct lys_module *)ctx->list.objs[u];
+                        if (mod) {
+                            if (m->implemented) {
+                                mod = m;
+                                break;
+                            } else if (m->latest_revision) {
+                                mod = m;
+                            }
+                        } else {
+                            mod = m;
```
fredgan: Maybe these three `mod=m;` can be put outside of this.

tools/lint/commands.c
```diff
@@ -58,7 +58,7 @@ cmd_clear_help(void)
 void
 cmd_print_help(void)
 {
-    printf("print [-f (yang | yin | tree [<tree-options>] | info [-P <info-path>] | jsons)] [-o <output-file>]"
+    printf("print [-f (yang | yin | tree [<tree-options>] | info [-P <info-path>] [-(-s)ingle-node])] [-o <output-file>]"
```
fredgan: I remember that in the master `-s` means `--strict`
I don't think change the user's habit is a good way.

src/tree_schema.c
```diff
@@ -184,6 +184,73 @@ lys_getnext(const struct lysc_node *last, const struct lysc_node *parent, const
...
+        }
+        /* precess ".." in relative paths */
+        while (!strncmp("../", id, 3)) {
+            id += 3;
+            if (!node) {
+                LOGERR(ctx, LY_EINVAL, "Invalid qpath \"%s\" - too many \"..\" in the path.", qpath);
+                return NULL;
+            }
+            node = node->parent;
+        }
```
fredgan: I wondered whether `./` should be parsed in this function?

src/printer_schema.h
rkrejci: agreed

src/tree_schema.c
rkrejci: no, I'll add it

rkrejci: well, the question is if it should be supported. There is no path definition in the YANG ABNF grammar (RFC 7950) that would support `./`, so at the end I would prefer to avoid supporting it

tools/lint/commands.c
rkrejci: But there it is for parsing, this is printing functionality. Strict in printing have no meanings.

fredgan: Exactly. However, I think  the upper case `-S` may be better.

rkrejci:
For the info printer, I will need to extend printer functions with some additional parameters (path of the node and some others). With the current API, I would have to add some additional functions next to the `lys_print_mem()`, `lys_print_clb()`, `lys_print_file()`, `lys_print_fd()` and `lys_print_path()` or extend them with additional parameters. Some of the parameters would depend on others - simply the API would be much more complex/complicated. So, redesigning printer API seemed to me as a better approach. Now, there is a single printer function requiring a printer output type handler to be created and provided via `lyp_new_*()` functions. And the handler can be used repeatedly, which is great mainly in case of writing into the memory chunk, so it is now possible to concatenate several printed items into a single memory chunk.

But there are 2 questions:
1. should be the current printer functions kept for backward compatibility? They would be just wrappers around the new concept, on the other hand they would extend API and not sure how clear would be using the different approaches for users
2. there is idea to do the same change also for parsers, so instead of a set of functions `lys_parse_mem()`, `lys_parse_...`, there would be just `lys_parse()` with input handler. Furthermore, input and output handle could be somehow unified behind a single structure.

What do you think? @michalvasko @fredgan @jktjkt @lperkov

michalvasko:
I think it would be definitely a positive change if *libyang* could be compressed into fewer functions. That should make the API smaller and easier to use (including bindings).

1. Keeping them as deprecated for some time may be easier for migrating current *libyang* applications but at the same time defeats the purpose I mentioned before (even though only temporarily) so I am not sure.
2. This seems like a good idea to me because of the same reason of making API smaller.

Regards,
Michal

jktjkt:
> What do you think?

I don't have a favorite horse in this race, but perhaps @syyyr (who has worked much more with libyang than me) can share his feedback.

michalvasko:
As for the change in parser, there are now additional arguments for:

  2. There would not be one general parsing function, but 2 - one for parsing full datastore, another for parsing operations (RPC/action/rpc-reply/notification). This seems like a good change because it would probably allow to get rid of variable arguments and would also make the code of each function cleaner than it is when the functionality is in one function.

fredgan:
Sorry for many notifications so that I didn't noticed this. I will be more careful about the notifications later.
From my experience,  Too many APIs for the same (or almost the same) function is not friendly because they have to read all the APIs, at least they have to know that there are so many APIs. So I support this change.

rkrejci:
So only the new API (`lys_print()` and `lyd_print()` with the output setting functions) is present, the previous API functions for specific output type were removed.

I'm going also to make similar changes to the parsers API, but it will be in a separated PR.

Issue #774, opened 2019-04-30
Double free or corruption in lys_parse_path


jvijtiuk:
The following file causes a double free.
The value is freed at tree_schema.c:2891 and seems to be allocated at parser.c:624.

```
module state-lists {
    yang-version 1.1;
    namespace "urn:state-lists";
    prefix sl;

    container cont {
		config false;
		grouping group1 {
            leaf leaf3 {
     			type tdef2 {
      				length "3..9 | 30..40";
      				pattern "[ac
				}*";
    		}

    		units "none";
    		default "aaa";
  		}

		typedef tdef2 {
			type string {
				length "2..17 | 20..50";
				pattern "[ab]*";
			}
		}

  		container cont1 {
    		uses group1 {
      			if-feature "feat2";
      			refine "leaf1" {
        			if-feature "feat3";
				    must "24 - 4 = number('20')";
        			default "25";
        			config true;
        			mandatory false;
        			description "dsc";
        			reference "none";
      			}
    		}

			leaf leaf4 {
				type int64 {
					range "1000 .. 50000" {
						error-message
						"Special e        
					}
					.";
				}
			}
    	}

      }
    }
  }
}
```

jvijtiuk:
This file also causes a hang in yanglint on the libyang2 branch.

michalvasko:
Hi Juraj,
we have decided continuing to fuzz the current YANG parser does not make much sense, so it would be great if you could use the one from `libyang2` branch. Also, best way of sharing binary modules is to attach them here directly (zipped maybe).

Regards,
Michal

jvijtiuk:
Ok, I will set up the fuzzer to fuzz the libyang2 parser, and test the remaining bugs the fuzzer found with the new parser.

Regards,
Juraj

rkrejci:
should be fixed in libyang2

jvijtiuk:
libyang2 now successfully reports that the input is not valid, so I'm closing the issue.

PR #951, opened 2019-11-19
tests/fuzz: New AFL target: function lyd_parse_path()

ikrumlova:
For each of the three supported formats: JSON, XML, LYB.
The JSON version crashes very quickly, xml after some time too.

### DIFF
--- a/tests/fuzz/CMakeLists.txt
+++ b/tests/fuzz/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 2.8.12)
 
-set(fuzz_targets yangfuzz xmlfuzz)
+set(fuzz_targets yangfuzz xmlfuzz jsonfuzz lybfuzz xml_lyd_parse_path_fuzz)
 
 foreach(target_name IN LISTS fuzz_targets)
     add_executable(${target_name} ${target_name}.c)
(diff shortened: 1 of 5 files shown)


jvijtiuk:
Hello ikrumlova,

as the fuzzer harnesses are identical except for the lys_parse_path argument, maybe the wrappers could be unified into a single fuzz target that receives a command line argument which determines which file format to parse, so if you want you can unify the wrappers and update this PR.

It would be great if you could report the bugs that have been found, so that libyang2 can be checked for the same issues. 

Also, in #774 it was mentioned that only libyang2 should be fuzzed, from then on, so if that is still the case and you are fuzzing libyang I guess it would be best if you could fuzz only on the ```libyang2``` branch.

@rkrejci, @michalvasko, do you have any comments on this?

Regards,
Juraj

rkrejci:
So far, I didn't have time to check the pull request, sorry.

However I'm fine even with fuzzing current master/devel. In this case, it is about fuzzing data formats parsers and it make sense. What I wanted to avoid is fuzzing YANG parser in master - we have very limited resources to fix it, so the schema formats should be really fuzzed in libyang2 branch.

michalvasko:
Hi @ikrumlova,
actually, previously we were fixing problems uncovered by fuzzing in separate issues. It would be great if you could create issue with the input file that caused the crash, it significantly simplifies fixing it.

Regards,
Michal

pspacek:
Sorry, this slipped through cracks, we will have a look.

jvijtiuk:
Hello @pspacek, @ikrumlova,

have there been any updates regarding the JSON and XML crashes that were found?

Regards,
Juraj

pspacek:
Sorry, there is no progress, we have to focus on other priorities. I would be more than happy if anyone else can take this forward.

michalvasko:
Not needed for *libyang* 1 anymore.

PR #954, opened 2019-11-19
fixed issue printing xml where namespace was printed multiple times on anydata

pschoenm:
when output xml and the data structure is anydata, with multiple namespaces, in addition to using a single namespace, the printer would output the same xmlns multiple times.

-- created test to verify fix
-- valgrind passes all tests

### DIFF
--- a/src/printer_xml.c
+++ b/src/printer_xml.c
@@ -55,8 +55,8 @@ modlist_add(struct mlist **mlist, const struct lys_module *mod)
         LY_CHECK_ERR_RETURN(!iter, LOGMEM(mod->ctx), EXIT_FAILURE);
         iter->next = *mlist;
         iter->module = (struct lys_module *)mod;
-	iter->printed = 0;
-	*mlist = iter;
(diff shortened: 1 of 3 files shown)


src/printer_xml.c
```diff
@@ -54,19 +55,30 @@ modlist_add(struct mlist **mlist, const struct lys_module *mod)
         LY_CHECK_ERR_RETURN(!iter, LOGMEM(mod->ctx), EXIT_FAILURE);
         iter->next = *mlist;
         iter->module = (struct lys_module *)mod;
-        *mlist = iter;
+	iter->printed = 0;
+	*mlist = iter;
```
fredgan: Still a TAB  not spaces for indent.
And some other line for the same issue.

pschoenm: Sorry about that.  I did double check for tabs.  should be fixed now.

pschoenm:
I corrected the valgrind test.  The test now runs cleanly.

PR #960, opened 2019-11-26
libyang2: yang parser BUGFIX null dereference in moveto_resolve_model

jvijtiuk:
Hello,

this should fix a null dereference found while fuzzing libyang2. 

The following file caused the crash:
```
module d{
	namespace "";
	prefix d;
	leaf f {
		type string;
		must ":e";
		default "";
	}
}
```

The issue was in moveto_resolve_model in src/xpath.c, a null mod was returned from lys_module_find_prefix, which then caused a null dereference when !mod->implemented was checked.

This commit fixes the issue and doesn't break any tests on my machine.

Regards,
Juraj

### DIFF
--- a/src/xpath.c
+++ b/src/xpath.c
@@ -5232,14 +5232,12 @@ moveto_resolve_model(const char **qname, uint16_t *qname_len, struct lyxp_set *s
             LOGINT_RET(set->ctx);
         }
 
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
you are right, thanks.

Regards,
Michal

PR #965, opened 2019-11-29
fix the bug about rpc and notification with the same name.

fredgan:
Hi,
As description in [RFC7950, Chapter 6.2.1](https://tools.ietf.org/html/rfc7950#section-6.2)
```
   o  All leafs, leaf-lists, lists, containers, choices, rpcs, actions,
      notifications, anydatas, and anyxmls defined (directly or through
      a "uses" statement) within a parent node or at the top level of
      the module or its submodules share the same identifier namespace.
      This namespace is scoped to the parent node or module, unless the
      parent node is a case node.  In that case, the namespace is scoped
      to the closest ancestor node that is not a case or choice node.
```
I use the following yang file to test and find a bug.
`mod1.yang:`
```
module mod1 {
  namespace "urn:mod1";
  prefix m1;

  rpc aaa {
    input {
      leaf le1 {
        type string;
      }
    }
  }
  
  notification aaa {
    leaf le2 {
      type string;
    }
  }
}
```
`the expected result is that:`
```
~/libyang/build$ ./yanglint -f yang mod1.yang
err : Duplicated notification identifier "aaa". (/mod1:aaa)
err : Module "mod1" parsing failed.
```
But the actual result is no error.

### DIFF
--- a/src/tree_schema.c
+++ b/src/tree_schema.c
@@ -660,6 +660,8 @@ lys_check_id(struct lys_node *node, struct lys_node *parent, struct lys_module *
     case LYS_LIST:
     case LYS_CONTAINER:
     case LYS_CHOICE:
+    case LYS_RPC:
+    case LYS_NOTIF:
     case LYS_ANYDATA:
         /* 6.2.1, rule 7 */
         if (parent) {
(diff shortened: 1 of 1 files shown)


michalvasko:
Hi,
thanks, it seems fine except as written in the RFC, please also add `action` to the checked statements.

Regards,
Michal

michalvasko:
Hi,
okay, now it seems good, thanks.

Regards,
Michal

PR #973, opened 2019-12-11
libyang2: compiled schema BUGFIX when validating empty leafref path

jvijtiuk:
Hello,

the following file causes a null pointer dereference in src/tree_schema_compile.c:2736, because the leafref path is NULL. The issue is that the while loop validating the path runs only as long as *id is a non string terminating character. If id is an empty path the loop never runs, and later on node->nodetype is checked, even though node was never set, and is NULL.

Here is a file that causes the crash:
```
module p{
	namespace "";
	prefix p;

	leaf mgmt-interface {
		type leafref {
			path "";
		}
	}
}
```

This commit adds a check for empty leafref paths which fixes the issue and doesn't break any tests.

Regards,
Juraj

### DIFF
--- a/src/tree_schema_compile.c
+++ b/src/tree_schema_compile.c
@@ -2674,6 +2674,12 @@ lys_compile_leafref_validate(struct lysc_ctx *ctx, struct lysc_node *startnode, 
 
     iter = 0;
     id = leafref->path;
+
+    if (!*id) {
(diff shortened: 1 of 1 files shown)


jvijtiuk:
I see that this decreases test coverage. Sorry for that, I'll add a test case for this tommorow.

michalvasko:
Hi,
coverage is not that important but you can add a test and we will merge it. Thanks anyway.

Regards,
Michal

PR #977, opened 2019-12-13
LOGBUG should be called before free

mruprich:
LOGBUF which dereferences exp should be called prior to freeing exp.


michalvasko:
Hi,
of course, thanks.

Regards,
Michal

PR #1016, opened 2020-01-31
Change Value::dec64 to also return fraction digits

syyyr:
Right now, if you only have `libyang::Value` and call `dec64()` on it, you only get the value, but not the fraction digits, the value is useless by itself. This patch changes the return value to a struct that has both the value and the fraction digits.

I'm not sure if a simple approach like this is a good idea, tell me what you think.

cc @jktjkt

### DIFF
--- a/swig/cpp/src/Tree_Data.hpp
+++ b/swig/cpp/src/Tree_Data.hpp
@@ -37,6 +37,12 @@ namespace libyang {
  * Class wrappers for data structures and functions to manipulate and access instance data tree.
  */
 
+struct Decimal64
+{
+    int64_t value;
+    uint8_t digits;
+};
+
 /**
  * @brief class for wrapping [lyd_val](@ref lyd_val).
  * @class Value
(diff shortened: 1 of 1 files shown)


swig/cpp/src/Tree_Data.hpp
```diff
@@ -37,6 +37,12 @@ namespace libyang {
  * Class wrappers for data structures and functions to manipulate and access instance data tree.
  */
 
+struct Decimal64
+{
+    int64_t value;
+    uint8_t digits;
+};
+
```
michalvasko: Not sure what the C++ convention is but I would expect a `pair` instead of this structure. @jktjkt ?

jktjkt: I think that a custom `struct` is better in this context because it gives a clear meaning to both values of that tuple. If it was a `std::pair`, the usage would look like, e.g.:
```c++
auto v = val->dec64();
double res = v.first * std::pow(10, -v.second);
```

I think that the following is more readable:
```c++
auto v = val->dec64();
double res = v.value * std::pow(10, -v.digits);
```

michalvasko: Alright.

PR #1025, opened 2020-03-03
libyang2: tree schema BUGFIX check for NULL extension definition before free

jvijtiuk:
Hello,

this PR fixes a NULL dereference in tree_schema_free.c

The following YANG file causes the crash:
```
module a {
	yang-version 1.1;
  	namespace "urn:all";
  	prefix all_mod;

  	grouping group1 {
    	leaf leaf1 {
       		type int64 {
        		range "1000 .. 50000" {
          			error:message
            		"Spec";
        		}
      		}
    	}
    }
}
```

The issue appeared due to no NULL check for ext->def when calling lysc_extension_free from lysc_ext_instance_free. The only other place where lysc_extension_free is called from is lysp_ext_free, and a NULL check is done there, so the issue can't appear there.

Regards,
Juraj


michalvasko:
Hi,
right, thanks, merging.

Regards,
Michal

PR #1035, opened 2020-03-18
libyang2: fuzz harness refactor, with support for LLVM's LibFuzzer

jvijtiuk:
Hello,

This PR includes a refactor of the fuzzing harness to use LibFuzzer's LLVMFuzzerTestOneInput. 
This enables standalone fuzzing with LibFuzzer, which doesn't require a main function, and uses LLVMFuzzerTestOneInput directly. Fuzzing with AFL also uses the LLVMFuzzerTestOneInput, by using a main wrapper which calls LLVMFuzzerTestOneInput.

LibFuzzer (http://llvm.org/docs/LibFuzzer.html) is an in-process LLVM fuzzing engine library. The fuzzing engine is based on similar principles that AFL is based on, but has some implementation differences, so it often reveals different bugs. Another advantage is that LibFuzzer has official working support for UBSAN, while AFL has no official UBSAN support.

The changes in this PR should allow somewhat easier addition of new harnesses, as now only a target file with the LLVM fuzzing function has to be added. The equivalent of the old yangfuzz harness is now in tests/fuzz/lys_parse_mem.c. 

I've also added 2 other targets as examples, one which fuzzes buf_add_char, and one which fuzzes parse_yang_module. The idea is to eventually have different functions separated into different harnesses, e.g. to separate lys_parse_mem into parse_yang_module and lys_compile, and so on. I've updated the root README and tests/fuzz/README, with two example asciinema videos, one for LibFuzzer and one for AFL. As a side effect, fuzzing smaller parts of the codes obviously achieves faster execution speeds. For example fuzzing parse_yang_module has on average been nearly two times faster than fuzzing lys_parse_mem.

I've also changed the AFL_LOOP value for the AFL main wrapper, after some testing with dedicated hardware, which should also increase the execution speed.

If anyone has any comments or suggestions regarding this, and if something should be changed, please let me know.

Regards,
Juraj


michalvasko:
Hi,
seems ok, merging.

Regards,
Michal

PR #1055, opened 2020-04-03
package fixes for CentOS 6 and 7

rwestphal:
CentOS 6 needs all the language bindings turned off and the cache
disabled as it doesn't have the needed swig version or the pcre_* functions

CentOS 7 needs swig3 package instead of swig package


michalvasko:
Hi,
you support CentOS 6? Well, as long as only these changes are needed I guess it is okay. Thanks, merging.

Regards,
Michal

michalvasko:
@rwestphal I had to revert these changes prior to the latest release because they caused the bindings not to be build on any RPM distribution we have OBS running on. I have not noticed this before because those packages were already built and could be used for other projects.

PR #1068, opened 2020-04-23
rework "local" Debian packaging scripts

eqvinox:
change local-deb to use files from the debian branch instead of having its own copy that would get out of date / have differences with the "official" Debian package


michalvasko:
Hi,
this generally seems like a reasonable idea but will probably be impossible to put into practice. The main problem is that this will break our OBS Debian *libyang* packages. That is not really a problem of its own but all the other packages (*libnetconf2*, *sysrepo*, *netopeer2*) depend on this one so the package must be build at least to satisfy their dependencies.

So, unless there is a way of building the "official" package using OBS, I think we will have to continue with the current setup and have 2 versions of Debian packages.

Regards,
Michal

PR #1078, opened 2020-05-06
{uint16_t} -> {LY_DATA_TYPE} value_type()

andrei-pavel:
Hi,
I get compiler errors reminding me to `static_cast<LY_DATA_TYPE>()` every `value_type()`. Any reason why it needs to be `uint16_t`?
Regards!


michalvasko:
Hi,
nope, I think it is a leftover from when `value_type` actually was `uint16_t`, it is not for quite some time. Merging.

Regards,
Michal

PR #1085, opened 2020-05-12
C++: Allow constructing Submodule from Module

syyyr:
I couldn't find a way to get info about a submodule (like the `belongsTo` variable of `lys_submodule`), so this patch adds a way to convert `libyang::Module` to `libyang::Submodule` to get info about a submodule.

The constructor follows the constructors of the `libyang::Schema_Node_*` family, taking in an `S_Module` and just assigning the private variables. I'm not sure if the deleter should just be assigned like this, but I don't get any memory errors.

### DIFF
--- a/swig/cpp/src/Tree_Schema.hpp
+++ b/swig/cpp/src/Tree_Schema.hpp
@@ -111,6 +111,7 @@ public:
 
     friend Context;
     friend Data_Node;
+    friend Submodule;
 
 private:
     struct lys_module *module;
@@ -126,6 +127,8 @@ class Submodule
 public:
     /** wrapper for struct [lys_submodule](@ref lys_submodule), for internal use only */
     Submodule(struct lys_submodule *submodule, S_Deleter deleter);
+    /** creates a Submodule from `module` if it is a submodule */
+    Submodule(S_Module module);
     ~Submodule();
     /** get ctx variable from [lys_submodule](@ref lys_submodule)*/
     S_Context ctx() LY_NEW(submodule, ctx, Context);
--- a/swig/cpp/src/Tree_Schema.cpp
+++ b/swig/cpp/src/Tree_Schema.cpp
@@ -39,6 +39,14 @@ Submodule::Submodule(struct lys_submodule *submodule, S_Deleter deleter):
     submodule(submodule),
     deleter(deleter)
 {};
+Submodule::Submodule(S_Module module):
+    submodule((lys_submodule*)module->module),
+    deleter(module->deleter)
+{
+    if (module->type() != 1) {
+        throw std::invalid_argument("Attempted to cast a YANG module into a YANG submodule");
+    }
+}
 std::vector<S_Schema_Node> Module::data_instantiables(int options) {
     std::vector<S_Schema_Node> s_vector;
     struct lys_node *iter = NULL;


swig/cpp/src/Tree_Schema.hpp
```diff
@@ -126,6 +127,15 @@ class Submodule
...
     Submodule(struct lys_submodule *submodule, S_Deleter deleter);
+    /** creates a Submodule from `module` if it is a submodule */
+    Submodule(S_Module module):
+        submodule((lys_submodule*)module->module),
+        deleter(module->deleter)
+    {
+        if (module->type() != 1) {
+            throw std::invalid_argument("Type of module must be 1 (submodule)");
+        }
+    }
```
jktjkt, line 138: I think this should go to the `.cpp` file -- at least the other constructor lives there.

swig/cpp/src/Tree_Schema.hpp
```diff
@@ -126,6 +127,15 @@ class Submodule
 public:
     /** wrapper for struct [lys_submodule](@ref lys_submodule), for internal use only */
     Submodule(struct lys_submodule *submodule, S_Deleter deleter);
+    /** creates a Submodule from `module` if it is a submodule */
+    Submodule(S_Module module):
+        submodule((lys_submodule*)module->module),
+        deleter(module->deleter)
+    {
+        if (module->type() != 1) {
+            throw std::invalid_argument("Type of module must be 1 (submodule)");
```
jktjkt, line 136: I think that this would be a more useful error message: `Attempted to cast a YANG module into a YANG submodule`

swig/cpp/src/Tree_Schema.hpp
```diff
@@ -126,6 +127,15 @@ class Submodule
 public:
     /** wrapper for struct [lys_submodule](@ref lys_submodule), for internal use only */
     Submodule(struct lys_submodule *submodule, S_Deleter deleter);
+    /** creates a Submodule from `module` if it is a submodule */
+    Submodule(S_Module module):
+        submodule((lys_submodule*)module->module),
+        deleter(module->deleter)
+    {
+        if (module->type() != 1) {
```
jktjkt, line 135: Not commenting on this patch on its own, but I really dislike the way how this is done. If only that `type` member was named more descriptively, perhaps something like `is_submodule`, for example. Compare this:
```C
if (node->module->type) {
    ly_print(out, ",\"included-from\":\"%s\"", node->module->name);
}
```
with this:
```C
if (node->module->is_submodule) {
    ly_print(out, ",\"included-from\":\"%s\"", node->module->name);
}
```

swig/cpp/src/Tree_Schema.hpp
syyyr: done

swig/cpp/src/Tree_Schema.hpp
syyyr: done

swig/cpp/src/Tree_Schema.hpp
syyyr: To be completely honest, I dislike a lot of things about the entire C++ bidings (especially how class are casted to each other), but I was trying to mimic what's already there - `Schema_Node_*` ctors are also defined in the header file (but no reason to mimic "bad" code I guess). About `::type`: it's called that even in the C struct, so no way of changing that, without breaking the API, but in C++, there could be a method which would do the same thing, but would be named `is_submodule`.

jktjkt: Yeah, my comment was meant as feedback to libyang maintainers.

michalvasko: There are a lot of things wrong about these C++ bindings mostly because they were not written by the authors but it will stay the way it is in libyang version 1.x and will be completely rewritten for libyang 2.x.

michalvasko:
Seems fine to me, too, merging. Thanks.

Issue #1044, opened 2020-03-22
libyang2: stack buffer overflow and undefined behaviour when passing '\x0a' to lys_parse_mem


jvijtiuk:
Hello,

the following file causes multiple issues on the libyang2 branch:

```
#include <stdio.h>
#include <stdlib.h>
#include <stdbool.h>

#include <libyang/libyang.h>

int LLVMFuzzerTestOneInput(uint8_t const *buf, size_t len)
{
	struct ly_ctx *ctx = NULL;
	LY_ERR err;

	err = ly_ctx_new(NULL, 0, &ctx);
	if (err != LY_SUCCESS) {
		fprintf(stderr, "Failed to create context\n");
		exit(EXIT_FAILURE);
	}

	lys_parse_mem(ctx, buf, LYS_IN_YANG);
	ly_ctx_destroy(ctx, NULL);
	return 0;
}

int main() {
	char buf[1] = {0xa};
	LLVMFuzzerTestOneInput(buf, 1);
}
```

I've compiled it with ASAN and UBSAN, and linked it to libyang compiled with ASAN and UBSAN, using the recent LibFuzzer changes. The command to build libyang like that was:
```cmake -DCMAKE_C_COMPILER=clang -DCMAKE_BUILD_TYPE=Debug -DENABLE_FUZZ_TARGETS=ON -DFUZZER=LibFuzzer .```

The command to build the .c file that causes the issue was:
```
clang main.c -lyang -fsanitize=address,undefined
```

Here is the ASAN and UBSAN output generated by the binary, in a file:

[asan-ubsan-out.txt](https://github.com/CESNET/libyang/files/4366268/asan-ubsan-out.txt)

If there are any issues with reproducing this, please let me know. I'm also not entirely sure if this overflow is really a libyang issue, as it depends on the user passing in a non null terminated string, in this case just 0xa. The undefined behaviour however appears for nearly everything that I run when i compile libyang with UBSAN, but it mostly seems to appear at various FREEs.

Regards,
Juraj

rkrejci:
There are probably 2 separate things here. First, libyang is not able to protect against not NULL-terminated strings on input. It would require getting the correct length of the input from a user and right now I'm not for such a change.

[...]

PR #1086, opened 2020-05-13
strict aliasing of sized arrays

rkrejci:
The patch fixes AddressSanitizer's warnings reported in #1044.

It should work correctly for 64b and fewer bits architectures. The change affects all the sized arrays, so it increases memory consumption (each sized array is now bigger by 4 bytes). It would be possible to detect the size of the pointer on the architecture where we build, but I'm afraid that this variability could cause interoperability problems when transferring data between architectures (maybe the LYD format). So this functionality is not present now but it can be easily added if needed.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -28,6 +28,7 @@
 #include "hash_table.h"
 #include "log.h"
 #include "set.h"
+#include "tree.h"
 
 struct ly_ctx;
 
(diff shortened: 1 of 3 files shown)


michalvasko:
As for the changes themselves, they seem fine to me. Also, I believe making the size variable should not generally cause any problems. These arrays are always only directly in memory and LYB uses its own structures for everything so that should be completely platform independent. But, there may possibly be some issues when cross-compiling, which could be solved by allowing to set the variable size manually using `cmake` or perhaps even possible to be automatically adjusted because the target platform must be known during compilation.

Regards,
Michal

PR #1087, opened 2020-05-18
New python bindings

rjarry:
Hi @michalvasko @rkrejci @Dajvid,

As promised, here is a first draft of the integration of https://github.com/rjarry/libyang-cffi.

Few notes:

- I tried to fit in with cmake as much as possible but I am not very proficient with it. There may be atrocities to fix.
- Python distutils/setuptools command line interface is horrible. The only "sane" way to ensure that arguments are taken into account is to use a custom "user configuration file" (generated `.pydistutils.cfg`) and force a value to `HOME` during the build so that distutils/setuptools load it. This approach was borrowed from [Piotr Ożarowski's pybuild](https://salsa.debian.org/python-team/tools/dh-python/-/blob/master/dhpython/build/plugin_distutils.py#L32) which is used extensively in Debian and Ubuntu to generate Python .deb packages.

What is missing:

- Write actual examples
- More unit tests?
- API functions not exported
- Python doc strings
- Integration in the `packages` subfolder (for automatic upload to [pypi.org](https://pypi.org/project/libyang/))

What do you think?


python/cffi/source.c
```diff
@@ -0,0 +1,303 @@
...
+{
+	return ly_errno;
+}
+
+static void lypy_set_errno(LY_ERR err)
+{
+	ly_errno = err;
+}
+
+static int
```
rjarry, line 130: As I said, there is some non-API C code that was needed into the python binding. I copied it since I had no way to call it directly. However, it would be best to leave this code into `libyang.so` and make it available in the API somehow.

python/cffi/source.c
```diff
@@ -0,0 +1,303 @@
...
+                return 0;
+            }
+        }
+
+        /* there is no case child that will be printed, print this node */
+        return 1;
+    }
+
+    return 1;
+}
```
rjarry, line 296: As I said, there is some non-API C code that was needed into the python binding. I copied it since I had no way to call it directly. However, it would be best to leave this code into `libyang.so` and make it available in the API somehow.

rjarry:
As I expected, the static build does not work...

I'll disable it for the time being.

packages/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 set(PACKAGE "libyang")
 set(CPP_PACKAGE "libyang-cpp")
-set(PYTHON_PACKAGE "python3-yang")
+set(PYTHON_PACAKGE "python3-libyang")
```
michalvasko, line 3: Do we want to change the name? Couldn't we keep it and just increase the version or something?

rjarry: Just saw the typo :)

About the name, I'm not sure. I find it confusing. I'll revert the previous name.

packages/CMakeLists.txt
```diff
@@ -22,9 +22,8 @@ configure_file(${PROJECT_SOURCE_DIR}/packages/debian.${CPP_PACKAGE}.install
                ${PROJECT_BINARY_DIR}/build-packages/debian.${CPP_PACKAGE}.install COPYONLY)
 configure_file(${PROJECT_SOURCE_DIR}/packages/debian.${CPP_PACKAGE}-dev.install
                ${PROJECT_BINARY_DIR}/build-packages/debian.${CPP_PACKAGE}-dev.install COPYONLY)
-# no python package for Debian because there is only SWIG 3.10 on Debian 9 :-/
-#configure_file(${PROJECT_SOURCE_DIR}/packages/debian.${PYTHON_PACKAGE}.install
-#               ${PROJECT_BINARY_DIR}/build-packages/debian.${PYTHON_PACKAGE}.install COPYONLY)
+configure_file(${PROJECT_SOURCE_DIR}/packages/debian.${PYTHON_PACKAGE}.install
+               ${PROJECT_BINARY_DIR}/build-packages/debian.${PYTHON_PACKAGE}.install COPYONLY)
```
michalvasko, line 26: Is everything required (mostly CFFI I guess) available on all supported Debians and Debian-based distributions? Debian 9 and Ubuntu 16.04 are the oldest supported.

packages/CMakeLists.txt
rjarry: I meant "confusing" because the python package name is `libyang`.

python/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+unset(PYTHON_EXECUTABLE CACHE)
+unset(PYTHON_INCLUDE_DIR CACHE)
+
+if(${GEN_PYTHON_VERSION} STREQUAL "2")
+    find_package(PythonInterp 2 REQUIRED)
+    find_package(PythonLibs 2 REQUIRED)
+    if(NOT PYTHONLIBS_FOUND)
+        message(WARNING "Did not found Python version 2.x")
+        message(STATUS "Libyang supports Python 2.x and Python 3.x")
+    endif()
```
michalvasko, line 12: So even Python 2 is supported?

packages/CMakeLists.txt
rjarry: Yes I guess CFFI has been present in debian/ubuntu for a long time.

https://packages.debian.org/stretch/python3-cffi
https://packages.ubuntu.com/xenial/python3-cffi

The version is a bit old but it "should" work.

python/CMakeLists.txt
rjarry: Yes since https://github.com/rjarry/libyang-cffi supports it. However, since Python 2 is officially EOL since the begining of 2020, I was not sure if you wanted to keep official Python 2 support.

packages/CMakeLists.txt
michalvasko: I do not really know so I wanted to discuss it, @rkrejci?

packages/CMakeLists.txt
michalvasko: Okay, just checking.

python/CMakeLists.txt
michalvasko: Only for some backwards compatibility, I guess, but it is better if @rkrejci decided this.

michalvasko:
So `source.c` contains all the internal functions you needed directly in the bindings?

rjarry:
> So `source.c` contains all the internal functions you needed directly in the bindings?

Yes. Most of them are trivial and could be implemented directly in python. However, for the sake of performance, I did some of the work in C. Maybe there are some that could be integrated in the C library:

https://github.com/CESNET/libyang/pull/1087/files#diff-8b2dff95a233a8f66dea33417af39c83R9
https://github.com/CESNET/libyang/pull/1087/files#diff-8b2dff95a233a8f66dea33417af39c83R43
https://github.com/CESNET/libyang/pull/1087/files#diff-8b2dff95a233a8f66dea33417af39c83R105

The most critical one (I guess) is `ly_toprint()` which is mandatory for the python `dict` data format. Would it be possible to export this function in the API? Or maybe another version of it.

https://github.com/CESNET/libyang/pull/1087/files#diff-8b2dff95a233a8f66dea33417af39c83R251
https://github.com/CESNET/libyang/pull/1087/files#diff-8b2dff95a233a8f66dea33417af39c83R131
https://github.com/CESNET/libyang/pull/1087/files#diff-bb57130e2bf44e477f946061dda45891R235

rjarry:
@michalvasko would that be OK to export `ly_toprint()` in the public symbols for `libyang.so`?

Also, would that be OK to add one utility function:

```c
/**
 * @brief Return the data path with list key placeholders.
 *
 * @param[in] node Schema node for which to get the data path.
 * @param[in] placeholder String that will be inserted in place of the list key values.
 * @return NULL on error, on success the buffer for the resulting path is
*          allocated and caller is supposed to free it with free().
 */
char *lys_data_path_pattern(const struct lys_node *node, const char *placeholder);
```

Example of use for the Python bindings:

```python
p = lib.lys_data_path_pattern(self._snode, '%s')
p = ffi.string(p).decode('utf-8')
# /module1:container/module2:list[key1='%s']/sublist[key2='%s']/leaf
p %= ('val1', 'val2')
# /module1:container/module2:list[key1='val1']/sublist[key2='val2']/leaf
```

packages/CMakeLists.txt
rkrejci: I think that since the API is probably different, changing the name makes sense to me.

rjarry:
@michalvasko I have made some additional commits to address what I mentioned about new functions in the C library. Please tell me if you prefer that I push them into a separate pull request.

Also, I'd like to support `ENABLE_STATIC` in the python bindings but there is an link error. I am not sure how to fix it:

```
x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I/tmp/build/python/include -I/usr/include/python3.5m -c /tmp/build/python/temp.linux-x86_64-3.5/_libyang.c -o /tmp/build/python/temp.linux-x86_64-3.5/tmp/build/python/temp.linux-x86_64-3.5/_libyang.o -Werror -std=c99
x86_64-linux-gnu-gcc -pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-Bsymbolic-functions -Wl,-z,relro -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 /tmp/build/python/temp.linux-x86_64-3.5/tmp/build/python/temp.linux-x86_64-3.5/_libyang.o -L/tmp/build -lyang -o /tmp/build/python/lib.linux-x86_64-3.5/_libyang.cpython-35m-x86_64-linux-gnu.so
/usr/bin/ld: /tmp/build/libyang.a(common.c.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
/tmp/build/libyang.a: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
```

michalvasko:
Hi,
sorry it took me so long to get to this, I do not know what to do first. But at least you nicely refactored it since I last saw it so I have added the 2 functions into *libyang*, so please pull it and adjust your code accordingly. As for the static build, I guess it is not really a priority but I will try to look at that. No idea when that will be but it should not be blocking your work, I hope.

Regards,
Michal

rjarry:
Thanks Michal, I rebased on the latest `devel` and changed the python code accordingly.

rkrejci:
Hi, would it be possible to keep also the SWIG bindings as an option? There are some ongoing projects already using this API (even our own libnetconf2) and I don't want to break them. I think that a good thing for this is your change of the Python package name, so they can be distinguished. They can be separated options in CMake and probably even both can be built if required.

michalvasko:
@rkrejci I would find that really confusing. What about adding an option that will allow changing to the old swig bindings, by default the CFFI ones would be compiled and installed?

rkrejci:
@michalvasko That's probably what I'm proposing - having an extra option for SWIG as well as for CFFI bindings, the default one can be the CFFI (as it is now set in the PR) and if required the SWIG Python bindings can be enabled the same way as e.g. the java or javascript bindings.

But currently, the SWIG Python bindings are being removed completely.

michalvasko:
Right, not allowing to build both bindings was my main point.

rkrejci:
@michalvasko well, by default. But I don't see any reason why to disable such a setup when I want to build both bindings to support old and new projects depending on different libyang bindings. If I understand it correctly, the old bindings are built as Python yang package while the CFFI bindings are Python libyang package, so they are different even by name. Is it right @rjarry ?

michalvasko:
Like I said, I would simply find it confusing. You could then even use both bindings in a single project, for example. But, as I will likely not be using them or maintaining them, it is not my decision to make.

rjarry:
Hi @rkrejci,

Sure I can leave the SWIG python bindings. However, leaving both bindings enabled by default may be confusing. Building both python bindings should be possible but not the default in my opinion.

The question is: which binding is enabled by default? And with which cmake option?

I am on vacation today, I'll have a closer look tomorrow.

rkrejci:
@rjarry sure, only one of them will be the default, but I'm not sure which one it should be. I probably slightly tend to have SWIG bindings as the default one. They are there for some time already and if someone is already using python bindings, he will expect the SWIG version. On the other hand, in README or somewhere in doc I would strongly recommend trying the CFFI version since it is expected to be more Pythonic (so it should provide better user experience).

rjarry:
@rkrejci I have pushed a new version with SWIG bindings left as they are. The default is to build only the legacy bindings. The new CFFI bindings can be enabled (independently from any other SWIG binding) with `GEN_PYTHON_CFFI_BINDINGS`.

Following this, I have not modified anything in `packages` since I don't know where you want to go with this.

michalvasko:
I suppose there can then be 2 separate Python binding packages. If you can make the changes to the scripts then go ahead, would save me some time. But it should not be too difficult.

rjarry:
Hi Michal, sure I'll have a look.

rjarry:
Hi Michal,

I have pushed some modifications in the packaging scripts. I only tested the `build-deb` target on Debian 10 (buster). I did not test the `build-rpm` target as I don't know what distributions you want to target (and I'm not sure which ones I have at hand).

Also, I have modified `.travis.yml` to handle automatic uploads on https://pypi.org when new commits are made on the `master` branch.

rjarry:
Hi @rkrejci @michalvasko,

I have removed the `__token__` username argument for twine as you requested.

Did you have any more comments on the changes?

michalvasko:
Hi,
as for me, it seems good, have you actually build the package using these scripts?

Other than that, I suppose you could have come up with better names for the test modules but you have your name there :) Also, maybe some simple examples would be nice.

Regards,
Michal

rjarry:
Hi Michal,

I have built the deb packages only. Not the rpms.

About the test module names, I can change them but I am not sure what names would be more appropriate.

I have put minimalist examples in the README file. I can add other examples if you want but I figured that the unit tests are good examples.

rjarry:
Haha you were talking about YANG module names! I thought you were talking about python files.

Well a bit of fantasy cannot hurt from time to time :)

michalvasko:
Hi,
okay, as far as I am concerned, it can be merged.

Regards,
Michal

rjarry:
Glad to hear that, let me know if there is something to do on my end.

rjarry:
Hi Michal,

I have fixed a minor problem. Seen by code review. Also, I rebased the commits on the latest devel.

When do you think this can be merged?

rkrejci:
Hi Robin, sorry, I'm still blocking it wanting to look at that.

I tried to compile and import the module, but I'm getting the following error. Am I doing something wrong? The build process seemed fine to me.
```
>>> import libyang
ImportError: dynamic module does not define module export function (PyInit_libyang)
```

rjarry:
Hi Radek,

it looks like you have the C libyang.so library file somewhere into `$PYTHONPATH` (or in the current directory). Maybe you are executing python3 from inside the build dir?

rjarry:
If you want to try the bindings without installing them you need to export the proper env vars. As it is done for the test commands:

https://github.com/6WIND/libyang/blob/6e6475d8d5ad175c78d523b73a0304ddc59e116b/python/CMakeLists.txt#L77-L80

rkrejci:
ok, I fine with it and going to merge it

rkrejci:
Thanks for your work, Robin!

rjarry:
Thanks Radek! And thank you Michal for the review!

rjarry:
@rkrejci @michalvasko it looks like the deployment does not work: https://travis-ci.org/github/CESNET/libyang/jobs/693729586#L3715-L3832

There are multiple errors:

https://travis-ci.org/github/CESNET/libyang/jobs/693729586#L3737-L3738

https://travis-ci.org/github/CESNET/libyang/jobs/693729586#L3812-L3817

https://travis-ci.org/github/CESNET/libyang/jobs/693729586#L3829

I'm sorry about that, I don't know how to test this and fix it properly. Can you help?

rjarry:
The twine error seems related to this: https://github.com/pypa/twine/issues/342

rjarry:
The codecov error seems related to this: https://github.com/codecov/codecov-python/issues/261

michalvasko:
Hi,
yes, everything broke for various reasons :( You think you can fix at least the `twine` problem? `codecov` is not that important right now.

Regards,
Michal

rjarry:
Yes I'll submit another PR with the fix.

PR #1088, opened 2020-05-20
Change tests splittage

rkrejci:
Reorganize unit test to better reflect functionality blocks of libyang and join the related tests together. This is just an initial part of a long term task to migrate all the current tests. The PR changes directory structure reflected in the build system and initially joins the schema's identity and feature statement tests.


tests/utests/schema/test_parser_yang.c
```diff
@@ -20,14 +20,12 @@
 #include <stdio.h>
 #include <string.h>
 
-#include "../../src/common.h"
-#include "../../src/tree_schema.h"
-#include "../../src/tree_schema_internal.h"
+#include "../../../src/common.h"
+#include "../../../src/tree_schema.h"
+#include "../../../src/tree_schema_internal.h"
```
michalvasko, line 25: It may be worth it to generate one config header for all the tests (there may already be one, I have not checked) and easily put these (absolute) paths there with any other needed paths or defines. It can then be included in the tests without worrying in what directory they are or even move them around without breaking anything.

But there still would be a problem of setting the path to that one config header, now that I think about it, so it may be easier to simply change these paths to `"common.h"` and so on and use `cmake` to add include directory `src` only for the tests. Should achieve the same thing.

rkrejci: maybe, I didn't want to touch this includes right now because many of them will not be necessary since we are moving from internal functions to calling API functions in tests. I think that, if it will be possible, this part can be redesigned as soon as the test functions are moved into the new joined form. I expect that some simplification then.

michalvasko: Right, makes sense.

tests/utests/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+    " "
+    " "
+    " "
+    " "
+    " "
+    " "
+    " "
+    " ")
+set(tests ${tests} ${local_tests} PARENT_SCOPE)
+set(tests_wraps ${tests_wraps} ${local_tests_wraps} PARENT_SCOPE)
```
michalvasko, line 40: We could change this so that all the tests do not have to be explicitly mentioned, just use a `*.c` regex instead, it makes adding/removing tests transparent. However, the one test with a linker wrap will need to be handled specially and excluded from the regex. I still think it is worth it but it is up to a discussion.

rkrejci: It will require complete redesign of the build process. I'm not sure it is worth it. 

My idea here is that the lines do not correspond to the test source files, but to the test processes. For example `utest:schema/test_schema` already includes tests from `tests/utests/schema/test_schema_common.c` and `tests/utests/schema/test_schema_stmts.c`. This way we can split source files to keep them reasonable big/small while keeping relatively small number of the test processes.

michalvasko: Well, it is not possible if several sources will be compiled into a single executable test. If that is the definite plan, then I guess it must stay this way.

PR #1091, opened 2020-05-25
Fix compile on older systems

mwinter-osr:
fixes compiling libyang on older Systems such as Debian 8 and CentOS 6

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE debug)
 endif()
 
-set(CMAKE_C_FLAGS         "${CMAKE_C_FLAGS} -Wall -Wextra -std=gnu11")
+set(CMAKE_C_FLAGS         "${CMAKE_C_FLAGS} -Wall -Wextra -std=gnu99")
 set(CMAKE_C_FLAGS_RELEASE "-O2 -DNDEBUG")
 set(CMAKE_C_FLAGS_PACKAGE "-g -O2 -DNDEBUG")
 set(CMAKE_C_FLAGS_DEBUG   "-g -O0")
--- a/src/tree_schema.c
+++ b/src/tree_schema.c
@@ -5235,8 +5235,9 @@ lys_data_path_pattern(const struct lys_node *node, const char *placeholder)
         if (node->nodetype == LYS_LIST) {
             /* add specific key values (placeholders) for list */
             const struct lys_node_list *list;
+            uint8_t j;
             list = (const struct lys_node_list *)node;
-            for (uint8_t j = 0; j < list->keys_size; j++) {
+            for (j = 0; j < list->keys_size; j++) {
                 k += sprintf(keys + k, "[%s=%s]", list->keys[j]->name, placeholder);
             }
         }


mwinter-osr:
The Build bugfix for "CMAKE_C_STANDARD" breaks older Linux. I don't see where this was needed. Might be nice for newer systems, but breaking compatibility on older systems is probably not enough justification.
The 2nd fix is just a simple fix to keep it C99 compatible (which is an issue for some old Linux as well)

michalvasko:
Hi,
what exactly is the problem with this change? I did it for the very opposite reasons, to set the standard on systems with `cmake` older than 3.1 (ScientificLinux 7.3) because on those the variable was not recognized and ignored. That is also the reason you had to change the `for` loop, variable declaration in it is allowed only in C99 standard and newer and on those old system compilers the default standard usually is C89 (which is used because no other was explicitly set).

Regards,
Michal

CMakeLists.txt
```diff
@@ -47,7 +47,8 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE debug)
 endif()
 
-set(CMAKE_C_FLAGS         "${CMAKE_C_FLAGS} -Wall -Wextra -std=gnu11")
+set(CMAKE_C_STANDARD 11)
```
oerdnj, line 50: I am confused here. If you are retaining C99 compatibility, this should be `set(CMAKE_C_STANDARD 99)`.

src/tree_schema.c
```diff
@@ -5235,8 +5235,9 @@ lys_data_path_pattern(const struct lys_node *node, const char *placeholder)
         if (node->nodetype == LYS_LIST) {
             /* add specific key values (placeholders) for list */
             const struct lys_node_list *list;
+            uint8_t j;
             list = (const struct lys_node_list *)node;
-            for (uint8_t j = 0; j < list->keys_size; j++) {
+            for (j = 0; j < list->keys_size; j++) {
```
oerdnj, line 5240: The old syntax is definitely supported on older systems.  It's C99, so unless you want to retain C89 (why-oh-why?) compatibility, this is fine.

mwinter-osr:
> Hi,
> what exactly is the problem with this change? I did it for the very opposite reasons, to set the standard on systems with `cmake` older than 3.1 (ScientificLinux 7.3) because on those the variable was not recognized and ignored. That is also the reason you had to change the `for` loop, variable declaration in it is allowed only in C99 standard and newer and on those old system compilers the default standard usually is C89 (which is used because no other was explicitly set).
> 
> Regards,
> Michal

I'm a cmake novice, so might have been something else. I just know that the revert fixed the cmake error for me.
Here is how the compile error looked before I reverted back the change\. Output from NetBSD 6, but same at least on CentOS 6:
```cd cibuild.1/libyang-source/build; cmake -D ENABLE_LYD_PRIV=ON -D CMAKE_BUILD_TYPE:String=Release ..
-- The C compiler identification is GNU 4.5.3
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found FLEX: /usr/bin/flex (found version "2.5.35") 
-- Found BISON: /usr/pkg/bin/bison (found version "3.0.4") 
-- Looking for vdprintf
-- Looking for vdprintf - not found
-- Looking for asprintf
-- Looking for asprintf - not found
-- Looking for vasprintf
-- Looking for vasprintf - not found
-- Looking for get_current_dir_name
-- Looking for get_current_dir_name - not found
-- Looking for strndup
-- Looking for strndup - not found
-- Looking for getline
-- Looking for getline - not found
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for sys/types.h
-- Looking for sys/types.h - not found
-- Looking for stdint.h
-- Looking for stdint.h - not found
-- Looking for stddef.h
-- Looking for stddef.h - not found
-- Check size of unsigned short
-- Check size of unsigned short - failed
-- Check size of unsigned int
-- Check size of unsigned int - failed
-- Check size of unsigned long
-- Check size of unsigned long - failed
CMake Error at /usr/pkg/share/cmake-3.9/Modules/TestBigEndian.cmake:49 (message):
  no suitable type found
Call Stack (most recent call first):
  CMakeModules/UseCompat.cmake:19 (TEST_BIG_ENDIAN)
  CMakeLists.txt:195 (use_compat)


-- Configuring incomplete, errors occurred!
See also "/home/ci/cibuild.1/libyang-source/build/CMakeFiles/CMakeOutput.log".
See also "/home/ci/cibuild.1/libyang-source/build/CMakeFiles/CMakeError.log".
```

If you have any other suggestion on how to fix this, then please let me know and I'll test it.

michalvasko:
Hi Martin,
it seems this is a `cmake` [bug](https://github.com/emscripten-core/emscripten/issues/400). I am not sure how exactly to fix it on these systems but what should work is to define `HAVE_IS_BIG_ENDIAN` and set `IS_BIG_ENDIAN` correctly for this check to be skipped. So that is what you can try in the scripts for the problematic distros but then the packages would be architecture-dependent.

Regards,
Michal

mwinter-osr:
> Hi Martin,
> it seems this is a `cmake` [bug](https://github.com/emscripten-core/emscripten/issues/400). I am not sure how exactly to fix it on these systems but what should work is to define `HAVE_IS_BIG_ENDIAN` and set `IS_BIG_ENDIAN` correctly for this check to be skipped. So that is what you can try in the scripts for the problematic distros but then the packages would be architecture-dependent.
> 
> Also, in the referenced issue they have replaced the used `cmake` scripts with their own fixed ones, so that is another option, I guess.
> 
> Regards,
> Michal

Michal,

not a cmake bug. It's just that old systems don't understand/support gnu-c11. From the `CMakeError.log`:
```
/usr/bin/cc  -D_GNU_SOURCE -D_POSIX_C_SOURCE=200809L -Wall -Wextra -std=gnu11   
 -o CMakeFiles/cmTryCompileExec1418531595.dir/CheckSymbolExists.c.o   -c /home/c
i/libyang/build/CMakeFiles/CMakeTmp/CheckSymbolExists.c
cc1: error: unrecognized command line option "-std=gnu11"
gmake[1]: Leaving directory `/home/ci/libyang/build/CMakeFiles/CMakeTmp'
gmake[1]: *** [CMakeFiles/cmTryCompileExec1418531595.dir/CheckSymbolExists.c.o] 
Error 1
gmake: *** [cmTryCompileExec1418531595/fast] Error 2
```
this causes all the `Looking for` and endian checks to fail.
std=gnu9
Is there really a need for c11? If we change it to `-std=gnu99` then it would work.
I'll update the PR to use the rewritten version, but with -std=gnu99.

mwinter-osr:
Changed to use gnu99 instead of gnu11 standard. This fixes it for me. The 2nd commit in tree_schema.c is now no longer needed as gnu99 supports this. But as it's different syntax at only one location, I'm not sure if you want keep it the same. Let me know if you want me to update the PR and not touch the tree_schema.c

So can we keep it as C99 ?

michalvasko:
Hi,
okay, I have not seen the output and was hoping the compilers are not that old. Anyway, the C11 standard was required because of the use of `_Generic` but on the common compilers it is not used anyway and a compiler extension is used instead. Also, the one loop comes from a PR, that is why it differs. So, thanks for looking into it and I think it can be merged this way to keep things moving.

Regards,
Michal

PR #1092, opened 2020-05-26
Fixing RPM for CentOS 6/7/8 and Fedora

mwinter-osr:
Fixes RPM Package build

### DIFF
--- a/packages/libyang.spec.in
+++ b/packages/libyang.spec.in
@@ -127,8 +127,8 @@ make DESTDIR=%{buildroot} install
 %{_datadir}/man/man1/yanglint.1.gz
 %{_datadir}/man/man1/yangre.1.gz
 %{_libdir}/libyang.so.*
-%{_libdir}/libyang/*
-%dir %{_libdir}/libyang/
+%{_libdir}/libyang1/*
+%dir %{_libdir}/libyang1/
 
 %files devel
 %defattr(-,root,root)
--- a/packages/libyang.spec.in
+++ b/packages/libyang.spec.in
@@ -26,7 +26,7 @@ BuildRequires:  libcmocka-devel
 %if %{with_lang_bind}
 
 BuildRequires:  gcc-c++
-%if 0%{?centos_version}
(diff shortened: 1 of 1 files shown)


mwinter-osr:
Trying a new attempt to fix the RPMs. I hope the OpenSuse Build System causes no issues, but a fix in this or similar way is needed to build libyang on a RPM based system
Tested (together with PR #1091 ) on CentOS & RedHat 6, 7, 8 and Fedora 29 and 31

If this still causes issues with OBS, then please respond with actual errors.

michalvasko:
Hi Martin,
the same problem occured, C++ and Python bindings were not build on Fedoras (27, 28, 30, 31, rawhide), ScientificLinux 7, and OpenSUSE Leap 15.1 and Tumbleweed. You can take a look at the [logs](https://build.opensuse.org/package/show/home:liberouter/libyang) but you will not really see much there except only the basic and devel packages were build while on CentOS 7, for example, you will also see the C++ and Python ones.

Regards,
Michal

mwinter-osr:
> Hi Martin,
> the same problem occured, C++ and Python bindings were not build on Fedoras (27, 28, 30, 31, rawhide), ScientificLinux 7, and OpenSUSE Leap 15.1 and Tumbleweed. You can take a look at the [logs](https://build.opensuse.org/package/show/home:liberouter/libyang) but you will not really see much there except only the basic and devel packages were build while on CentOS 7, for example, you will also see the C++ and Python ones.
> 
> Regards,
> Michal

Michal,
ok, thanks for the input. So I assume the build works, it just doesn't build all the necessary packages. I'll get back to you with an update on this.

mwinter-osr:
Updated the PR. I do believe that this should solve the issue seen with the missing bindings.

michalvasko:
Hi Martin,
yes, now it seems to be working as it should. So, are you fine now with the current `devel` (with this PR) and you can build all the packages successfully? Can I make a release out of it?

Regards,
Michal

PR #1096, opened 2020-05-28
build BUGFIX fix detection of vasprintf() on FreeBSD >= 11

rwestphal:
When vasprintf() isn't detected, the fallback implementation
from compat/compat.c doesn't compile on FreeBSD (>= 11) because
of a strange problem with va_copy() not being available. Define
__BSD_VISIBLE to fix the detection of vasprintf(), which in turn
fixes the build errors on FreeBSD >= 11.


michalvasko:
Hi Renato,
okay, no problem, thanks.

Regards,
Michal

PR #1101, opened 2020-06-02
python BUGFIX only include required files in source archive

rjarry:
When building into a subdirectory of the root of the repository, a lot
of build artefacts are included into the source archive (intended for
upload on pypi.org). This makes the archive bigger and these files are
not required.

Be more restrictive about files that are included into the source
archive.

Fixes: 6e6475d8d5ad ("bindings CHANGE add new python CFFI bindings")
Signed-off-by: Robin Jarry <robin.jarry@6wind.com>


rjarry:
@michalvasko if that is ok with you, after you merge this, I'll upload the source archive for 1.0.176 again with the extraneous files removed.

michalvasko:
No problem, really. It affects only the bindings so I am okay with what you are okay.

PR #1109, opened 2020-06-11
python: expose LYD_OPT_GET and lyd_merge()

rwestphal:
While I loved the new Python bindings, I missed those two things when writing a script that receives data fragments and combines them into a single data tree. This prompted me to open this PR.

I hope these additions are welcome, but bear in mind I'm not a Python programmer so please review carefully!


python/libyang/data.py
```diff
@@ -152,17 +154,29 @@ def path(self):
...
+        flags = 0
+        if destruct:
+            flags |= lib.LYD_OPT_DESTRUCT
+        if no_siblings:
+            flags |= lib.LYD_OPT_NOSIBLINGS
+        if explicit:
+            flags |= lib.LYD_OPT_EXPLICIT
+        ret = lib.lyd_merge(self._node, source._node, flags)
+        if ret != 0:
+            self.context.error('merge failed')
```
rjarry, line 178: This should be:

```python
raise self.context.error('merge failed')
```

I just noticed that`raise` is mising in `DNode.validate()` as well. Too bad, you took the only wrong example :)

python/libyang/data.py
```diff
@@ -152,17 +154,29 @@ def path(self):
...
             self.context.error('validation failed')
 
+    def merge(self, source, destruct=False, no_siblings=False, explicit=False):
+        flags = 0
+        if destruct:
+            flags |= lib.LYD_OPT_DESTRUCT
+        if no_siblings:
+            flags |= lib.LYD_OPT_NOSIBLINGS
+        if explicit:
+            flags |= lib.LYD_OPT_EXPLICIT
```
rjarry, line 175: Can these flags be used for something else than `lyd_merge()`? If so, maybe you should wrap this in a function like `parser_flags()`. If not, it can stay like this.

rjarry: In fact, I looked at the C header, and these flags are into the "parser options". Can you handle them directly into `parse_flags()` along with the others? Of course, the boolean arguments can stay the same.

python/libyang/data.py
rwestphal: Fixed!

python/libyang/data.py
rwestphal: Good suggestion. Done!

rkrejci:
Thanks guys!

PR #1112, opened 2020-06-13
Redesign unit tests creation

rkrejci:
Some changes in tests to make its creation process more transparent and common.


michalvasko:
I like it much more now, well done.

Issue #1114, opened 2020-06-19
The validation of the filter conditions and returned results of query operation are too strict


jiayinhaoran:
We recently noticed that paring data has become more rigorous than ever before,  from this commit :https://github.com/CESNET/libyang/commit/0abb1bc3f95876f82625e8a0aadd973ee4a58db5.

But in our actual scenario, query operation(such as get, get-config, get-data...) usually need a more relaxed validation requirement. We need more relaxed value validation for the filter condition and returned result, because this will improve the user experience.
Please correct me if I am missing something.

We take the returned result of query operation as an example to describe the requirements. For example, for a full query operation of running db:
```
<rpc xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="1">
  <get-config xmlns="urn:ietf:params:xml:ns:netconf:base:1.0">
    <source>
      <running/>
    </source>
  </get-config>
</rpc>
```
If there are many apps' config data stored in here, but sometimes one or more apps stored some data inconsistent with model definition in db by some ways(such as model and data upgrade not synchronized...), the current libyang will have errors in the validation phase when parsing these data. 

However, this will cause the full query operation to be unavailable until the app solves these data inconsistencies. This will greatly affect the user experience during this period and if there are enough apps, this will maybe happen from time to time. Filter conditions is similar to this.

So, we want libyang to provide a mechanism to reduce the verification of filter conditions and returned results during query operations.

In the actual scenario, we currently use a temporary patched solution:
https://github.com/CESNET/libyang/pull/1115/files

Sincerely hope to get some more effective suggestions, thanks.

PR #1115, opened 2020-06-20
weaken the validation of trusted query operation

jiayinhaoran:
Target Parsing option combination[lyd_parse_mem, lyd_parse_xml...]: LYD_OPT_GET|LYD_OPT_TRUSTED or LYD_OPT_GETCONFIG|LYD_OPT_TRUSTED

Related issue:
https://github.com/CESNET/libyang/issues/1114

### DIFF
--- a/src/parser_xml.c
+++ b/src/parser_xml.c
@@ -75,13 +75,18 @@ xml_data_search_schemanode(struct lyxml_elem *xml, struct lys_node *start, int o
 
 /* logs directly */
 static int
-xml_get_value(struct lyd_node *node, struct lyxml_elem *xml, int editbits)
+xml_get_value(struct lyd_node *node, struct lyxml_elem *xml, int editbits, int options)
 {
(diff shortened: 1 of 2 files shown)


michalvasko:
Hi,
I think your use-case can be supported but this patch is not a good solution. So, please rewrite it so that the values are always attempted to be loaded normally and only if that fails and you have the mentioned option combination, it is stored as `LY_TYPE_UNKNOWN`. But then the printer cannot simply print the value as if it was a valid one of the specific type, that can easily lead to various problems. Instead, simply its `value_str` should be printed directly.

Regards,
Michal

jiayinhaoran:
Hi,
Sorry to reply so late. First of all, thank you for your approval of the rationality of use cases. Secondly, you are right and this method is just a temporary method, in which many of the surrounding effects have not been carefully considered. Next time, I will give a more comprehensive solution to this problem for further consideration.

PR #1120, opened 2020-06-23
Python fixes

rjarry:
The conversion from python `dict` to `struct lyd_node *` is very slow and does not scale linearly with the size of the dictionary. This commit series contains a major rework of `dict_to_dnode` which brings significant performance improvements.

Here are some ballpark numbers for parsing a dictionary with lists that contain nested containers, leafs, etc. The "num. objects" field is only to show that the previous algorithm did not scale linearly with the size of the dictionary and that this has been fixed with the optimization:

| num. objects           |    500  |    1000   |  2000  |   5000   | 10000 |
|  :---                |  ---: |  ---: |  ---: |  ---: |  ---: |
| before optimization |     7.67s |   31.18s |  117.74s  | 758.51s    |   -- |
| after optimization  |      .07s  |    .13s   |  .24s   |  .57s  |   1.16s |

I did not manage to test with 10000 objects without the optimization, it is completely unusable anyway. I had to interrupt the test after 20 minutes of processing before it was finished.

The `SNode.parse_data_dict` method was unfortunately **removed** from the API. This method may not work in all cases. For example, if the schema node object is nested into a list, information about the actual list element instance in which the data node should be created is missing. A "replacement" method `DNode.merge_data_dict` was added to allow subtree parsing.

Other fixes/enhancements are related to python `dict` parsing as well.


rkrejci:
Thanks Robin, interesting improvements!

rjarry:
My pleasure, thanks Radek :)

PR #1121, opened 2020-06-23
data tree FEATURE functions for creating and applying diff

michalvasko:


src/tree_data.c
```diff
@@ -2336,6 +2339,987 @@ lyd_merge(struct lyd_node **target, const struct lyd_node *source, int options)
...
+/**
+ * @brief Learn operation of a diff node.
+ *
+ * @param[in] diff_node Diff node.
+ * @param[out] op Operation.
+ * @param[out] key_or_value Optional list instance keys predicate or leaf-list value for move operation.
+ * @return LY_ERR value.
+ */
+static LY_ERR
+lyd_diff_get_op(const struct lyd_node *diff_node, const char **op, const char **key_or_value)
```
rkrejci, line 3029: I would prefer to have the `op` as an enum - checking it in ifs against some explicit enum value with a clear meaning would be much better than deciding the following steps based on the first character of the operation name. In case of adding some new operation sometime in the future, this approach based on the different leading characters can simply invalidate.

tests/utests/data/test_diff.c
```diff
@@ -0,0 +1,671 @@
+/**
```
rkrejci, line 1: There is no test for failure? The diff is really supposed to not fail for some kind of input?

michalvasko: Added a test returning inval arg.

src/tree_data.c
michalvasko: Changed.

PR #1122, opened 2020-06-23
First PORT to Win32 (rebased to devel)

mekleo:
I just ported libyang to Win32. This shall make it and its companion tools such as yanglint available to a larger audience.
Moreover, a critical segmentttion fault in Bison parser's YANG_ADDELEM was found and fixed:
When the element counter depends on the "actual" pointer, it must not be incremented after "actual" has been changed.

This second pull request addresses the comments on the first one. Thank you for these fruitful reviews.


CMakeLists.txt
```diff
@@ -126,13 +132,25 @@ else()
...
+        )
+		add_definitions(/DYY_NO_UNISTD_H)
+	else()
+	    add_custom_target(bison
+            WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/src
+            COMMAND ${BISON_EXECUTABLE} -l -o parser_yang_bis.c --defines=parser_yang_bis.h ${PROJECT_BINARY_DIR}/src/yang.y
+            COMMAND ${FLEX_EXECUTABLE} -8 -L -o parser_yang_lex.c --header-file=parser_yang_lex.h yang.l
+            COMMAND ${SED_TOOL} -i 's/int yychar\;/int yychar\;\\nchar *s = NULL, *tmp_s = NULL, *ext_name = NULL\;\\nstruct lys_module *trg = NULL\;\\nstruct lys_node *tpdf_parent = NULL, *data_node = NULL\;\\nstruct lys_ext_instance_complex *ext_instance = NULL\;\\nint is_ext_instance\;\\nvoid *actual = NULL\;\\nenum yytokentype backup_type, actual_type = MODULE_KEYWORD\;\\nint64_t cnt_val = 0\;\\nint is_value = 0\;\\nvoid *yang_type = NULL\;/' parser_yang_bis.c
+        )
+	endif()
```
michalvasko, line 153: I can see only 2 differences between WIN and LINUX version, the new definition and using `"` instead of `'`. I believe the quote change is not really worth this whole change and I am quite sure it can be unified. Could you please try to do that?

mekleo: Sure. I now use a temporary variable to store quoted sed statements.

michalvasko:
Hi,
you did some improvements and it looks better but it still does not solve the main problem. Which is the fact that you had to add a lot of `cmake` code/modules and also several whole small libraries specific to Windows. That is just too much clutter to have in one repository, one can easy lose track of the purpose of each of the components.

So, my suggestion would be for you to keep this code in your repository and we would simply add a link to it in our README. You could then remove all the Unix-only scripts/code and have a nice clean Windows port. What do you think?

Regards,
Michal

mekleo:
Hi Michal,
I understand your take on this.

However, it is way easier for future maintenance to keep a unique codebase for Windows and Unix.
I can keep it in my repositiory, but it will always be CESNET's *nix libyang + Windows support, to simplify rebasing.

Thank you for the reference in the README. That might greatly help those in need of a Windows port.

Regards,
Leonce

michalvasko:
Hi,
right, so we should reference the project `https://github.com/mekleo/libyang/tree/Branch_v1.0.176_Win32`?

Regards,
Michal

mekleo:
Hi Michal,
Yes:  https://github.com/mekleo/libyang/tree/Branch_v1.0.176_Win32
Regards,
Leonce

PR #1127, opened 2020-06-29
LYB printer & parser

michalvasko:


src/parser_lyb.c
```diff
@@ -0,0 +1,1211 @@
...
+
+    ((char *)*str)[len] = '\0';
+    return LY_SUCCESS;
+}
+
+static void
+lyb_read_stop_subtree(struct lyd_lyb_ctx *lybctx)
+{
+    if (LYB_LAST_SUBTREE(lybctx).written) {
+        LOGINT(lybctx->ctx);
```
rkrejci, line 173: is it fine to continue?

src/tree_data_helpers.c
```diff
@@ -158,3 +159,98 @@ lyd_parse_check_keys(struct lyd_node *node)
...
+            lyd_free_meta(LYD_NODE_CTX(node), meta2, 0);
+            break;
+        }
+
+        prev_meta = meta2;
+    }
+}
+
+LYB_HASH
+lyb_hash(struct lysc_node *sibling, uint8_t collision_id)
```
rkrejci, line 207: just consider a standalone header (source) for the common LYB functions/types/macros/... for LYB format (used in printer, parsers,...)

src/tree_data_internal.h
```diff
@@ -317,6 +428,57 @@ LY_ERR lyd_parse_xml_notif(const struct ly_ctx *ctx, const char *data, struct ly
...
 LY_ERR lyd_parse_xml_reply(const struct lyd_node *request, const char *data, struct lyd_node **tree, struct lyd_node **op);
 
+/**
+ * @brief Parse binary data as YANG data tree.
+ *
+ * @param[in] ctx libyang context
+ * @param[in] data Pointer to the input data to parse.
+ * @param[in] options @ref dataparseroptions
+ * @param[out] tree Parsed data tree. Note that NULL can be a valid result.
+ * @param[out] parsed_p Optional number of parsed bytes.
```
rkrejci, line 438: the `_p` suffix seems confusing to me, what about `parsed_bytes` or just `bytes`? The same applies to other functions with this parameter.

src/parser_lyb.c
michalvasko: Not really, fixed.

src/tree_data_helpers.c
michalvasko: Separate header was added but not a source, it would be only for 2 functions.

src/tree_data_internal.h
michalvasko: Renamed.

PR #1134, opened 2020-07-01
Data parser api

rkrejci:
unify data parser's API with schema parser and use the input handler

### DIFF
--- a/src/parser_data.h
+++ b/src/parser_data.h
@@ -1,0 +1,290 @@
+/**
+ * @file parser_data.h
+ * @author Radek Krejci <rkrejci@cesnet.cz>
+ * @brief Data parsers for libyang
+ *
+ * Copyright (c) 2015-2020 CESNET, z.s.p.o.
+ *
+ * This source code is licensed under BSD 3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#ifndef LY_PARSER_DATA_H_
+#define LY_PARSER_DATA_H_
+
+#include "tree_data.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct ly_in;
+
+/**
+ * @addtogroup datatree
+ * @{
+ */
+
+/**
+ * @defgroup dataparseroptions Data parser options
+ *
+ * Various options to change the data tree parsers behavior.
+ *
+ * Default parser behavior:
+ * - complete input file is always parsed. In case of XML, even not well-formed XML document (multiple top-level
+ * elements) is parsed in its entirety,
+ * - parser silently ignores data without matching schema node definition,
+ * - list instances are checked whether they have all the keys, error is raised if not.
+ *
+ * Default parser validation behavior:
+ * - the provided data are expected to provide complete datastore content (both the configuration and state data)
+ * and performs data validation according to all YANG rules, specifics follow,
+ * - list instances are expected to have all the keys (it is not checked),
+ * - instantiated (status) obsolete data print a warning,
+ * - all types are fully resolved (leafref/instance-identifier targets, unions) and must be valid (lists have
+ * all the keys, leaf(-lists) correct values),
+ * - when statements on existing nodes are evaluated, if not satisfied, a validation error is raised,
+ * - if-feature statements are evaluated,
+ * - invalid multiple data instances/data from several cases cause a validation error,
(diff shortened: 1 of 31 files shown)


mwinter-osr:
CI:rerun

src/printer_lyb.c
```diff
@@ -747,8 +748,8 @@ lyb_print_metadata(struct ly_out *out, const struct lyd_node *node, struct lyd_l
 
     /* with-defaults */
     if (node->schema->nodetype & LYD_NODE_TERM) {
-        if (((node->flags & LYD_DEFAULT) && (lybctx->options & (LYDP_WD_ALL_TAG | LYDP_WD_IMPL_TAG))) ||
-                ((lybctx->options & LYDP_WD_ALL_TAG) && ly_is_default(node))) {
+        if (((node->flags & LYD_DEFAULT) && (lybctx->print_options & (LYDP_WD_ALL_TAG | LYDP_WD_IMPL_TAG))) ||
```
michalvasko, line 751: Seeing `LYDP_WD_ALL_TAG` here, it should probably be renamed to `LYD_PRINT_WD_ALL_TAG` together with other print options.

src/validation.c
```diff
@@ -828,9 +829,9 @@ lyd_validate_must(const struct lyd_node *node, int val_opts)
         break;
     case LYS_RPC:
     case LYS_ACTION:
-        if (val_opts & LYD_VALOPT_INPUT) {
+        if (val_opts & LYD_VALIDATE_OP_RPC) {
```
michalvasko, line 832: Enum should not be checked with `&` so it would be better to separate these into another parameter.

src/tree_data.c
```diff
@@ -280,117 +282,171 @@ lyd_meta2str(const struct lyd_meta *meta, int *dynamic)
...
+        return lyd_parse_json_data(ctx, in->current, parse_options, validate_options, tree);
 #endif
     case LYD_LYB:
-        lyd_parse_lyb_data(ctx, data, options, &result, NULL);
-        break;
+        return lyd_parse_lyb_data(ctx, in->current, parse_options, validate_options, tree, NULL);
     case LYD_SCHEMA:
-        LOGINT(ctx);
-        break;
+        LOGINT_RET(ctx);
```
michalvasko, line 333: There should not be a return yet, but final top-level validation should be performed here for all the formats. Meaning move the `if` block at `parser_xml.c:698` here and remove `TODO` in `parser_lyb.c:1060`.

src/printer_lyb.c
rkrejci: ok

src/tree_data.c
rkrejci: The joint top-level validation for all the data formats parsers will be done later together with the JSON format data parser.

src/validation.c
rkrejci: true, changed

src/parser_data.h
```diff
@@ -0,0 +1,290 @@
...
+ * @defgroup datavalidateop Operation to validate
+ * @ingroup datatree
+ *
+ * Operation provided to lyd_validate_op() to validate. The operation cannot be determined automatically since RPC/action and a reply to it
+ * share the common top level node referencing the RPC/action schema node.
+ */
+typedef enum {
+    LYD_VALIDATE_OP_RPC = 1,   /**< Validate RPC/action request (input parameters). */
+    LYD_VALIDATE_OP_REPLY,     /**< Validate RPC/action reply (output parameters). */
+    LYD_VALIDATE_OP_NOTIF      /**< Validate RPC/action reply (output parameters). */
```
michalvasko, line 119: Copy-paste problem.

src/parser_data.h
```diff
@@ -0,0 +1,290 @@
...
+//#define LYD_VALIDATE_DATA_TEMPLATE 0x0008   /**< Data represents YANG data template. */
+
+/** @} datavalidationoptions */
+
+/**
+ * @defgroup datavalidateop Operation to validate
+ * @ingroup datatree
+ *
+ * Operation provided to lyd_validate_op() to validate. The operation cannot be determined automatically since RPC/action and a reply to it
+ * share the common top level node referencing the RPC/action schema node.
```
michalvasko, line 114: I would add `... and may not have any input/output children to use for distinction.`

src/parser_data.h
rkrejci: fixed

src/parser_data.h
rkrejci: added

PR #1138, opened 2020-07-03
Add ABI laboratory tools

oerdnj:
This PR adds support for checking the ABI difference between master (committed as libyang.dump) and the working branch.  When the API change is intentional, the `libyang.dump` needs to be updated with `build/libyang.dump` (created by `tools/abi-check.sh`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,12 @@ branches:
     - master
     - devel
 
+addons:
+  apt:
+    update: true
+
 before_install:
+  - if [ "$DEPLOY_BUILD" = "TRUE" ]; then sudo apt-get -y install abi-dumper abi-compliance-checker pkg-config; fi
   - git clone git://git.cryptomilk.org/projects/cmocka.git
   - cd cmocka && mkdir build && cd build
   - cmake .. && make -j2 && sudo make install
@@ -54,6 +59,7 @@ script:
   - if [ "$TRAVIS_OS_NAME" = "linux" -a "$TRAVIS_ARCH" = "arm64" ]; then cmake -DGEN_LANGUAGE_BINDINGS=ON -DENABLE_VALGRIND_TESTS=OFF ..; fi
   - make -j2 && ctest --output-on-failure
   - cd -
+  - if [ "$DEPLOY_BUILD" = "TRUE" ]; then ./tools/abi-check.sh; fi
 
 deploy:
   - provider: script
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ addons:
     update: true
 
 before_install:
-  - if [ "$DEPLOY_BUILD" = "TRUE" ]; then sudo apt-get -y install abi-dumper abi-compliance-checker pkg-config; fi
+  - if [ "$DEPLOY_BUILD" = "TRUE" ]; then sudo apt-get -y install abi-dumper abi-compliance-checker pkg-config w3m; fi
   - git clone git://git.cryptomilk.org/projects/cmocka.git
   - cd cmocka && mkdir build && cd build
   - cmake .. && make -j2 && sudo make install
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -38,22 +46,45 @@ branches:
...
-  - cd ../..
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew update; fi
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update -qq; sudo apt-get install -y valgrind libpcre3-dev python3-dev swig; fi
+  - OLD_CMOCKA_VERSION=$(sed -n 's/project(cmocka VERSION \([0-9.]*\) LANGUAGES C)/\1/p' cmocka/CMakeLists.txt)
+  - if [ -f "cmocka/CMakeLists.txt" ] && [ "$OLD_CMOCKA_VERSION" != "$CMOCKA_VERSION" ]; then echo "Purging the cmocka-$OLD_CMOCKA_VERSION from cache"; rm -rf cmocka; fi
+  - if [ ! -d "cmocka" ] || [ ! -f "cmocka/CMakeLists.txt" ]; then echo "Downloading cmocka-$CMOCKA_VERSION"; rm -rf cmocka && curl -SLO https://cmocka.org/files/1.1/cmocka-$CMOCKA_VERSION.tar.xz && tar -xJf cmocka-$CMOCKA_VERSION.tar.xz && mv "cmocka-$CMOCKA_VERSION" cmocka; fi
+  - if [ ! -d "cmocka/build" ] || [ ! -f "cmocka/build/Makefile" ]; then mkdir -p cmocka/build && (cd cmocka/build && cmake ..); fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ] && [ ! -f "cmocka/build/src/libcmocka.dylib" ]; then (cd cmocka/build && make -j2); fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ] && [ ! -f "cmocka/build/src/libcmocka.so" ]; then (cd cmocka/build && make -j2); fi
```
michalvasko, line 76: We are compiling `cmocka` from source but that is obsolete and actually the packages should be available so can you simply remove all these commands and install it as a package? Or do you require some specific version?

tools/abi-check.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh -e
 status=0
 # Dump the current ABI
-abi-dumper ./build/libyang.so -o ./build/libyang.dump -lver "$(PKG_CONFIG_PATH=./build pkg-config --modversion libyang)"
+LC_ALL=C.UTF-8 PATH=/snap/bin:$PATH abi-dumper ./build/libyang.so -o ./build/libyang.dump -lver "$(PKG_CONFIG_PATH=./build pkg-config --modversion libyang)" -public-headers ./src -public-headers ./build/src
```
rkrejci, line 4: Why the snap is needed here? The ABI tools are installed via apt in Travis, so I guess they will be in the standard location.

rkrejci: ahh, the universal-ctags are installed via snap

oerdnj: Unfortunately, the `universal-ctags` package is only available from Ubuntu eoan and up. The `abi-dumper` produces slightly better output when it can parse the public headers, so I wanted to have the `ctags` available for it.  When Travis CI starts to provide Ubuntu focal, this can go away and the packaged version can be used.

.travis.yml
```diff
@@ -1,5 +1,10 @@
 language: c
 
+cache:
+  pip: true
+  directories:
+    - cmocka
```
michalvasko, line 6: This is redundant now since it is installed as a package, right?

oerdnj: Yes, this should have been dropped and I missed that.

PR #1141, opened 2020-07-08
Rename RPM package to libyang1 for coexisting with libyang 0.x

mwinter-osr:
Renamed packages similar to official debian package
Moved yanglint/yangre utilities to new yang-tools package
Changes allow libyang 0.x version and libyang 1.x lib package
to be installed at the same time.

### DIFF
--- a/packages/CMakeLists.txt
+++ b/packages/CMakeLists.txt
@@ -1,6 +1,7 @@
 set(PACKAGE "libyang")
 set(CPP_PACKAGE "libyang-cpp")
 set(PYTHON_PACKAGE "python3-yang")
+set(TOOLS_PACKAGE "yang-tools")
 
 find_program(DEB_BUILDER NAMES debuild)
 find_program(RPM_BUILDER NAMES rpmbuild)
--- a/packages/CMakeLists.txt
+++ b/packages/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(PACKAGE "libyang")
 set(CPP_PACKAGE "libyang-cpp")
 set(PYTHON_PACKAGE "python3-yang")
-set(TOOLS_PACKAGE "yang-tools")
+set(TOOLS_PACKAGE "libyang-tools")
 
 find_program(DEB_BUILDER NAMES debuild)
 find_program(RPM_BUILDER NAMES rpmbuild)
--- a/packages/libyang.spec.in
+++ b/packages/libyang.spec.in
@@ -43,7 +43,7 @@ BuildRequires:  python36-devel
 
 %endif
 
-%package devel
+%package -n @PACKAGE@-devel
 Summary:    Headers of libyang library
 Requires:   %{name} = %{version}-%{release}
 Requires:   pcre-devel
@@ -77,7 +77,7 @@ Headers of bindings to c++ language.
 Bindings of libyang library to python language.
 %endif
 
-%description devel
+%description -n @PACKAGE@-devel
 Headers of libyang library.
 
 %description -n @TOOLS_PACKAGE@
@@ -136,7 +136,7 @@ make DESTDIR=%{buildroot} install
 %{_libdir}/libyang1/*
 %dir %{_libdir}/libyang1/
 
-%files devel
+%files -n @PACKAGE@-devel
 %defattr(-,root,root)
 %{_libdir}/libyang.so
 %{_libdir}/pkgconfig/libyang.pc
--- a/packages/libyang.spec.in
+++ b/packages/libyang.spec.in
@@ -1,4 +1,4 @@
-name: @PACKAGE@1
+name: @PACKAGE@@LIBYANG_MAJOR_SOVERSION@
 Version: @LIBYANG_VERSION@
 Release: 0
 Summary: Libyang library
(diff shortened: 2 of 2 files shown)


packages/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 set(PACKAGE "libyang")
 set(CPP_PACKAGE "libyang-cpp")
 set(PYTHON_PACKAGE "python3-yang")
```
michalvasko, line 3: Please, modify these variables instead of adding `1` for all their uses, that is what they are here for.

packages/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
...
+set(TOOLS_PACKAGE "yang-tools")
```
michalvasko, line 4: This name seems a bit too generic, maybe `libyang-tools` would be better?

packages/CMakeLists.txt
mwinter-osr: I've considered this and tried it initially. But the challenge is that the rename should not happen for all packages.
Currently the packages (before the change) are:
* libyang-1.X.YYY-0.src.rpm
* libyang-1.X.YYY-0.x86_64.rpm
* libyang-cpp-1.X.YYY-0.x86_64.rpm
* libyang-cpp-debuginfo-1.X.YYY-0.x86_64.rpm
* libyang-cpp-devel-1.X.YYY-0.x86_64.rpm
* libyang-debuginfo-1.X.YYY-0.x86_64.rpm
* libyang-debugsource-1.X.YYY-0.x86_64.rpm
* libyang-devel-1.X.YYY-0.x86_64.rpm
* python3-yang-1.X.YYY-0.x86_64.rpm
* python3-yang-debuginfo-1.X.YYY-0.x86_64.rpm

In Debian (and similar to RedHat), I do believe that the devel packages should not be renamed, so the correct new naming should be:
* libyang1-1.X.YYY-0.src.rpm
* libyang1-1.X.YYY-0.x86_64.rpm
* libyang-cpp1-1.X.YYY-0.x86_64.rpm
* libyang-cpp1-debuginfo-1.X.YYY-0.x86_64.rpm
* libyang-cpp-devel-1.X.YYY-0.x86_64.rpm
* libyang1-debuginfo-1.X.YYY-0.x86_64.rpm
* libyang1-debugsource-1.X.YYY-0.x86_64.rpm
* libyang-devel-1.X.YYY-0.x86_64.rpm
* python3-yang1-1.X.YYY-0.x86_64.rpm
* python3-yang1-debuginfo-1.X.YYY-0.x86_64.rpm
* yang-tools-1.X.YYY-0.x86_64.rpm
* yang-tools-debuginfo-1.X.YYY-0.x86_64.rpm

So basically, the libyang-devel (and the other devel) packages as well as the `yang-tools` should not be renamed. This supports the install of the binary (runtime) of both versions, but only latest on development.

If I change the `packages/CMakeLists.txt`, then the dependencies will get a bit messy, eg in
```
%package -n @CPP_PACKAGE@-devel
Summary:    Headers of bindings to c++ language
Requires:   @CPP_PACKAGE@ = %{version}-%{release}
Requires:   pcre-devel
```
Here I would need to use a 2nd `CPP_PACKAGE_WITHVERSION` or something similar name to reflect that `libyang-cpp-devel` depends on `libyang-cpp1` (and not on `libyang-cpp`). Similar on the other devel packages. Could be done, but looked messy to me.

Additionally, if someone uses the CMakeLists file to rename the packages, this would cause a much greater complication to get it right.

So based on this, I decided to go for idea to add version number directly in the rpm spec file.

But happy to change this over if this is still you preference.

packages/CMakeLists.txt
mwinter-osr: I picked the same name as it uses on Debian Packages to keep things consistent

mwinter-osr:
While reading the comments, I've noticed an error in my submission for the `libyang-devel` package. My submission incorrectly names it `libyang1-devel`.

I'll fix this as soon as there is an answer (= decision) on the `packages/CMakeLists` comment above. Please wait with merging until this is fixed

mwinter-osr:
Already fixed the `libyang1-devel` wrong name for the `libyang-devel` package with the latest commit.

This is based on existing scheme (not adding the package version in the `packages/CMakeLists` file. Will redo if this is still required.

packages/CMakeLists.txt
michalvasko: Oh, I see. Well, there is a reason why I hate all the packaging stuff. Anyway, I would then ask you to just replace `1` with `@LIBYANG_MAJOR_SOVERSION@` so that we can reuse these scripts for the new *libyang*, unless you think there is some other problem, thanks.

packages/CMakeLists.txt
michalvasko: I see, would it be a serious problem to rename it and so the name would be different in the next Debian release?

packages/CMakeLists.txt
mwinter-osr: Good suggestion. Not sure why I missed that variable before.
Done in latest update

packages/CMakeLists.txt
mwinter-osr: Contacted Ondřej Surý (our Debian Expert) for his input on renaming debian (sub-)packages.

mwinter-osr: Updated in Debian by @oerdnj and updated in the PR for RPM packages

michalvasko: Great!

PR #1142, opened 2020-07-08
fix deviation replace config failure  when a leaf node  has not the property 'config' explicitly

fredgan:
Hi, Michal,
I find deviation replace the property 'config' which does not exist in a leaf node fails,the error msg is that :
`err : Invalid keyword "config".
err : Replacing a property that does not exist.
err : Module "mod1" parsing failed.`
It seems like the devition fails beacase leaf node  does not explicitly have  `config true` or `config false`, but according to  RFC 7950 (https://tools.ietf.org/html/rfc7950#section-7.21.1), `config true` is implied when it is not present, in this case, I think it is no need to check whether there is the propery `config`. 

yang modules are as follows:
1 )  mod.yang
```
module mod {
    namespace "urn:cesnet:mod";
    prefix m;

    leaf l {
        type string;
        //config true;
    }
}
```
2 ) mod1.yang
```
module mod1 {
    namespace "urn:cesnet:mod1";
    prefix m1;

    import mod {
        prefix m;
    }

    deviation "/m:l" {
        deviate replace {
            config true;
        }
    }
}
```
using yanglint, the result  is as follows:
```
root@oss-0007:/libyang/deviation# yanglint -f yang mod1.yang
err : Invalid keyword "config".
err : Replacing a property that does not exist.
err : Module "mod1" parsing failed.

```

### DIFF
--- a/src/parser_yang.c
+++ b/src/parser_yang.c
@@ -4491,10 +4491,6 @@ yang_check_deviate(struct lys_module *module, struct unres_schema *unres, struct
             LOGVAL(module->ctx, LYE_INSTMT, LY_VLOG_NONE, NULL, "config");
             LOGVAL(module->ctx, LYE_SPEC, LY_VLOG_NONE, NULL, "Adding property that already exists.");
             goto error;
-        } else if ((deviate->mod == LY_DEVIATE_RPL) && !(dev_target->flags & LYS_CONFIG_SET)) {
-            LOGVAL(module->ctx, LYE_INSTMT, LY_VLOG_NONE, NULL, "config");
-            LOGVAL(module->ctx, LYE_SPEC, LY_VLOG_NONE, NULL, "Replacing a property that does not exist.");
-            goto error;
         }
 
         /* add and replace are the same in this case */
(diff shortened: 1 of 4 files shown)


michalvasko:
Hi Fred,
you are right that the RFC is quite vague on this but in the [grammar](https://tools.ietf.org/html/rfc7950#page-201) you can see that `config` can be both added and replaced. That means that in your case you should use `add` instead of `replace` and replacing should really fail as it does.

Regards,
Michal

fredgan:
Hi Michal,
I also understood it like this before. However, you can see https://github.com/mbj4668/pyang/issues/267

and  also I email asked him, he confirmed what he said in 267 is right, 

what's more, I have a long discussion in https://github.com/mbj4668/pyang/issues/646 with another user.

So I think the RFC grammar may have some errors in deviate-add-stmt.

michalvasko:
Hi Fred,
I am sorry but if there is a problem, it should be fixed with an errata. As mentioned [here](https://www.rfc-editor.org/errata/eid5663), the grammar is actually the place to look for the exact statements that are allowed. So, unless `deviate add` for `config` is forbidden by an errata, *libyang* behavior will not change.

Regards,
Michal

PR #1152, opened 2020-07-20
Release 1.0.184

rkrejci:
Merge latest changes from the devel branch.


rkrejci:
seems that master - devel need some cleanup

PR #1159, opened 2020-08-05
FIX segmentation fault in yang.y.in's YANG_ADDELEM

mekleo:
This fixes a segmentation fault in Bison parser's macro YANG_ADDELEM.
When the element counter in "size" depends on the "actual" pointer, it must not be incremented after "actual" has been changed.

An occurrence of this segmentation fault is at line 1629: YANG_ADDELEM(((struct yang_type *)actual)->type->info.enums.enm, ((struct yang_type *)actual)->type->info.enums.count, "enums");
This clearly manifests in MSVC. When it does not on a given compiler, let's say GCC, I suspect a memory ordering artifact to cover this bug up.


michalvasko:
Hi,
okay, no problem.

Regards,
Michal

PR #1175, opened 2020-08-21
Uncrustify source formatting

michalvasko:


rkrejci:
@rjarry yes, that's the idea, but it still needs some tuning, this is just very basic setting not checking all the things we would like to check

rjarry:
Very nice, do you plan on doing the same thing for sysrepo?

michalvasko:
Most likely but we first want to have it set up properly for *libyang*.

PR #1177, opened 2020-08-26
libyang REFACTOR use size-specific int types instead of generic types

rkrejci:
Align parameters' types with the values in (internal) structures.

### DIFF
--- a/compat/compat.c
+++ b/compat/compat.c
@@ -128,7 +128,7 @@ getline(char **lineptr, size_t *n, FILE *stream)
 {
     static char line[256];
     char *ptr;
-    unsigned int len;
+    ssize_t len;
 
     if (!lineptr || !n) {
         errno = EINVAL;
--- a/src/tree_schema_internal.h
+++ b/src/tree_schema_internal.h
@@ -184,8 +184,8 @@ struct lysc_ctx {
     struct ly_set leafrefs;     /**< when/must to check */
     struct ly_set dflts;        /**< set of incomplete default values */
     struct ly_set tpdf_chain;
-    uint16_t path_len;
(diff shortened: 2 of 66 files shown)


compat/compat.c
```diff
@@ -128,7 +128,7 @@ getline(char **lineptr, size_t *n, FILE *stream)
 {
     static char line[256];
     char *ptr;
-    unsigned int len;
+    ssize_t len;
```
michalvasko, line 131: Why `ssize_t` and not `size_t`?

src/tree_schema_internal.h
```diff
@@ -781,12 +781,13 @@ char *lysc_path_until(const struct lysc_node *node, const struct lysc_node *pare
 const struct lysc_node *lysc_data_parent(const struct lysc_node *schema);
 
 /**
- * @brief Learn whether a node is in an operation output.
+ * @brief Learn whether a node is inside an operation output.
  *
  * @param[in] schema Schema node to examine.
  * @return non-zero is the node is in output,
  * @return 0 if it is not.
+ * @return 1 if the node is under an operation output
```
michalvasko, line 789: First line should probably be removed.

michalvasko (changes requested):

compat/compat.c
rkrejci: The function returns ssize_t, so in the end, it would cast anyway. The whole compat's getline implementation would need rewrite, the read line is limited to 256B now.

src/tree_schema_internal.h
rkrejci: agree

michalvasko (approved):

PR #1182, opened 2020-08-27
python binding BUGFIX Should be taking global lock before calling back to python code + Data_Node missing .reset()

johnbeckettn2e:
sysrepo python oper example shows calling .reset() when populating parent.
    
This is missing when using the libyang python binding definition, which prevents returning data in the parent node


michalvasko:
Hi,
okay, seems correct to me, thanks. But maybe you are better off using the [CFFI bindings](https://github.com/CESNET/libyang-python) instead, but it is up to you.

PR #1185, opened 2020-08-31
packages BUGFIX fix local-deb script

idryzhov:


michalvasko:
Hi,
we were actually not using these scripts for building the packages but thanks anyway, they are meant to work.

PR #1187, opened 2020-09-01
libyang FEATURE introduce and use new ly_bool type

rkrejci:
To indicate simple flags or true/false return values, use a standalone
ly_bool type.

We do not use stdbool's bool type to avoid need to mimic all its
features on platforms that do not provide it. ly_bool is just a simple
rename for uint8_t and the reason to use it is only a better readability
of the meaning of the variables or function's return values.

### DIFF
--- a/src/tree_data.c
+++ b/src/tree_data.c
@@ -54,7 +54,7 @@ static LY_ERR lyd_find_sibling_schema(const struct lyd_node *siblings, const str
         struct lyd_node **match);
 
 LY_ERR
-lyd_value_parse(struct lyd_node_term *node, const char *value, size_t value_len, uint8_t *dynamic, uint8_t second,
+lyd_value_parse(struct lyd_node_term *node, const char *value, size_t value_len, ly_bool *dynamic, ly_bool second,
         uint32_t value_hint, LY_PREFIX_FORMAT format, void *prefix_data, const struct lyd_node *tree)
 {
     LY_ERR ret = LY_SUCCESS;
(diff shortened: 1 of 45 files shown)


src/tree_data.c
```diff
@@ -693,7 +693,7 @@ lyd_create_any(const struct lysc_node *schema, const void *value, LYD_ANYDATA_VA
 
 LY_ERR
 lyd_create_opaq(const struct ly_ctx *ctx, const char *name, size_t name_len, const char *value, size_t value_len,
-        uint8_t *dynamic, uint32_t value_hint, LYD_FORMAT format, struct ly_prefix *val_prefs, const char *prefix, size_t pref_len,
+                ly_bool *dynamic, uint32_t value_hint, LYD_FORMAT format, struct ly_prefix *val_prefs, const char *prefix, size_t pref_len,
```
michalvasko, line 696: Formatting.

src/xml.c
```diff
@@ -38,16 +38,16 @@
...
-        uint8_t *dynamic);
+static LY_ERR lyxml_next_attr_content(struct lyxml_ctx *xmlctx, const char **value, size_t *value_len, ly_bool *ws_only,
+        ly_bool *dynamic);
 
 /**
  * @brief Ignore any characters until the delim of the size delim_len is read
  *
  * Detects number of read new lines.
- * Returns 0 if delim was found, 1 if was not.
+ * Returns Boolean valueif delim was found or not.
```
michalvasko, line 48: Typo.

PR #1194, opened 2020-09-05
dictionary REFACTOR unify functions pattern to return LY_ERR

rkrejci:
Change API functions manipulating with libyang dictionary to return
LY_ERR value as a primary indicator of the result and the value from the
dictionary is returned via optional output parameter.

### DIFF
--- a/src/tree_data.c
+++ b/src/tree_data.c
@@ -696,6 +696,7 @@ lyd_create_opaq(const struct ly_ctx *ctx, const char *name, size_t name_len, con
         ly_bool *dynamic, uint32_t value_hint, LYD_FORMAT format, struct ly_prefix *val_prefs, const char *prefix, size_t pref_len,
         const char *module_key, size_t module_key_len, struct lyd_node **node)
 {
+    LY_ERR ret = LY_SUCCESS;
     struct lyd_node_opaq *opaq;
 
     assert(ctx && name && name_len);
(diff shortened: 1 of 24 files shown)


src/tree_data.c
```diff
@@ -709,27 +710,35 @@ lyd_create_opaq(const struct ly_ctx *ctx, const char *name, size_t name_len, con
 
     opaq->prev = (struct lyd_node *)opaq;
 
-    opaq->name = lydict_insert(ctx, name, name_len);
+    LY_CHECK_GOTO(ret = lydict_insert(ctx, name, name_len, &opaq->name), finish);
+    LY_CHECK_GOTO(ret, finish);
```
michalvasko, line 714: Seems redundant.

src/tree_data.c
```diff
@@ -709,27 +710,35 @@ lyd_create_opaq(const struct ly_ctx *ctx, const char *name, size_t name_len, con
...
+
     opaq->format = format;
     if (pref_len) {
-        opaq->prefix.id = lydict_insert(ctx, prefix, pref_len);
+        LY_CHECK_GOTO(ret = lydict_insert(ctx, prefix, pref_len, &opaq->prefix.id), finish);
+        LY_CHECK_GOTO(ret, finish);
```
michalvasko, line 719: Also redundant.

PR #1197, opened 2020-09-14
Refactor preompiled structures

rkrejci:
Do not move features and identities between `lys_module` and `lysc_module`. It is confusing and it is not necessary. The features and identities will be always present in the `lys_module`.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -608,12 +608,12 @@ ylib_feature(struct lyd_node *parent, const struct lys_module *cur_mod)
         return LY_SUCCESS;
     }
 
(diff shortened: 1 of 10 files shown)


michalvasko (approved):
Generally, I have no problems with it and it should be just a simplification. However, the use-case of recompiling a module (which is not present yet) will need some additional minor work. Specifically, changing the features/identity array to the pre-compiled state (with no links to other features because their modules may be freed) and then normally compile again. The features cannot be completely freed because the information about any enabled features would be lost. However, if this is really all that is required, it should not be a problem.

Issue #1027, opened 2020-03-05
libyang2: invalid pointer dereference in lyd_path probably due to type confusion


jvijtiuk:
Hello,

The following file causes yanglint to crash due to use of a invalid pointer:

```
module d{
	namespace "";
	prefix d;
	leaf f {
		type string;
		must ":e";
		default "";
	}
}
```

The issue appears because iter->schema is equal to 0x4, so dereferencing iter->schema->module accesses memory at 0xC in src/tree_data.c:1821.

The schema variable was set to 0x4 previously by parse_leaf in src/parser_yang.c:2090, as LYS_LEAF is 4, which is later interpreted as a pointer to 0x4. The issue seems to be that iter is used as a lyd_node, and the data was previously set through a pointer to lysc_node. Also, the node is actually a part of a lyxp_set which is passed through the call stack, and lyxp_set has a union which can be interpreted as lyd_node or lysc_node.

Changing xpath.c:5179 from LY_VLOG_LYD to LY_VLOG_LYSC seems to fix the issue, but I'm not sure whether that fix is entirely appropriate, or if the issue should be fixed somewhere else. 
Running yanglint with the fix applied produces the following output for the crashing input:
```
err : Unknown module "". (/d:f)
err : Invalid must restriction ":e". (/d:f)
```
If the fix is appropriate I can open a PR with the fix.

Regards,
Juraj

michalvasko:
Hi Juraj,
it should always be possible to determine whether `ctx_node` or `ctx_scnode` can be used based on some function context, that is why this information is not stored directly in the set. It is true for this function as well except it was not done. Should work now.

Regards,
Michal

jvijtiuk:
That makes sense, thanks for the explanation. 
2104e9f fixed this so I'm closing the issue.

PR #1199, opened 2020-09-16
libyang2: fuzz regression tests

jvijtiuk:
Hello,

this PR adds regression tests for previous issues that the fuzzer has found in XML and YANG parsers.
The motivation for this are previous cases where the same input file would cause issues at multiple points in time. For examle #960 first caused a null dereference, and later on another invalid pointer dereference in #1027. Additionally, other projects like [OpenSSL](https://github.com/openssl/openssl/blob/master/fuzz/test-corpus.c) do the same form of regression testing with previously problematic fuzz inputs.

This works by reusing the fuzz harnesses in tests/fuzz/ and adding a separate test target for every harness. If `ENABLE_BUILD_TESTS` is active, the targets will be built and used as simple wrappers around the functions that would usually be fuzzed.

The files that previously caused issues are now located in `tests/fuzz/corpus/${target_name}`, in a separate directory for every target. I've named the files by the issue/PR number they originally appeared in.
The actual add_test command invokes a simple C executable, fuzz_regression_test that receives a fuzz target and a path to the inputs directory and then passes every input file to the fuzz harness.

The tests' WORKING_DIRECTORY is set to ${CMAKE_BINARY_DIR}/tests/fuzz/${target_name} as that was the easiest way to get the regression testing setup to work with issues which were caused by multiple files.
I thought about receiving the search dir as a run time argument in the fuzz harness. However, the fuzz harnesses primarily use the LLVMFuzzerTestOneInput LLVM fuzzer function, which enables the harnesses to be used with multiple fuzzers in an easy way. As there was no easy way to receive the search dir as an argument, I opted to keep the harnesses easily portable and solve the problem of testing YANG models with includes through changing the WORKING_DIRECTORY.

I've included basic descriptions in README flies.

The PR also includes updates to lys_parse_mem and lyd_parse_mem fuzz harnesses to work with API changes introduced in 7931b19, since after those changes the previous harnesses couldn't be built.

If there's anything that should be changed or improved please let me know.

Regards,
Juraj


jvijtiuk:
There's still some issues with this, like some unused variables I have to remove, and a check for ENABLE_TESTING instead of ENABLE_BUILD_TESTS, so the PR isn't ready to be merged yet, sorry for that. I'm working on fixing those issues now

jvijtiuk:
I've fixed the issues with the fuzz regression tests, so I think this PR should be fine now. 
However, some of the tests are failing, as some of the old fuzzing inputs for the YANG parser are now causing crashes again. I'm looking into the issues and will report those in separate issues or PRs if I manage to fix them.

michalvasko:
Hi,
after briefly looking at it, it seems useful. But please add `fuzz_` prefix for these tests so they are clearly differentiated from unite tests (`utest`), thanks.

rkrejci:
I've re-run Travis CI - please check the logs, there is some warning in `tests/fuxx/main.c`. Is it reasonable to make the fuzzer tests to run in Travis CI (or Github Actions)? Maybe as a separate job?

jvijtiuk:
> after briefly looking at it, it seems useful. But please add fuzz_ prefix for these tests so they are clearly > > differentiated from unit tests (utest), thanks.

I've added a fuzz_ prefix to the fuzz regression tests, and changed the _fuzz postfix to _fuzz_harness for the fuzz harnesses, to make the distinction more clear.

>I've re-run Travis CI - please check the logs, there is some warning in tests/fuxx/main.c. Is it reasonable to make the fuzzer tests run in Travis CI (or Github Actions)? Maybe as a separate job?

The warnings should be fixed now. The fuzz regression tests are already running on Travis CI, as I added the tests via CMake's add_test, so make test automatically runs those too. The [latest](https://travis-ci.org/github/CESNET/libyang/jobs/729926309) Travis build log has them listed as passing. 

The tests could run as a separate job, probably with a separate target, in a similar way that coverage is a separate target.

michalvasko:
Okay, I tried to compile it and it seems to be working but I would like some more details. When I ran `make test`, I got 2 new tests `fuzz_lys_parse_mem_test` and `fuzz_lyd_parse_mem_test` that took milliseconds to complete. Meaning they are not fuzz tests, I think, so what exactly are they? Do they simply try to parse schema/data files that previously caused crashes found by fuzzing? If that is the case, something like `regress_fuzz_lys_parse_mem` seems more appropriate as the name since they are not fuzz tests but regression tests.

Also, I think the README is outdated because I do think the mentioned binary for fuzzing is being built, instead there are several of them.

jvijtiuk:
> Meaning they are not fuzz tests, I think, so what exactly are they? Do they simply try to parse schema/data files that previously caused crashes found by fuzzing?

Yes, those are regression tests for previously found fuzzing crashes, not fuzz tests. The fuzz tests are separate executables, generated only when ```-DENABLE_FUZZ_TARGETS``` is enabled. They are named similarly to the regression tests, but have a ```fuzz_harness``` postfix in the name. So, the regression test will then be called ```regress_fuzz_lyd_parse_mem``` and the fuzz test will be called ```lyd_parse_mem_fuzz_harness```.  I've updated the prefix to ```regress_fuzz_``` and tested that everything still works.

>Also, I think the README is outdated because I do think the mentioned binary for fuzzing is being built, instead there are several of them.

I've updated the README to reference ```lyd_parse_mem_fuzz_harness```, instead of the old ```yangfuzz```, and updated some other minor things in the README.

michalvasko:
Okay, seems good to me now, merging.

PR #1200, opened 2020-09-17
libyang2: schema tree BUGFIX lys_compile_unres_xpath when condition null dereference

jvijtiuk:
Hello,

this PR fixes a NULL dereference in lys_compile_unres_xpath. The file causing the crash is below:

```
module d00000000 {
	namespace "n";
    prefix d;
    leaf l1 {
        type string;
        when "/l0{k='when']";
    }
}
```

This was found through the fuzz regression tests in #1199. As fuzz regression testing uses the lys_parse_mem fuzz harness for testing, and #1199 contains a fix or the harness to bring it up to date with the new API this issue can't be reproduced with it on the libyang2 branch. However, I've attached an example that can be used to reproduce the issue below.

```
#include <stdio.h>
#include <libyang/libyang.h>

	struct ly_ctx *ctx = NULL;
	FILE *f = NULL;
	size_t len = 0;
	int err = 0;
	char *data = NULL;

	if (argc != 2) {
		printf("invalid number of arguments\n");
		return -1;
	}

	ly_log_options(0);
	err = ly_ctx_new(NULL, 0, &ctx);
	if (err != LY_SUCCESS) {
		printf("context fail\n");
		return -1;
	}

	f = fopen(argv[1], "r");
	if (f == NULL) {
		printf("fopen fail\n");
		return -1;
	}

	fseek(f, 0, SEEK_END);
	len = ftell(f);
	fseek(f, 0, SEEK_SET);

	data = malloc(len + 1);
	if (data == NULL) {
		printf("malloc fail\n");
		fclose(f);
		return -1;
	}

	fread(data, len, 1, f);
	data[len] = 0;

	lys_parse_mem(ctx, data, LYS_IN_YANG, NULL);
	fclose(f);
	free(data);

	return 0;
}
```

When reproducing the crash, libyang logging has to be disabled, which I've done with `ly_log_options(0)`. Otherwise the crash doesn't seem to appear.

### DIFF
--- a/src/tree_schema_compile.c
+++ b/src/tree_schema_compile.c
@@ -6958,7 +6958,11 @@ lys_compile_unres_xpath(struct lysc_ctx *ctx, const struct lysc_node *node)
         ret = lyxp_atomize(when[u]->cond, LY_PREF_SCHEMA, when[u]->module, when[u]->context ? when[u]->context : node,
                            when[u]->context ? LYXP_NODE_ELEM : LYXP_NODE_ROOT_CONFIG, &tmp_set, opts);
         if (ret != LY_SUCCESS) {
-            LOGVAL(ctx->ctx, LY_VLOG_LYSC, node, LYVE_SEMANTICS, "Invalid when condition \"%s\".", when[u]->cond->expr);
+	    if (when[u]->cond == NULL) {
+                LOGVAL(ctx->ctx, LY_VLOG_LYSC, node, LYVE_SEMANTICS, "Invalid, when condition shouldn't be NULL");
+	    } else {
+                LOGVAL(ctx->ctx, LY_VLOG_LYSC, node, LYVE_SEMANTICS, "Invalid when condition \"%s\".", when[u]->cond->expr);
+	    }
             goto cleanup;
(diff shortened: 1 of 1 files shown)


rkrejci:
I'm afraid that your fix just fixes a sign not the reason of the problem. And I'm afraid that the problem you discovered is complex.

The point is in disabling logging because in that case the error information is not stored in the context. But unfortunately, we apparently have some code (checking the `cond` member in `lys_compile_when()`) which depends on having error information stored in the context.

I'm not sure about the solution. One option is to avoid the possibility to switch off storing error information in the context. The other option is to avoid using the stored error information inside the library. The second option is probably more challenging, but probably more correct and also follows what we are trying to fulfill in public API - all the functions are supposed to return LY_ERR, other return information are provided via output parameters of the function.

What do you think, @michalvasko ?

michalvasko:
Hi,
this should be fixed now by @rkrejci pull request.

jvijtiuk:
Thanks, #1205 fixed this. The file doesn't crash on my system anymore, so I'm closing the issue.

PR #1206, opened 2020-09-21
xpath REFACTOR internal functions should always return LY_ERR

michalvasko:


src/tree_schema_compile.c
```diff
@@ -2770,10 +2770,10 @@ lys_compile_type_(struct lysc_ctx *ctx, struct lysp_node *context_node_p, uint16
             lref->require_instance = 1;
         }
         if (type_p->path) {
-            lref->path = lyxp_expr_dup(ctx->ctx, type_p->path);
+            lyxp_expr_dup(ctx->ctx, type_p->path, &lref->path);
```
rkrejci, line 2773: isn't the point of the change to be able to check for the error return code of the modified functions? The same applies for all the changes in this file.

rkrejci (changes requested):

src/tree_schema_compile.c
michalvasko: Yes, for cases where the error was checked, I did not change the functionality. But okay, it is checked now and the whole function was refactored because the error handling was weird (probably outdated).

rkrejci (approved):

PR #1208, opened 2020-09-22
add escape character in xml attr-value

olderma:
struct lyxml_attr *a, the a->ns->prefix is "yang", a->name is "key", a->value is "[name=\"test\"]",  in dump_elem function process the attributes, the result is `<interface yang:key="[name=\"test\"]">`, this is invalid xml message,it should be `<interface yang:key="[name=&quot;test&quot;]">`


michalvasko:
Thanks, merged into `devel`.

PR #1209, opened 2020-09-23
fix calculate json-anydata length error， when leaf/leaflist value con… … f0067eb …tains ‘{' or '}'

olderma:
rpc-message ： {"ietf-netconf-nmda:edit-data":{"datastore":"ietf-datastores:running","config":{password:"Root{{{{123"}}}

when calculate 'password' len， the result is wrong.

### DIFF
--- a/src/parser_json.c
+++ b/src/parser_json.c
@@ -398,22 +398,23 @@ json_get_anydata(struct lyd_node_anydata *any, const char *data)
     do {
         switch (data[len]) {
         case '{':
-            if (skip == 0) {
+            if (!skip) {
                 c++;
             }
             break;
         case '}':
-            if (skip == 0) {
+            if (!skip) {
                 c--;
             }
             break;
         case '\\':
-            if (skip == 1 && data[len + 1]) {
+            /* when parsing a string, ignore escaped quotes to not end it prematurely */
+            if (skip && data[len + 1]) {
                 len++;
             }
             break;
         case '\"':
-            skip = (skip == 1) ? 0 : 1;
+            skip = !skip;
             break;
         default:
             break;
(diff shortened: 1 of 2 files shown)


src/parser_json.c
```diff
@@ -398,10 +398,22 @@ json_get_anydata(struct lyd_node_anydata *any, const char *data)
     do {
         switch (data[len]) {
         case '{':
-            c++;
+            if (skip == 0) {
```
michalvasko, line 401: Use `!skip` condition instead.

src/parser_json.c
```diff
@@ -398,10 +398,22 @@ json_get_anydata(struct lyd_node_anydata *any, const char *data)
...
+                c++;
+            }
             break;
         case '}':
-            c--;
+            if (skip == 0) {
```
michalvasko, line 406: Replace with `!skip`.

src/parser_json.c
```diff
@@ -398,10 +398,22 @@ json_get_anydata(struct lyd_node_anydata *any, const char *data)
...
+                c--;
+            }
+            break;
+        case '\\':
+            if (skip == 1 && data[len + 1]) {
```
michalvasko, line 411: Replace `skip == 1` with just `skip` and please add a comment with an explanation. Something like `when parsing a string, ignore escaped quotes to not end it prematurely`.

src/parser_json.c
```diff
@@ -398,10 +398,22 @@ json_get_anydata(struct lyd_node_anydata *any, const char *data)
...
+                len++;
+            }
+            break;
+        case '\"':
+            skip = (skip == 1) ? 0 : 1;
```
michalvasko, line 416: Could be replaced with `skip = !skip;`.

michalvasko (changes requested):

michalvasko:
Hi,
the idea is right, this was not handled, but please keep the code clean by treating boolean values the way they should be.

olderma:
> Hi,
> the idea is right, this was not handled, but please keep the code clean by treating boolean values the way they should be.

ok，I have fix the issues.

michalvasko (approved):

michalvasko:
Okay, looks good now, thanks.

PR #1222, opened 2020-10-03
build FEATURE new build type to generate only docs without building sources

rkrejci:
Adds new DocOnly build type for cmake to allow generate makefile only to
build doxygen documentation. This allows to create docs even in
environment with missing build dependencies (pcre2).

The change require some refactoring of CMakeList.txt to have all the
variables available before processing cmake build types and preparing
specific make targets.

### DIFF
--- a/CMakeModules/Doc.cmake
+++ b/CMakeModules/Doc.cmake
@@ -1,5 +1,3 @@
-cmake_minimum_required(VERSION 2.8.12)
-
 # Prepare building doxygen documentation
 macro(LIBYANG_DOC)
     find_package(Doxygen)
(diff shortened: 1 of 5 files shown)


CMakeModules/Doc.cmake
```diff
@@ -0,0 +1,20 @@
+cmake_minimum_required(VERSION 2.8.12)
```
michalvasko, line 1: This command may cause problems with [policies](https://cmake.org/cmake/help/v3.0/command/cmake_minimum_required.html) so I would avoid using it at other places than the main `CMakeLists.txt`.

michalvasko (changes requested):

CMakeModules/Doc.cmake
rkrejci: this is not the only module using it, so I've added a separated commit removing it from all the cmake modules in libyang.

michalvasko: Okay, that should probably be our policy from now.

michalvasko (approved):

PR #1235, opened 2020-10-16
Refactor files schema compile

michalvasko:


rkrejci (approved):

src/path.c
```diff
@@ -246,10 +246,17 @@ ly_path_parse(const struct ly_ctx *ctx, const struct lysc_node *ctx_node, const
...
         /* is the path relative? */
         if (lyxp_next_token(NULL, exp, &tok_idx, LYXP_TOKEN_OPER_PATH)) {
-            /* '..' */
-            while ((lref == LY_PATH_LREF_TRUE) && !lyxp_next_token(NULL, exp, &tok_idx, LYXP_TOKEN_DDOT)) {
-                /* '/' */
-                LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, &tok_idx, LYXP_TOKEN_OPER_PATH), ret = LY_EVALID, error);
+            /* relative path check specific to leafref */
+            if (lref == LY_PATH_LREF_TRUE) {
+                /* mandatory '..' */
+                LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, &tok_idx, LYXP_TOKEN_DDOT), ret = LY_EVALID, error);
```
rkrejci, line 252: Maybe some explaining error message would be useful, this way there is just "unexpected token" message, but as a user I would appreciate what is expected in case it is a specific token.

michalvasko: Fixed.

PR #1237, opened 2020-10-19
Refresh leaflist flag when adding attrs circularly.

zhuxiaoyun138159:
In the first cycle, if iter->index is 1, flag_leaflist is set to 1.In the next cycle, flag_leaflist is still 1 although the iter->index is 0.


michalvasko:
Seems okay, the CI has some unrelated problem.

PR #1244, opened 2020-10-20
libyang REFACTOR printer -> in, parser -> out

michalvasko:


src/parser_data.h
```diff
@@ -156,6 +156,11 @@ struct ly_in;
                                                  revision. */
 /** @} dataparseroptions */
 
+/**
```
rkrejci, line 159: it should be part of the `dataparseroptions` group, shouldn't it?

src/parser_data.h
```diff
@@ -184,6 +189,11 @@ struct ly_in;
 
 /** @} datavalidationoptions */
 
+/**
```
rkrejci, line 192: it should be part of the `datavalidationoptions` group, shouldn't it?

rkrejci (changes requested):

src/parser_data.h
michalvasko: Well, it was not public before and I moved it here only so that the bits and their mask are together and one is not modified without the other. But it is not really needed to be used so that is why I put it separately. I do not care either way, it can be a part of the options.

rkrejci: I would put them together, otherwise it is difficult to find it in doxygen docs

src/parser_data.h
michalvasko: Fixed.

src/parser_data.h
michalvasko: Fixed.

rkrejci (approved):

PR #1247, opened 2020-10-21
Solve the problem of performance degradation after repeated insertion and deletion of hash tables

shuailily0108:
After the hash table is repeatedly inserted and deleted, the status of each record is deleted. When a new record is inserted, the entire hash table will be traversed, resulting in performance degradation.The solution is to count the delete state in the hash table. When deleting records, count the idle state, and re-hash when the idle state is less than 2%.

### DIFF
--- a/src/hash_table.c
+++ b/src/hash_table.c
@@ -326,7 +326,7 @@ lyht_new(uint32_t size, uint16_t val_size, values_equal_cb val_equal, void *cb_d
 
     ht->used = 0;
     ht->size = size;
-    ht->deleted = 0;
+    ht->invalid = 0;
     ht->val_equal = val_equal;
     ht->cb_data = cb_data;
     ht->resize = (uint16_t)resize;
@@ -411,7 +411,7 @@ lyht_resize(struct hash_table *ht, int enlarge)
 
     /* reset used, it will increase again */
     ht->used = 0;
-    ht->deleted = 0;
+    ht->invalid = 0;
 
     /* add all the old records into the new records array */
     for (i = 0; i < old_size; ++i) {
@@ -765,7 +765,7 @@ lyht_insert_with_resize_cb(struct hash_table *ht, void *val_p, uint32_t hash,
     /* insert it into the returned record */
     assert(rec->hits < 1);
     if (rec->hits < 0) {
-        --ht->deleted;
+        --ht->invalid;
     }
     rec->hash = hash;
     rec->hits = 1;
@@ -877,7 +877,7 @@ lyht_remove_with_resize_cb(struct hash_table *ht, void *val_p, uint32_t hash, va
     /* check size & shrink if needed */
     ret = 0;
     --ht->used;
-    ++ht->deleted;
+    ++ht->invalid;
     if (ht->resize == 2) {
         r = (ht->used * 100) / ht->size;
         if ((r < LYHT_SHRINK_PERCENTAGE) && (ht->size > LYHT_MIN_SIZE)) {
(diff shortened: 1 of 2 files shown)


michalvasko:
> When a new record is inserted, the entire hash table will be traversed, resulting in performance degradation.

This is not accurate. There is [an optimization](https://github.com/CESNET/libyang/blob/master/src/hash_table.c#L522) that should cause such a situation to happen only once. However, this is quite a complex theoretical problem so I may be missing something. But I would still like you to include more details about the exact problem because shrinking the hash table does not really seem to me is generally the right idea, performance-wise.

shuailily0108:
> > When a new record is inserted, the entire hash table will be traversed, resulting in performance degradation.
> 
> This is not accurate. There is [an optimization](https://github.com/CESNET/libyang/blob/master/src/hash_table.c#L522) that should cause such a situation to happen only once. However, this is quite a complex theoretical problem so I may be missing something. But I would still like you to include more details about the exact problem because shrinking the hash table does not really seem to me is generally the right idea, performance-wise.

michalvasko:
Okay, thanks for these numbers. Based on them it seems to me the problem starts to appear at around <2% idle (empty) records so why did you decide for <50%? But it seems to me your idea is correct, we just need to tune all the details. I can do that myself or I can make a full review so that you can, whichever you prefer.

shuailily0108:
> Okay, thanks for these numbers. Based on them it seems to me the problem starts to appear at around <2% idle (empty) records so why did you decide for <50%? But it seems to me your idea is correct, we just need to tune all the details. I can do that myself or I can make a full review so that you can, whichever you prefer.

Sorry, I made a mistake for 50%.  According to the test data, when the idle less than 2%, the performance is greatly reduced,so we rehash the table at this point.Please fully review the code, thanks a lot!

src/hash_table.c
```diff
@@ -880,11 +886,24 @@ lyht_remove_with_resize_cb(struct hash_table *ht, void *val_p, uint32_t hash, va
...
 
             /* shrink */
-            ret = lyht_resize(ht, 0);
+            ret = lyht_resize(ht, -1);
 
             if (resize_val_equal) {
                 lyht_set_cb(ht, old_val_equal);
             }
+        } else {
+            if ((ht->size > LYHT_MIN_SIZE) && (ht->size - ht->used - ht->deleted) < (ht->size / 50)) {
```
michalvasko, line 895: Expression `(ht->size > LYHT_MIN_SIZE)` seems redundant here. Also, refactor the rest of the expression so that the current percentage of `empty` (that is the term used in the code) records is calculated and then compared to a macro that will be set to `2`. Similar to shrinking/enlarging. Finally, this check should be performed every time the hash table is not shrunk (even if `ht->resize` is not `2`). And please commend the code more.

src/hash_table.h
```diff
@@ -75,6 +75,7 @@ struct ht_rec {
 struct hash_table {
     uint32_t used;        /* number of values stored in the hash table (filled records) */
     uint32_t size;        /* always holds 2^x == size (is power of 2), actually number of records allocated */
+    uint32_t deleted;     /* number of delete records in the hash table */
```
michalvasko, line 78: Please stick to the terminology, this should be `invalid`. The comment can mostly stay, the meaning is fine, just some explanation to add that deleted records are invalid.

michalvasko (changes requested):

michalvasko:
Also, please change the base branch to `devel`, we are not merging PRs to `master` directly.

src/hash_table.h
shuailily0108: lastest code is :
uint32_t invalid;     /* number of invaild records in the hash table (deleted records) */

PR #1248, opened 2020-10-22
Lysc tree dfs

michalvasko:


src/tree_schema.h
```diff
@@ -1857,13 +1859,35 @@ const struct lysc_notif *lysc_node_notifs(const struct lysc_node *node);
 
 /**
  * @brief Get the children linked list of the given (compiled) schema node.
- * Decides the node's type and in case it has a children list, returns it.
+ * Skips over input and output nodes.
```
rkrejci, line 1854: please refer alternative lysc_node_children_all() and vice versa

src/tree_schema.h
```diff
@@ -1857,13 +1859,35 @@ const struct lysc_notif *lysc_node_notifs(const struct lysc_node *node);
...
+/**
+ * @brief Get the children linked list of the given (compiled) schema node.
+ * Returns all children node types including input and output.
+ *
+ * @param[in] node Node to examine.
+ * @param[in] flags Config flag to distinguish input (LYS_CONFIG_W) and output (LYS_CONFIG_R) child in case of RPC/action node.
+ * @return Children linked list if any,
+ * @return NULL otherwise.
+ */
+const struct lysc_node *lysc_node_children_all(const struct lysc_node *node, uint16_t flags);
```
rkrejci, line 1871: the suffix '_all' is a little weird (all children, all parents), but I don't have any alternative to it :(

michalvasko: I have used `full` for the DFS so that may be an alternative, I do not care either way.

rkrejci: I like the '_data', but since it may be confusing with the "schema nodes that can be instantiated in data tree" meaning, I think that the '_full' suffix will be the best.

michalvasko: Fixed.

src/tree_schema.h
michalvasko: Fixed.

rkrejci (approved):

PR #1249, opened 2020-10-22
context FEATURE flag for implementing referenced modules

michalvasko:
Also some refactoring of flag names included.


rkrejci (approved):
Next time, the type options change should be in a separate commit.

PR #1250, opened 2020-10-22
Solve the problem of performance degradation after repeated insertion and deletion of hash tables

shuailily0108:
After the hash table is repeatedly inserted and deleted, the status of each record is deleted. When a new record is inserted, the entire hash table will be traversed, resulting in performance degradation.The solution is to count the delete state in the hash table. When deleting records, count the idle state, and re-hash when the idle state is less than 2%.

### DIFF
--- a/src/hash_table.c
+++ b/src/hash_table.c
@@ -896,7 +896,7 @@ lyht_remove_with_resize_cb(struct hash_table *ht, void *val_p, uint32_t hash, va
     }
     
     r = ((ht->size - ht->used - ht->invalid) * 100) / ht->size;
-    if ((r < LYHT_REHASH_PERCENTAGE) && (ht->size > LYHT_MIN_SIZE)) {
+    if (r < LYHT_REHASH_PERCENTAGE) {
         if (resize_val_equal) {
             old_val_equal = lyht_set_cb(ht, resize_val_equal);
         }
--- a/src/hash_table.c
+++ b/src/hash_table.c
@@ -326,6 +326,7 @@ lyht_new(uint32_t size, uint16_t val_size, values_equal_cb val_equal, void *cb_d
 
     ht->used = 0;
     ht->size = size;
+    ht->invalid = 0;
     ht->val_equal = val_equal;
     ht->cb_data = cb_data;
     ht->resize = (uint16_t)resize;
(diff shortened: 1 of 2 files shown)


src/hash_table.c
```diff
@@ -880,14 +887,28 @@ lyht_remove_with_resize_cb(struct hash_table *ht, void *val_p, uint32_t hash, va
...
+            ret = lyht_resize(ht, -1);
 
             if (resize_val_equal) {
                 lyht_set_cb(ht, old_val_equal);
             }
         }
     }
+    
+    r = ((ht->size - ht->used - ht->invalid) * 100) / ht->size;
+    if ((r < LYHT_REHASH_PERCENTAGE) && (ht->size > LYHT_MIN_SIZE)) {
```
michalvasko, line 899: Like I said before, `(ht->size > LYHT_MIN_SIZE)` is redundant because you are not making the table smaller.

src/hash_table.h
```diff
@@ -52,6 +52,9 @@ typedef int (*values_equal_cb)(void *val1_p, void *val2_p, int mod, void *cb_dat
 /** when the table is less than this much percent full, it is shrunk (half the size) */
 #define LYHT_SHRINK_PERCENTAGE 25
 
+/** when the table is less than this much percent full, it is rehash */
```
michalvasko, line 55: Incorrect comment, something like `when the table has less than this much percent empty records, it is rehashed to get rid of all the invalid records` would be better.

src/hash_table.h
```diff
@@ -75,6 +78,7 @@ struct ht_rec {
 struct hash_table {
     uint32_t used;        /* number of values stored in the hash table (filled records) */
     uint32_t size;        /* always holds 2^x == size (is power of 2), actually number of records allocated */
+    uint32_t invalid;     /* number of invaild records in the hash table (deleted records) */
```
michalvasko, line 81: Typo `invaild`.

michalvasko (changes requested):
Better, but there are still a few things to improve.

src/hash_table.c
shuailily0108: modify

michalvasko (approved):

PR #1252, opened 2020-10-23
Invalid member not copied in lyht_dup

shuailily0108:
in function lyht_dup，invalid member is not copied.


michalvasko:
It was already fixed.

PR #1253, opened 2020-10-24
fix calculate key-value length error， when key-value contains quotation. 

olderma:
When using ly_ctx_get_node to get schema node, if there are double/single quotation in data path,  it will parse error.
ex : id: interface[name="test\"dest"]
the name value is test"dest， in parse_schema_json_predicate function get the value is test


michalvasko:
Alright, the patch seems fine even though it could have been written as a much shorter `for` loop.

PR #1254, opened 2020-10-27
fix calculate key-value length error， when key-value contains quotation & update pr 1253

olderma:
fix calculate key-value length error， when key-value contains quotation & update pr 1253


michalvasko:
Right, you even rewritten the loop exactly the way I expected it to be, thanks.

PR #1255, opened 2020-10-30
C++: Fix return type

syyyr:
The type was wrong, so I fixed it


jktjkt (approved):

michalvasko:
We generally try to avoid using `bool` type because of compatibility problems and it not being available on all platforms (with its special properties). So the return type is intentional.

syyyr:
`bool` has always been a part of C++, I'm not aware of any platforms that don't support C++ `bool`.

michalvasko:
Right, I did not realize the situation in C++ is very different from the one in C. Then it should be fine.

PR #1259, opened 2020-11-04
Fix libyang not working in winter in Central Europe between 23:00 and midnight

syyyr:
I work late into the night, so I really need this fix. Feel free to change the commit message and what not


rkrejci (approved):

michalvasko:
Right, thanks, it seems getting that check right is way more difficult than I thought.

PR #1260, opened 2020-11-04
Fixes

michalvasko:


src/parser_yang.c
```diff
@@ -4483,6 +4483,46 @@ parse_submodule(struct lys_yang_parser_ctx *ctx, struct ly_in *in, struct lysp_s
     return ret;
 }
 
+/**
+ * @brief After all statements were parsed, skip any whitespaces and comments.
```
rkrejci, line 4487: what about the beginning of the module, shouldn't we support the same also before the module definition starts?

michalvasko: I suppose, fixed.

rkrejci (approved):

PR #1261, opened 2020-11-04
libyang2: travis FEATURE add test job with ASAN and UBSAN

jvijtiuk:
Hello,

I've added a test job with ASAN and UBSAN enabled that runs libyang unit tests to detect issues that valgrind might miss.
Valgrind tests are disabled for this build as valgrind can't run with sanitizers enabled.

There seems to be no need to explicitly enable leak sanitizer as it is already included with [ASAN](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer).

There doesn't seem to be any need for enabling memory sanitizer either, as the [wiki](https://github.com/google/sanitizers/wiki/MemorySanitizer) states that it implements a subset of functionality of valgrind.

For reference [here](https://github.com/google/sanitizers/wiki/AddressSanitizerComparisonOfMemoryTools) is a comparison between ASAN and valgrind.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,6 +89,20 @@ jobs:
       after_success:
         - bash <(curl -s https://codecov.io/bash)
     - stage: Test
+      name: Linux with CLang ASAN and UBSAN
+      os: linux
+      compiler: clang
+      before_install:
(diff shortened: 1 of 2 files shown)


michalvasko:
Okay, looks good and working, thanks.

PR #1263, opened 2020-11-05
Code maintenance

rkrejci:
Revision of includes and formatting.


michalvasko (approved):
No problem except I am not sure why the plugins are now part of `tree_schema.h` and what was the problem of the previous solution.

rkrejci:
Well, after reviewing it once again, I'm not sure that it was the right decision. The `yangre` code made me feel that caller (a generic one, not implementing any data type plugin) may want to call the plugin's callbacks. And the `plugins_types.h` was originally supposed to be just an API extension for type plugin implementation, so I'm moving the plugin structure and callbacks typedefs into the `tree_schema.h`. But maybe the code in `yangre` is wrong and should be adapted to the last changes in data plugins API.

michalvasko:
Right, `lys_value_validate` should be possible to use instead of calling the plugin store callback directly.

rkrejci:
@michalvasko you are right, I have reverted the change of the type plugins definitions, they are back in `plugins_types.h`, so hidden from a standard user that includes `libyang.h`. `yangre` code is updated to use `lys_value_validate()`. And I have added one commit regarding warnings when compiling the Release build type.

michalvasko (approved):

PR #1267, opened 2020-11-09
Fixes

michalvasko:


src/tree_data.h
```diff
@@ -770,6 +771,16 @@ struct lyd_node *lyd_child_no_keys(const struct lyd_node *node);
  */
 const struct lys_module *lyd_owner_module(const struct lyd_node *node);
 
+/**
+ * @brief Check whether this data node existence depends on any when conditions. Be careful, this function is
+ * not recursive and checks only conditions that apply to this node directly. Meaning if there are any conditions
+ * associated with any parent of @p node, they are not returned.
```
rkrejci, line 777: maybe noticing why just checking `node->schema->whne` is not enough would be a good motivation to use this function

doc/transition.dox
```diff
@@ -280,6 +280,14 @@
...
  * process before the final validation.
  *
+ * Many validation flags were removed because they became obsolete (LYD_OPT_DESTRUCT, LYD_OPT_WHENAUTODEL,
+ * LYD_OPT_NOEXTDEPS, LYD_OPT_DATA_NO_YANGLIB, LYD_OPT_VAL_DIFF) or we consider them redundant (LYD_OPT_OBSOLETE,
+ * LYD_OPT_NOSIBLINGS, LYD_OPT_DATA_ADD_YANGLIB). As for LYD_OPT_TRUSTED, this option was mostly replaced with
+ * ::LYD_PARSE_ONLY because both skip validation but with the significant difference that LYD_OPT_TRUSTED also sets
+ * the data node flags to be considered validated. The current option ::LYD_PARSE_ONLY does not do this because
+ * there is now much better support for working with non-validated data trees. Also, in case an invalid tree was marked
+ * as validated, it could lead to some undesired behavior, which is now avoided.
+ *
```
rkrejci, line 290: don't you want to add a sentence or two about preserving the tree validated in LYB format?

src/tree_data.h
michalvasko: Fixed.

doc/transition.dox
michalvasko: Fixed.

PR #1269, opened 2020-11-10
libyang2: json parser BUGFIX processing arrays is not correctly handled

jvijtiuk:
Hello,

this PR stops `lyd_parse_json_data` from trying to process JSON data that starts with an array.

This fixes two issues, one where the input was an empty array, and another where the input was an array with data. The issues were discovered by fuzzing `lyd_parse_mem_json`.
When running `lyd_parse_mem_json` under debug mode, the assert for LYJSON_OBJECT would catch both issues, however the release build would segfault in `lydjson_subtree_r` due to null dereference when trying to get the array value. 

I've added checks for empty arrays and arrays so `lyd_parse_json_data` exits if any of the cases are encountered at the start of input. That looks like an appropriate fix since [RFC 7951](https://tools.ietf.org/html/rfc7951) states that the top level JSON encoded YANG data is assumed to be an object.

This PR also adds fuzz regression tests for both issues.

I think that `LY_CHECK_GOTO(status == LYJSON_END || status == LYJSON_OBJECT_EMPTY || status == LYJSON_ARRAY || status == LYJSON_ARRAY_EMPTY, cleanup);` could be replaced with just `LY_CHECK_GOTO(status != LYJSON_OBJECT, cleanup);` and then the assert could be removed too. I've run the tests with that version too and everything seems to be working, so if that is fine I can push that version of the fix.


rkrejci:
I've rewritten the fix and just committed. The problem was in more places, not only in `lyd_parse_json_data()`, so now checking for the top-level object should be much better. Thanks for your fix, I've kept your patch for the fuzzer.

Issue #1268, opened 2020-11-10
How to get the final type in series of typedefs


syyyr:
Hi, I have this kind of program, which prints how many bits are in a typedefed type:
```cpp
#include <iostream>
#include <libyang/libyang.h>

const auto schema = R"(
module test-schema {
    yang-version 1.1;
    namespace "http://example.com";
    prefix "AHOJ";

    typedef myBits {
        type bits {
            bit A;
            bit B;
        }
    }

    leaf bitLeaf {
        type myBits;
    }
}
)";

int main(int argc, char* argv[])
{
    auto ctx = ly_ctx_new(nullptr, 0);
    lys_parse_mem(ctx, schema, LYS_IN_YANG);
    auto schemaNode = (lys_node_leaf*)ly_ctx_get_node(ctx, nullptr, "/test-schema:bitLeaf", 0);
    std::cout << "schemaNode->type.info.bits.count = " << schemaNode->type.info.bits.count << "\n";
    std::cout << "schemaNode->type.der->type.info.bits.count = " << schemaNode->type.der->type.info.bits.count << "\n";
    return 0;
}
```
The count isn't in the immediate `type.info.bits.count`, but rather somewhere deeper, inside the typedef. Is there a way to know how deep I should go when doing `->der` or is there a function which would resolve all typedefs and give me the final type? 

Also, typedefed bits are broken in C++ too. It is impossible retrieve bit values, if the type is a typedef and the reason is that the code doesn't look into them (the code I'm referring to is `libyang::Value::bit`). If there was a function, which would give the resolved type, then it could also be used in the C++ bindings (and "unbreak" them). An example program is here: https://gist.github.com/syyyr/bbb35dd65e747987e1e6785df42fd4aa

michalvasko:
Hi,
you should go through the `der` pointers until you find a type with some `info.bits.count`. We know this is quite a messy solution and *libyang* version 2 compiles types into one coherent type so you do not need to look into any derived types. But, like always, current *libyang* will remain the way it is.

syyyr:
So my fix for the C++ binding uses the correct way. Thanks!

> libyang version 2 compiles types into one coherent type so you do not need to look into any derived types

Nice!

PR #1270, opened 2020-11-11
C++: Fix bit value printing

syyyr:
This seems like an easy thing to forget about, so maybe there should be a function that would always give the resolved base type. I'm not sure whether `lyd_leaf_type` does that (the documentation doesn't say that). Anyway, this patch at least unbreaks bit value retrieval. More info here https://github.com/CESNET/libyang/issues/1268 (also example program).


michalvasko:
Looks good, thanks.

PR #1271, opened 2020-11-12
Fixes

michalvasko:


src/tree_data.h
```diff
@@ -906,19 +906,21 @@ LY_ERR lyd_new_any(struct lyd_node *parent, const struct lys_module *module, con
...
  * @param[in] val_str String form of the value of the metadata. In case of an instance-identifier or identityref
  * value, the JSON format is expected (module names instead of prefixes).
- * @param[out] meta Optional created metadata.
+ * @param[in] clear_dflt Whether to clear the default flag starting from @p parent, recursively all NP containers.
+ * @param[out] meta Optional created metadata. Must be set if @p parent is NULL.
  * @return LY_ERR value.
  */
-LY_ERR lyd_new_meta(struct lyd_node *parent, const struct lys_module *module, const char *name, const char *val_str,
-        struct lyd_meta **meta);
+LY_ERR lyd_new_meta(const struct ly_ctx *ctx, struct lyd_node *parent, const struct lys_module *module, const char *name,
```
rkrejci, line 922: I don't think that the commit is BUGFIX when it changes API

src/tree_data.h
```diff
@@ -477,13 +477,14 @@ struct lyd_meta {
  *
  * The union is used as a reference to the data's module and according to the format, it can be used as a key for
  * ::ly_ctx_get_module_implemented_ns() or ::ly_ctx_get_module_implemented(). While the module reference is always present,
- * the id member can be omitted in case it is not present in the source data as a reference to the default namespace.
+ * the id member can be omitted in case it is not present in the source data as a reference to the default module/namespace.
```
rkrejci, line 480: s/id member/prefix member/ - but rather think about renaming the `name` member. In both cases, the structure is instantiated as `name` member. `name.name` seems weird. However, it is something I can live with.

rkrejci (changes requested):

src/tree_data.h
michalvasko: Right, the API change was probably supposed to be a separate commit, is it a problem if it stays this way?

src/tree_data.h
michalvasko: `id` should be fixed, but as for `name` member, I would say it is the best fitting name even though when accessing it it is repeated. Any ideas? Maybe the structure instances can be called something other than `name`, such as `oname` (opaque name) but I am not sure it would be an improvement.

rkrejci: what about the `id` - imho it is the `identifier` from the YANG Grammar perspective

src/tree_data.h
rkrejci: rewording the message (replacing BUGFIX by CHANGE) should be quite easy with git rebase, I would prefer to change it

src/tree_data.h
michalvasko: I suppose, but these are data, so I am not sure it can really be called that way. In any case, this is a specific identifier `name` and I think it is always best to be as specific as possible. Also, the rest of the members are also identifiers.

src/tree_data.h
michalvasko: Okay, changing the message is not a problem, fixed. Splitting the commit into 2 would be more difficult.

src/tree_data.h
rkrejci: ok, as I said, I can live with it

rkrejci (approved):

PR #1273, opened 2020-11-13
Find xpath speedup

jvportal:
These changes were based on benchmark and flamegraph. The benchmark file "benchmark_lyv1.zip" (the files "frr-filter.yang" and "l1_test.json" are used by program "l1_perf.c") and the interactive flamegraph "l1_perf_2ac5bb59_svg.zip" (needed to zip the SVG file to attach here, maybe Github doesn't accept this format) are attached in this comment. The benchmark program loads module "frr-filter" from yang file, loads a json file with ~550k lines and executes "lyd_find_path" function. The objective is to make the execution of "lyd_find_path" faster.
  
[benchmark_lyv1.zip](https://github.com/CESNET/libyang/files/5534011/benchmark_lyv1.zip)
[l1_perf_2ac5bb59_svg.zip](https://github.com/CESNET/libyang/files/5534026/l1_perf_2ac5bb59_svg.zip)
  
There are 3 changes in this pull request, each in a separated commit.
- The function "lyd_create_leaf", called by "lyd_make_canonical", created a fake node during the execution of "lyd_find_path". Because of this, most of the time spent by this function was in "lyd_hash", "lydict_insert", "lydict_remove", etc. and this was unnecessary, as this node was created to test.
- The function "moveto_node", called by "eval_node_test", used the string dict to improve performance, calling the same functions cited above ("lyd_hash", "lydict_insert", "lydict_remove", etc.). But in fact, in that specific case, it is slower to use the dict (that needs to compute a number from all string characters), it is faster to simply compare strings.
- The option ENABLE_CACHE present on CMake file turned the time of the benchmark program slower, so it was changed to OFF by default.
  
The results are below. The execution time of "lyd_find_path" was cut by ~43%.
  
Results on branch 'devel' at 98fb0d0b:
find 1x (s): 0.097888
same find 10x (s): 0.978431
total prog time (s): 1.337355
total non-find time (s): 0.261036
  
Results on branch 'find_xpath_speedup' at 8fccceba:
find 1x (s): 0.055731
same find 10x (s): 0.557739
total prog time (s): 0.852393
total non-find time (s): 0.238923

### DIFF
--- a/src/tree_data.c
+++ b/src/tree_data.c
@@ -1646,12 +1646,6 @@ lyd_make_canonical(const struct lys_node *schema, const char *val_str, int val_s
 
     assert(schema->nodetype & (LYS_LEAF | LYS_LEAFLIST));
 
-    str = strndup(val_str, val_str_len);
-    if (!str) {
-        LOGMEM(schema->module->ctx);
-        return NULL;
-    }
-
     /* parse the value into a fake leaf */
     struct lyd_node_leaf_list *node_leaf_list;
 
@@ -1661,7 +1655,7 @@ lyd_make_canonical(const struct lys_node *schema, const char *val_str, int val_s
     node_leaf_list->schema = (struct lys_node *)schema;
     node_leaf_list->prev = (struct lyd_node *)node_leaf_list;
     node_leaf_list->value_type = ((struct lys_node_leaf *)schema)->type.base;
-    node_leaf_list->value_str = lydict_insert(schema->module->ctx, str ? str : "", 0);
+    node_leaf_list->value_str = lydict_insert(schema->module->ctx, val_str ? val_str : "", val_str_len);
     node_leaf_list->dflt = 0;
 
     if (!lyp_parse_value(&((struct lys_node_leaf *)schema)->type, &node_leaf_list->value_str, NULL, node_leaf_list, NULL, NULL, 0, 0)) {
@@ -1671,7 +1665,6 @@ lyd_make_canonical(const struct lys_node *schema, const char *val_str, int val_s
 
     node = (struct lyd_node *)node_leaf_list;
 
-    free(str);
     if (!node) {
         return NULL;
     }
--- a/src/tree_data.c
+++ b/src/tree_data.c
@@ -1653,7 +1653,24 @@ lyd_make_canonical(const struct lys_node *schema, const char *val_str, int val_s
     }
 
     /* parse the value into a fake leaf */
-    node = lyd_create_leaf(schema, str, 0, 0);
+    struct lyd_node_leaf_list *node_leaf_list;
+
+    node_leaf_list = calloc(1, sizeof *node_leaf_list);
+    LY_CHECK_ERR_RETURN(!node_leaf_list, LOGMEM(schema->module->ctx), NULL);
+
+    node_leaf_list->schema = (struct lys_node *)schema;
+    node_leaf_list->prev = (struct lyd_node *)node_leaf_list;
+    node_leaf_list->value_type = ((struct lys_node_leaf *)schema)->type.base;
+    node_leaf_list->value_str = lydict_insert(schema->module->ctx, str ? str : "", 0);
+    node_leaf_list->dflt = 0;
+
+    if (!lyp_parse_value(&((struct lys_node_leaf *)schema)->type, &node_leaf_list->value_str, NULL, node_leaf_list, NULL, NULL, 0, 0)) {
+        lyd_free((struct lyd_node *)node_leaf_list);
+        return NULL;
+    }
+
+    node = (struct lyd_node *)node_leaf_list;
+
     free(str);
     if (!node) {
         return NULL;
--- a/src/xpath.c
+++ b/src/xpath.c
@@ -5521,8 +5521,8 @@ moveto_node_check(struct lyd_node *node, enum lyxp_node_type root_type, const ch
             return -1;
         }
     } else {
-        if (strcmp(node_name, "*") && (strncmp(node->schema->name, node_name, node_name_len) ||
-            node->schema->name[node_name_len] != '\0')) {
+        if ((strncmp(node_name, "*", 1) || node_name_len != 1) &&
+            (strncmp(node->schema->name, node_name, node_name_len) || node->schema->name[node_name_len] != '\0')) {
             return -1;
         }
     }
--- a/src/xpath.c
+++ b/src/xpath.c
@@ -5493,15 +5493,17 @@ moveto_snode_root(struct lyxp_set *set, struct lys_node *cur_node, int options)
  *
  * @param[in] node Node to check.
  * @param[in] root_type XPath root node type.
- * @param[in] node_name Node name to move to. Must be in the dictionary!
+ * @param[in] node_name_dict Optional parameter, contains the node name to move to when it is on the string dict.
+ * @param[in] node_name Contains the node name to move to. Used when node_name_dict is NULL (not provided).
+ * @param[in] node_name_len Length of \p node_name.
  * @param[in] moveto_mod Expected module of the node.
  * @param[in] options Whether to apply data node access restrictions defined for 'when' and 'must' evaluation.
  *
  * @return EXIT_SUCCESS on success, EXIT_FAILURE on unresolved when, -1 on error.
  */
 static int
-moveto_node_check(struct lyd_node *node, enum lyxp_node_type root_type, const char *node_name,
-                  struct lys_module *moveto_mod, int options)
+moveto_node_check(struct lyd_node *node, enum lyxp_node_type root_type, const char *node_name_dict,
+                  const char *node_name, int node_name_len, struct lys_module *moveto_mod, int options)
 {
     /* module check */
     if (moveto_mod && (lyd_node_module(node) != moveto_mod)) {
@@ -5514,8 +5516,15 @@ moveto_node_check(struct lyd_node *node, enum lyxp_node_type root_type, const ch
     }
 
     /* name check */
-    if (strcmp(node_name, "*") && !ly_strequal(node->schema->name, node_name, 1)) {
-        return -1;
+    if (node_name_dict) {
+        if (strcmp(node_name_dict, "*") && !ly_strequal(node->schema->name, node_name_dict, 1)) {
+            return -1;
+        }
+    } else {
+        if (strcmp(node_name, "*") && (strncmp(node->schema->name, node_name, node_name_len) ||
+            node->schema->name[node_name_len] != '\0')) {
+            return -1;
+        }
     }
 
     /* when check */
@@ -5581,6 +5590,7 @@ moveto_node(struct lyxp_set *set, struct lyd_node *cur_node, struct lys_module *
             uint16_t qname_len, int options)
 {
     uint32_t i;
+    int comparison_counter = 0;
     int replaced, pref_len, ret;
     const char *ptr, *name_dict = NULL; /* optimalization - so we can do (==) instead (!strncmp(...)) in moveto_node_check() */
     struct lys_module *moveto_mod;
@@ -5621,15 +5631,17 @@ moveto_node(struct lyxp_set *set, struct lyd_node *cur_node, struct lys_module *
         moveto_mod = local_mod;
     }
 
-    /* name */
-    name_dict = lydict_insert(ctx, qname, qname_len);
-
     for (i = 0; i < set->used; ) {
         replaced = 0;
 
         if ((set->val.nodes[i].type == LYXP_NODE_ROOT_CONFIG) || (set->val.nodes[i].type == LYXP_NODE_ROOT)) {
             LY_TREE_FOR(set->val.nodes[i].node, sub) {
-                ret = moveto_node_check(sub, root_type, name_dict, moveto_mod, options);
+                /* avoid using string dict (lydict_insert call) for few node checks */
+                comparison_counter++;
+                if (!name_dict && (comparison_counter > LYXP_MIN_NODE_CHECKS_TO_USE_DICT)) {
+                    name_dict = lydict_insert(ctx, qname, qname_len);
+                }
+                ret = moveto_node_check(sub, root_type, name_dict, qname, qname_len, moveto_mod, options);
                 if (!ret) {
                     /* pos filled later */
                     if (!replaced) {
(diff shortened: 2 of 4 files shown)


michalvasko:
Hi,
okay, firstly, you have chosen one specific use-case of not only `lyd_find_path()` but of the whole libyang API, which gives extremely biased results. I am sure even in your case you are not calling only this single function from libyang but more on that later.

  - You seem to be right that `lyd_make_canonical()` performance is suboptimal and some optimizations are possible. However, your change requires a lot of other unrelated changes in the code and we will simply not accept it, it makes it really difficult to read. I would suggest you look at simply optimizing the function itself and not call `lyd_create_leaf()` but `lyp_parse_value()` directly.
  - You are again right that there are obviously cases when using the dictionary actually slows things down. Specifically, then there are very few nodes that are actually compared. However, I am positive that when there are a lot, this optimization makes the comparison work in `O(n)` time instead of `O(n*m)` where `n` is the number of comparisons and `m` the length of the node name. The problem in this case is that `n` is 3 and so the additional constant complexity of using the dictionary is actually not worth it. So, to make this work well my idea would be to count the number of comparisons and if they reach a specific constant (defined separately with value 10 or something similar), the name is added into dictionary. This way for cases with hundreds of comparisons the dictionary would still be used. Also, when comparing the names with `strcmp`, you do not need to duplicate the name and just use `strncmp`, that should be even faster.
  - Turning the cache off performance-wise is insane, sorry for being so blunt. Yes, in this specific case it may make things a bit faster, but it decreases the average complexity of many libyang operations by one order and I guarantee you would see the difference, even in xpath. What could be done, though, is again use the hashes only in case an xpath set has more than X members (look into `set_insert_node()`), we actually use similar technique for data node children. That should get you the same speedup in this case as you observed when turning the cache off.

jvportal:
Hi Michal,

  Thanks for your answer. We can leave the topic about turning "ENABLE_CACHE OFF" out of this PR, as it is very polemic and would need a lot more of tests and analysis, and focus in the other objectives that are more important.
  
- About lyd_make_canonical() changes, I can rewrite the changes to improve readability, following your advises.

- About when is worth or not to use the dictionary, we have a lot of options, as the one you suggested.

I can get these changes ready soon. I think that a change that provides a significant percentage decrease in some/most execution cases is better than no change.

But what caught my attention was the libyang2 branch performance. The execution time of "lyd_find_xpath" executed once was 0.000024 s, it appears that the time complexity was improved. From this, I have 3 questions (maybe here it is not the best place to ask, but since we are here):

1) Do you see any way to have most of the speed of branch libyang2 in devel (libyang version 1) without refactoring a lot of code? I noticed a lot of changes between devel and libyang2 branches, including changes in structs. Suggestions are appreciated.

2) For someone who has its own project using libyang version 1 and can't/won't adapt at present moment to the code of libyang2, do you think it is too difficult or a bad idea to create a wrapper with the function signatures of libyang version 1 and its implementation calling functions of libyang2?

3) Didn't found a version attached to the message commits of libyang2 branch. The version is annotated in another place or it is still work in progress?

Regards,

J. Portal.

jvportal:
There are 3 changes present in this branch:
  
- "lyd_make_canonical" fake leaf: to call "lyp_parse_value" directly I would need to create either a node or an attribute structure. Used the fake node to have less changes. Still, the fake node is now faster and the changes were fewer.
- "moveto_node" dict usage: as suggested, the dict is used only when there are more calls to moveto_node_check.
- "-O3" compilation optimization: got a little bit more than 9% speedup changing from -O2 to -O3. GCC 9.3.0 used.
  
Times collected:
find 1x (s): 0.073371
same find 10x (s): 0.721139
total prog time (s): 1.056789
total non-find time (s): 0.262279
  
25% of execution time cut for 'find' in this specific benchmark.

jvportal:
Ok, soon I'll push the changes suggested.

jvportal:
Did the changes. Same values for the benchmark used.

src/xpath.c
```diff
@@ -5514,8 +5516,14 @@ moveto_node_check(struct lyd_node *node, enum lyxp_node_type root_type, const ch
...
 
     /* name check */
-    if (strcmp(node_name, "*") && !ly_strequal(node->schema->name, node_name, 1)) {
-        return -1;
+    if (node_name_dict) {
+        if (strcmp(node_name_dict, "*") && !ly_strequal(node->schema->name, node_name_dict, 1)) {
+            return -1;
+        }
+    } else {
+        if (strcmp(node_name, "*") && strncmp(node->schema->name, node_name, node_name_len)) {
```
michalvasko, line 5524: Be careful here, `strncmp` is not enough, the exact length (ending 0) must also be checked.

src/xpath.c
```diff
@@ -5621,15 +5630,16 @@ moveto_node(struct lyxp_set *set, struct lyd_node *cur_node, struct lys_module *
...
-
     for (i = 0; i < set->used; ) {
         replaced = 0;
 
         if ((set->val.nodes[i].type == LYXP_NODE_ROOT_CONFIG) || (set->val.nodes[i].type == LYXP_NODE_ROOT)) {
             LY_TREE_FOR(set->val.nodes[i].node, sub) {
-                ret = moveto_node_check(sub, root_type, name_dict, moveto_mod, options);
+                /* avoid using string dict for few comparisons */
+                comparison_counter++;
+                if (!name_dict && (comparison_counter > LYXP_MIN_COMPARISONS_TO_USE_DICT))
```
michalvasko, line 5640: Please always use `{}` even for one-line conditions or loops.

src/xpath.c
```diff
@@ -5640,7 +5650,8 @@ moveto_node(struct lyxp_set *set, struct lyd_node *cur_node, struct lys_module *
                     }
                     ++i;
                 } else if (ret == EXIT_FAILURE) {
-                    lydict_remove(ctx, name_dict);
+                    if (name_dict)
```
michalvasko, line 5653: Again missing `{}` but you do not even need to test for `NULL` (same problem 3x in the function).

src/xpath.h
```diff
@@ -80,6 +80,9 @@
 #define LYXP_STRING_CAST_SIZE_START 64
 #define LYXP_STRING_CAST_SIZE_STEP 16
 
+/* Number of comparisons when it starts being worth using the string dict*/
```
michalvasko, line 83: Missing space at the end and please be a bit more verbose. Something about node name matching with `strcmp` on few nodes being faster than inserting it into the dictionary and then comparing pointers.

src/xpath.c
```diff
@@ -5493,15 +5493,17 @@ moveto_snode_root(struct lyxp_set *set, struct lys_node *cur_node, int options)
...
  * @param[in] options Whether to apply data node access restrictions defined for 'when' and 'must' evaluation.
+ * @param[in] node_name Contains the node name to move to, used when not in the dict (when node_name_dict is NULL).
+ * @param[in] node_name_len Node name lenght used when the string is not in dict.
  *
  * @return EXIT_SUCCESS on success, EXIT_FAILURE on unresolved when, -1 on error.
  */
 static int
-moveto_node_check(struct lyd_node *node, enum lyxp_node_type root_type, const char *node_name,
-                  struct lys_module *moveto_mod, int options)
+moveto_node_check(struct lyd_node *node, enum lyxp_node_type root_type, const char *node_name_dict,
```
michalvasko, line 5505: Please put the `node_name` parameters together, `moveto_mod` can be the last, I guess.

src/xpath.c
```diff
@@ -5493,15 +5493,17 @@ moveto_snode_root(struct lyxp_set *set, struct lys_node *cur_node, int options)
  *
  * @param[in] node Node to check.
  * @param[in] root_type XPath root node type.
- * @param[in] node_name Node name to move to. Must be in the dictionary!
+ * @param[in] node_name_dict Contains the node name to move to, when it is on the dict. NULL otherwise.
```
michalvasko, line 5496: Doc should probably say that it is an optional name in the dictionary. What you wrote I find a bit confusing.

src/xpath.c
```diff
@@ -5493,15 +5493,17 @@ moveto_snode_root(struct lyxp_set *set, struct lys_node *cur_node, int options)
...
  * @param[in] moveto_mod Expected module of the node.
  * @param[in] options Whether to apply data node access restrictions defined for 'when' and 'must' evaluation.
+ * @param[in] node_name Contains the node name to move to, used when not in the dict (when node_name_dict is NULL).
+ * @param[in] node_name_len Node name lenght used when the string is not in dict.
```
michalvasko, line 5500: Just put `Length of @p node_name.`

src/tree_data.c
```diff
@@ -1653,7 +1653,28 @@ lyd_make_canonical(const struct lys_node *schema, const char *val_str, int val_s
...
+    node_leaf_list->validity = ly_new_node_validity(schema);
+    if (resolve_applies_when(schema, 0, NULL)) {
+        node_leaf_list->when_status = LYD_WHEN;
+    }
+    node_leaf_list->prev = (struct lyd_node *)node_leaf_list;
+    node_leaf_list->value_type = ((struct lys_node_leaf *)schema)->type.base;
+    node_leaf_list->value_str = strdup(str);
+    node_leaf_list->dflt = 0;
+
+    if (!lyp_parse_value(&((struct lys_node_leaf *)schema)->type, &node_leaf_list->value_str, NULL, node_leaf_list, NULL, NULL, 0, 0)) {
```
michalvasko, line 1671: Now that the code can be tailored to the specific use-case, please do so. `lyp_parse_value` does not need `validity` or `when_status` set, so do not bother with that (some minor performance gain).

michalvasko (changes requested):
All the ideas seem correct to me, just some minor problems to fix and it should be good to be merged.

src/xpath.c
```diff
@@ -5514,8 +5516,15 @@ moveto_node_check(struct lyd_node *node, enum lyxp_node_type root_type, const ch
...
     /* name check */
-    if (strcmp(node_name, "*") && !ly_strequal(node->schema->name, node_name, 1)) {
-        return -1;
+    if (node_name_dict) {
+        if (strcmp(node_name_dict, "*") && !ly_strequal(node->schema->name, node_name_dict, 1)) {
+            return -1;
+        }
+    } else {
+        if (strcmp(node_name, "*") && (node->schema->name[node_name_len] != '\0' ||
+            strncmp(node->schema->name, node_name, node_name_len))) {
```
michalvasko, line 5525: It is not safe in this order because until you call `strncmp` you do not know that `node->schema->name` has at least `node_name_len` characters.

src/xpath.h
```diff
@@ -80,6 +80,9 @@
 #define LYXP_STRING_CAST_SIZE_START 64
 #define LYXP_STRING_CAST_SIZE_STEP 16
 
+/* Number of node checks when it starts being worth using the string dict */
```
michalvasko, line 83: Like I said, please be a bit more verbose. Something about node name matching with strcmp on few nodes being faster than inserting it into the dictionary and then comparing pointers.

michalvasko (changes requested):
Okay, last few changes and it should be fine.

src/xpath.c
jvportal: Yes, it was still WIP, I'll change here.

src/tree_data.c
```diff
@@ -1653,7 +1653,24 @@ lyd_make_canonical(const struct lys_node *schema, const char *val_str, int val_s
...
-    node = lyd_create_leaf(schema, str, 0, 0);
+    struct lyd_node_leaf_list *node_leaf_list;
+
+    node_leaf_list = calloc(1, sizeof *node_leaf_list);
+    LY_CHECK_ERR_RETURN(!node_leaf_list, LOGMEM(schema->module->ctx), NULL);
+
+    node_leaf_list->schema = (struct lys_node *)schema;
+    node_leaf_list->prev = (struct lyd_node *)node_leaf_list;
+    node_leaf_list->value_type = ((struct lys_node_leaf *)schema)->type.base;
+    node_leaf_list->value_str = lydict_insert(schema->module->ctx, str ? str : "", 0);
```
jvportal, line 1664: Here I needed to use lydict_insert agian, some tests were failing.

src/xpath.h
jvportal: Ok.

jvportal:
Ready for review.

michalvasko (approved):

michalvasko:
Seems fine to me now but all the tests must pass before I can merge it.

jvportal:
Ok, I'll focus on the tests.

michalvasko:
By the way, you can convert your PR to a draft and change it to a PR only once you think it is ready.

jvportal:
> By the way, you can convert your PR to a draft and change it to a PR only once you think it is ready.

It still runs the CI in as draft?

michalvasko:
Yes, it should.

jvportal:
Added 1 commit to fix problems detected by CI tests.
Ready for review.

src/tree_data.c
```diff
@@ -1646,14 +1646,33 @@ lyd_make_canonical(const struct lys_node *schema, const char *val_str, int val_s
 
     assert(schema->nodetype & (LYS_LEAF | LYS_LEAFLIST));
 
+    /* parse the value into a fake leaf */
+    struct lyd_node_leaf_list *node_leaf_list;
+
+    node_leaf_list = calloc(1, sizeof *node_leaf_list);
+    LY_CHECK_ERR_RETURN(!node_leaf_list, LOGMEM(schema->module->ctx), NULL);
+
     str = strndup(val_str, val_str_len);
```
michalvasko, line 1655: This string duplication seems useless, just assign the value from the dictionary directly.

michalvasko (changes requested):
Another detail I noticed just now.

src/tree_data.c
jvportal: Done.

src/tree_data.c
```diff
@@ -1646,15 +1646,25 @@ lyd_make_canonical(const struct lys_node *schema, const char *val_str, int val_s
...
+    /* parse the value into a fake leaf */
+    struct lyd_node_leaf_list *node_leaf_list;
+
+    node_leaf_list = calloc(1, sizeof *node_leaf_list);
+    LY_CHECK_ERR_RETURN(!node_leaf_list, LOGMEM(schema->module->ctx), NULL);
+
+    node_leaf_list->schema = (struct lys_node *)schema;
+    node_leaf_list->prev = (struct lyd_node *)node_leaf_list;
+    node_leaf_list->value_type = ((struct lys_node_leaf *)schema)->type.base;
+    node_leaf_list->value_str = lydict_insert(schema->module->ctx, val_str ? val_str : "", 0);
```
michalvasko, line 1658: You need to use `val_str_len` for `val_str` length instead of `0`.

michalvasko (changes requested):

src/tree_data.c
jvportal: Done.

michalvasko (approved):

PR #1274, opened 2020-11-13
Fixes

michalvasko:


src/diff.c
```diff
@@ -82,6 +82,21 @@ lyd_diff_add(const struct lyd_node *node, enum lyd_diff_op op, const char *orig_
 
     assert(diff);
 
+    /* replace leaf always needs orig-default and orig-value */
```
rkrejci, line 85: Next time, I don't think that adding asserts is a FEATURE, it's just debugging not even changing anything in the Release build.

rkrejci (approved):

src/diff.c
michalvasko: Okay but these asserts have somewhat extended functionality of documenting (by code) what metadata exactly are expected in which case, that is why using MAINTENANCE or some similar tag did not seems appropriate to me.

rkrejci: so then it's DOC, isn't it?

michalvasko: Not sure, can DOC be something non-comment?

rkrejci: can FEATURE be something that is not compiled? :)

michalvasko: Like I said, I do not know what exactly I was supposed to use because none of the tags seem to be suitable.

PR #1275, opened 2020-11-16
Fix error for printing json schema for augmented node

euikook:
Fix error missing `,` from printing json schema for augmented node.


rkrejci:
Thanks! Seems fine, but please target it into the devel branch. The master is a release branch getting changes only from devel.

euikook:
OK. I see.

PR #1280, opened 2020-11-17
libyang2: data BUGFIX check JSON BMP character length

jvijtiuk:
Hello,

this PR adds a fix for an OOB read when parsing basic multilingual plane characthers in JSON
data. Previously it wasn't checked whether there were actually four
characters available after the "\u" sequence, and if there weren't an
OOB read would occur.

This PR also adds a fuzz regression test case, as the issue was found by fuzzing
lyd_parse_mem_json. The example input that caused a crash is shown below:

```
"\u1
```

The input didn't cause any crashes, only an invalid read, as shown by the valgrind output below:

```
==1908== Invalid read of size 1
==1908==    at 0x122882: lyjson_string_ (json.c:223)
==1908==    by 0x122D35: lyjson_string (json.c:314)
==1908==    by 0x123AAC: lyjson_value (json.c:585)
==1908==    by 0x123D56: lyjson_ctx_new (json.c:635)
==1908==    by 0x137A1B: lyd_parse_json_init (parser_json.c:1404)
==1908==    by 0x137B00: lyd_parse_json_data (parser_json.c:1425)
==1908==    by 0x1252CD: lyd_parse_data (tree_data.c:311)
==1908==    by 0x12559A: lyd_parse_data_mem (tree_data.c:376)
==1908==    by 0x1117DF: LLVMFuzzerTestOneInput (lyd_parse_mem_json.c:77)
==1908==    by 0x111885: main (main.c:39)
==1908==  Address 0x4eb6fc5 is 0 bytes after a block of size 5 alloc'd
==1908==    at 0x483A77F: malloc (vg_replace_malloc.c:307)
==1908==    by 0x111781: LLVMFuzzerTestOneInput (lyd_parse_mem_json.c:70)
==1908==    by 0x111885: main (main.c:39)
```

If any changes for the fix are needed, please let me know.

Thanks,
Juraj


rkrejci:
Thanks! I have slightly rewritten the patch. and merged manually. However, I was not able to reproduce the issue (valgrind complaining). There is definitely a bug and needs to fix, but I'm not sure that the regression tests will work correctly, In my case, they were passing even without the fix.

PR #1286, opened 2020-11-18
test REFACTOR put repetitive commands into macros

radek-isa:
Test refactorization. Create macros for variable checking to delete repetitive commands.


michalvasko (approved):
Seems nice to me, so it can be merged. But I have to write a test using the macros myself to be certain, it does not matter now, though.

rkrejci:
It is still work in progress, I've just went through the basic tests.

michalvasko:
Okay, I can take another look once you make it into an actual PR.

michalvasko (approved):

PR #1287, opened 2020-11-19
yanglint refactoring

rkrejci:
Rewrite yanglint(1) for the libyang 2.0 API.


src/log.h
```diff
@@ -84,6 +84,7 @@ typedef uint8_t ly_bool;
  */
 typedef enum
 {
+    LY_LLCUR = -1, /**< Special value for ::ly_log_level() which does not change the verbosity level and make it to return the current level. */
```
michalvasko, line 87: Is this really a good idea? I think some other API would be better, even if a new getter would be defined.

tools/lint/cmd_add.c
```diff
@@ -0,0 +1,166 @@
...
+            "  -i, --makeimplemented\n"
+            "                  Make the imported modules \"referenced\" from any loaded\n"
+            "                  <schema> module also implemented. If specified a second time,\n"
+            "                  all the modules are set implemented.\n");
+}
+
+void
+cmd_add(struct ly_ctx **ctx, const char *cmdline)
+{
+
```
michalvasko, line 52: Extra space.

tools/lint/cmd_data.c
```diff
@@ -0,0 +1,384 @@
...
+
+cleanup:
+    ly_out_free(out, NULL, 0);
+    ly_set_erase(&inputs, free_cmdline_file);
+    ly_set_erase(&requests, free_cmdline_file);
+    ly_set_erase(&request_paths, NULL);
+    ly_set_erase(&xpaths, NULL);
+    free_cmdline_file(operational);
+    free_cmdline(argv);
+
```
michalvasko, line 383: Extra line.

tools/lint/common.c
```diff
@@ -0,0 +1,586 @@
...
+            return 3;
+        }
+        (*schema) = informat_s;
+    }
+
+    return 0;
+}
+
+int
+print_list(struct ly_out *out, struct ly_ctx *ctx, LYD_FORMAT outformat)
```
michalvasko, line 339: Functions such as this one are called only once in the specific command so why not make the function static and put it there? I would find that better organized.

tools/lint/main_ni.c
```diff
@@ -196,933 +288,519 @@ libyang_verbclb(LY_LOG_LEVEL level, const char *msg, const char *path)
...
+            }
+        }
     }
 
-    if (informat_s) {
-        return 1;
-    } else {
-        return 2;
+    if (request_expected) {
+
```
michalvasko, line 389: Probably an extra line.

michalvasko (changes requested):

tools/lint/cmd_add.c
rkrejci: fixed

tools/lint/cmd_data.c
rkrejci: fixed

tools/lint/common.c
rkrejci: It's called twice - typically these functions are called from `cmd_*` file and from `main.ni.c` - libyang has 2 modes, interactive and non-interactive. I hope that whenever possible, I have unified the code to avoid duplication. What must be solved individually is command line processing, but executing command is in many cases common.

tools/lint/main_ni.c
rkrejci: fixed

src/log.h
rkrejci: Well, the cons of the getter are more functions (setter and getter) and changing API. So far, `ly_log_level()` served as both getter and setter, the problem was that the getter was providing only the previous value and it was not possible to not change the value. I believe that an extra value with no effect is quite an elegant way to fill this gap.

What exactly you see as cons of this change?

tools/lint/common.c
michalvasko: Ah, okay, did not realize that. Then it makes sense.

src/log.h
michalvasko: Having a bogus `LY_LOG_LEVEL` value, which is technically not a log level. I would prefer not to have it (because one must always be aware when working with this enum that there is this special value) even if it meant changing the API. Maybe simply 2 functions, one getter, one setter (having `void` return)?

rkrejci: we have agreed to unify all the logger's control functions, so `LY_LLCUR` will not be added, `ly_log_dbg_groups()` will return the previous value and in all cases, it will be necessary to call the function twice to get the current value without changing it (actually by resetting it). This behavior follows the `umask()`.

src/log.h
```diff
@@ -141,20 +146,21 @@ uint32_t ly_log_options(uint32_t opts);
  */
 
 #define LY_LDGDICT  0x01 /**< Dictionary additions and deletions. */
-#define LY_LDGYANG  0x02 /**< YANG parser messages. */
-#define LY_LDGYIN   0x04 /**< YIN parser messages. */
 #define LY_LDGXPATH 0x08 /**< XPath parsing end evaluation. */
```
michalvasko, line 149: I think it should be ordered, no reason to skip the values (except for backwards compatibility maybe, but one definitely needs to recompile an app when migrating from libyang 1 to 2).

michalvasko (changes requested):

src/log.h
rkrejci: fixed

michalvasko (approved):

PR #1289, opened 2020-11-24
build BUGFIX gcc miss paremeters for compiling program when coverage option is enabled

radek-isa:
In branch libyang2 is a bug. Command cmake -DENABLE_COVER=ON doesn't add parameters " --coverage -fprofile-arcs -ftest-coverage" into a variable CMAKE_C_FLAGS. These parameters are required for generating coverage.


rkrejci (approved):

rkrejci:
Seems that besides the CMakeLists.txt, there should be also a change in `.travis.yml` to use the cmake option to build coverage reports instead of exporting CFLAGS options directly as environment variable. Could you please look at that, @radek-isa ?

rkrejci (approved):

PR #1291, opened 2020-11-26
schema compile BUGFIX fix owerflow int32_t when LY_TYPE_ENUM compiling

radek-isa:


src/schema_compile_node.c
```diff
@@ -1185,35 +1185,35 @@ lys_compile_type_enums(struct lysc_ctx *ctx, struct lysp_type_enum *enums_p, LY_
...
-                if (!u || (cur_val >= value)) {
-                    value = cur_val + 1;
-                }
             } else {
                 /* assign value automatically */
-                if (u && (value == INT32_MIN)) {
-                    /* counter overflow */
+                if (u == 0) {
+                    cur_val = 0;
+                }
```
michalvasko, line 1197: This block seems redundant to me but some refactorization of the `if` block is needed so that the `0` is kept (or do not initialize `cur_val`).

michalvasko (changes requested):
Seems fine but please run `make format` to fix some formatting problems and perform a similar change for bits positions (similar code below this one). If you could also reduce code duplication for enum values and bits positions, it would be great.

src/json.c
```diff
@@ -221,7 +221,7 @@ lyjson_string_(struct lyjson_ctx *jsonctx)
                 for (value = i = 0; i < 4; i++) {
                     if (!in[offset + i]) {
                         LOGVAL(jsonctx->ctx, LY_VLOG_LINE, &jsonctx->line, LYVE_SYNTAX,
-                            "Invalid basic multilingual plane character \"%s\".", &in[slash]);
+                                "Invalid basic multilingual plane character \"%s\".", &in[slash]);
```
rkrejci, line 224: please move this into a standalone commit

src/schema_compile_node.c
```diff
@@ -1154,7 +1154,7 @@ lys_compile_type_enums(struct lysc_ctx *ctx, struct lysp_type_enum *enums_p, LY_
 {
     LY_ERR ret = LY_SUCCESS;
     LY_ARRAY_COUNT_TYPE u, v, match = 0;
-    int32_t value = 0, cur_val = 0;
+    int32_t  highest_value = INT32_MIN, cur_val;
```
rkrejci, line 1157: too many spaces

src/schema_compile_node.c
```diff
@@ -1185,10 +1185,8 @@ lys_compile_type_enums(struct lysc_ctx *ctx, struct lysp_type_enum *enums_p, LY_
 
         if (basetype == LY_TYPE_ENUM) {
             if (enums_p[u].flags & LYS_SET_VALUE) {
+                /* value assigned by model */
                 cur_val = (int32_t)enums_p[u].value;
```
rkrejci, line 1189: I'm afraid that here we need to check for collision with the previously (auto-)assigned values.

rkrejci (changes requested):

rkrejci:
I would appreciate extending tests to cover the assigned values/positions - as @michalvasko wrote, there should be a similar fix for bits type

src/schema_compile_node.c
rkrejci: The wrong sequence can be e.g.: 1, 2, - (auto-assigned 3), 3

radek-isa: Hi. The check for collision is on line 1191.

rkrejci: ok, sorry, I didn't notice

michalvasko (approved):

rkrejci (approved):

PR #1298, opened 2020-12-01
libyang2 fixes

rkrejci:
various fixes and changes

### DIFF
--- a/src/parser_yang.c
+++ b/src/parser_yang.c
@@ -104,8 +104,9 @@ LY_ERR parse_grouping(struct lys_yang_parser_ctx *ctx, struct ly_in *in, struct 
 LY_ERR
 buf_add_char(struct ly_ctx *ctx, struct ly_in *in, size_t len, char **buf, size_t *buf_len, size_t *buf_used)
 {
+#define BUF_STEP 16;
     if (*buf_len <= (*buf_used) + len) {
-        *buf_len += 16;
+        *buf_len += BUF_STEP;
         *buf = ly_realloc(*buf, *buf_len);
         LY_CHECK_ERR_RET(!*buf, LOGMEM(ctx), LY_EMEM);
     }
@@ -117,6 +118,7 @@ buf_add_char(struct ly_ctx *ctx, struct ly_in *in, size_t len, char **buf, size_
 
     (*buf_used) += len;
     return LY_SUCCESS;
+#undef BUF_STEP
 }
 
 /**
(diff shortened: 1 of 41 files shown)


src/parser_yang.c
```diff
@@ -1987,14 +2007,14 @@ parse_type_pattern(struct lys_yang_parser_ctx *ctx, struct ly_in *in, struct lys
 
     /* add special meaning first byte */
     if (buf) {
-        buf = realloc(buf, word_len + 2);
+        buf = ly_realloc(buf,  word_len + 2);
```
michalvasko, line 2010: Extra space.

src/xpath.h
```diff
@@ -223,6 +223,13 @@ struct lyxp_set {
         struct lyxp_set_scnode {
             struct lysc_node *scnode;
             enum lyxp_node_type type;
+
+#define LYXP_SET_SCNODE_TRAVERSED_IN            -2
+#define LYXP_SET_SCNODE_TRAVERSED_NOSTART_NOTIN -1
+#define LYXP_SET_SCNODE_TRAVERSED_NOTIN          0
+#define LYXP_SET_SCNODE_IN                       1
+#define LYXP_SET_SCNODE_IN_JUST_ADDED            2
+#define LYXP_SET_SCNODE_NOTIN                    3
```
michalvasko, line 232: I am not sure about these, I have nothing against using named constants instaed of numbers so that the code is self-explaining but in this case it is not that simple. Value `3` and any higher ones are all used for predicates, there is not just one. So, it may be better to just leave it as it is.

michalvasko (changes requested):

src/parser_yang.c
rkrejci: fixed

src/parser_yang.c
```diff
@@ -293,9 +301,14 @@ static LY_ERR
...
-    /* string: 0 - string ended, 1 - string with ', 2 - string with ", 3 - string with " with last character \,
-     *         4 - string finished, now skipping whitespaces looking for +,
-     *         5 - string continues after +, skipping whitespaces */
+    /* string parsing status: */
+#define STRING_ENDED 0 /* string ended */
+#define STRING_SINGLE_QUOTED 1 /* string with ' */
+#define STRING_DOUBLE_QUOTED 2 /* string with " */
+#define STRING_DOUBLE_QUOTED_ESCAPED 3 /* string with " with last character \ */
+#define STRING_PAUSED                4 /* string finished, now skipping whitespaces looking for + */
+#define STRING_PAUSED_NEXTLINE       5 /* string continues after +, skipping whitespaces */
```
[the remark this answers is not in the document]
rkrejci: sure, no problem

src/xpath.h
rkrejci: I've tried to improve the description here. But I don't see any problem with using `LYXP_SET_SCNODE_NOTIN` as you describe and as previously the value `3` was used. Any higher value was assigned by incrementing the current (or some other) value. What I did in the code was just replacing numbers with the named constants.

michalvasko: Okay, then I wanted to say that not all the uses of these values were refactored, now they seem to be (I have not searched for them myself assuming you did it). Then I will only suggest better names for the values.
`LYXP_SET_SCNODE_TRAVERSED_IN` - `LYXP_SET_SCNODE_START`
`LYXP_SET_SCNODE_TRAVERSED_NOSTART_NOTIN` - `LYXP_SET_SCNODE_START_USED`
`LYXP_SET_SCNODE_TRAVERSED_NOTIN` - `LYXP_SET_SCNODE_ATOM`
`LYXP_SET_SCNODE_IN` - `LYXP_SET_SCNODE_ATOM_CTX`
`LYXP_SET_SCNODE_IN_JUST_ADDED` - `LYXP_SET_SCNODE_ATOM_NEW_CTX`
`LYXP_SET_SCNODE_NOTIN` - `LYXP_SET_SCNODE_ATOM_PRED_CTX`
They are not perfect but with the commen it should be fairly clear.

rkrejci: yes, first I forgot to go through all uses of the `in_ctx` member, but it should be fixed since the previous update.

ok, I'm going to change the names.

rkrejci: fixed

michalvasko (approved):

PR #1301, opened 2020-12-02
Prevent debug buffer overflow

tmtom:
Originally there is no check for debug buffer overrun. Running with debug/verbose (-V -V -V) causes segfault on long debug logs (e.g. printing long pattern).
This increases the buffer and also prevents overruns.

### DIFF
--- a/src/xpath.c
+++ b/src/xpath.c
@@ -151,6 +151,17 @@ print_token(enum lyxp_token tok)
     }
 }
 
+#define DBG_BUFF_SIZE 8192
+
+void
+dbg_sprintf_append(char *buff, size_t buff_size, const char *fmt, ...)
+{
+    va_list argptr;
+    va_start(argptr,fmt);
+    size_t offset = strlen(buff);
+    vsnprintf(buff + offset, buff_size - offset - 1, fmt, argptr);
+}
+
 /**
  * @brief Print the whole expression \p exp to debug output.
  *
@@ -160,7 +171,7 @@ static void
 print_expr_struct_debug(struct lyxp_expr *exp)
 {
     uint16_t i, j;
-    char tmp[128];
+    char tmp[DBG_BUFF_SIZE];
 
     if (!exp || (ly_log_level < LY_LLDBG)) {
         return;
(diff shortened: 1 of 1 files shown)


src/xpath.c
```diff
@@ -160,22 +162,44 @@ static void
 print_expr_struct_debug(struct lyxp_expr *exp)
 {
     uint16_t i, j;
-    char tmp[128];
+    char tmp[DBG_BUFF_SIZE];
+    char *buff = tmp;
+    int buff_left = DBG_BUFF_SIZE - 1;
```
michalvasko, line 167: `buff_left` is redundant, it is always `DBG_BUFF_SIZE - 1 - buff_used`, right? But okay, I suppose, not a bug deal.

src/xpath.c
```diff
@@ -160,22 +162,44 @@ static void
...
     LOGDBG(LY_LDGXPATH, "expression \"%s\":", exp->expr);
     for (i = 0; i < exp->used; ++i) {
-        sprintf(tmp, "\ttoken %s, in expression \"%.*s\"", print_token(exp->tokens[i]), exp->tok_len[i],
+        buff_used = snprintf(buff, buff_left, "\ttoken %s, in expression \"%.*s\"", print_token(exp->tokens[i]), exp->tok_len[i],
                &exp->expr[exp->expr_pos[i]]);
+        if (buff_used > 0) {
+            buff_used = buff_used>buff_left ? buff_left : buff_used;
+            buff += buff_used;
+            buff_left -= buff_used;
+        }
```
michalvasko, line 182: Put this into a function or macro instead of duplicating it 3 times.

michalvasko (changes requested):

src/xpath.c
```diff
@@ -151,6 +151,16 @@ print_token(enum lyxp_token tok)
...
 }
 
+#define DBG_BUFF_SIZE 8192
+
+void dbg_sprintf_append(char *buff, size_t buff_size, char *fmt, ...)
+{
+    va_list argptr;
+    va_start(argptr,fmt);
+    size_t offset = strlen(buff);
+    snprintf(buff+offset, buff_size-offset-1, fmt, argptr);
```
michalvasko, line 161: Did you test this? I think you wanted to use `vsnprintf(3)` instead. Also, put spaces around `+` and `-`.

src/xpath.c
```diff
@@ -151,6 +151,16 @@ print_token(enum lyxp_token tok)
     }
 }
 
+#define DBG_BUFF_SIZE 8192
+
+void dbg_sprintf_append(char *buff, size_t buff_size, char *fmt, ...)
```
michalvasko, line 156: Function declaration should on a separate line, after the return type.

michalvasko (changes requested):

src/xpath.c
```diff
@@ -151,6 +151,17 @@ print_token(enum lyxp_token tok)
...
 
+#define DBG_BUFF_SIZE 8192
+
+void
+dbg_sprintf_append(char *buff, size_t buff_size, char *fmt, ...)
+{
+    va_list argptr;
+    va_start(argptr,fmt);
+    size_t offset = strlen(buff);
+    vsnprintf(buff+offset, buff_size - offset - 1, fmt, argptr);
```
michalvasko, line 162: Missing spaces around `+`.

michalvasko (changes requested):

src/xpath.c
```diff
@@ -151,6 +151,17 @@ print_token(enum lyxp_token tok)
     }
 }
 
+#define DBG_BUFF_SIZE 8192
+
+void
+dbg_sprintf_append(char *buff, size_t buff_size, char *fmt, ...)
```
michalvasko, line 157: Sorry, just now noticed that `char *fmt` should be `const char *fmt`.

michalvasko (changes requested):

michalvasko (approved):

PR #1310, opened 2020-12-09
Libyang2 fixes

rkrejci:


src/common.h
```diff
@@ -556,19 +556,20 @@ LY_ERR ly_strcat(char **dest, const char *format, ...);
...
  * @param[in] CTX libyang context for logging.
  * @param[in,out] ARRAY Pointer to the array to create.
- * @param[in] SIZE Number of items the array is supposed to hold. The size of the allocated
+ * @param[in] SIZE Number of the new items the array is supposed to hold. The size of the allocated
  * space is then counted from the type of the ARRAY, so do not provide placeholder void pointers.
+ * @param[in] ERR Additional action(s) in case of error (passed to LY_CHECK_ERR_RET).
  * @param[in] RETVAL Return value for the case of error (memory allocation failure).
  */
-#define LY_ARRAY_CREATE_RET(CTX, ARRAY, SIZE, RETVAL) \
+#define LY_ARRAY_CREATE_ERR_RET(CTX, ARRAY, SIZE, ERR, RETVAL) \
```
michalvasko, line 564: I do not really like `ERR` (because the error is logged so this is used for cleanup actions) or `RETVAL` (because it should almost always be `LY_EMEM`) parameters. I like having 2 variants, which would in this case be `LY_ARRAY_CREATE_RET` and `LY_ARRAY_CREATE_GOTO`. The former one is the simple version used in most cases without redundant parameters and the latter can be used in all other cases.

rkrejci: ok, it needed some code refactorization because the previous RESIZE variant of the macro was accepting ERR cleanup and the code used it. Now it should be fixed

michalvasko (approved):

PR #1311, opened 2020-12-09
Fixes

michalvasko:


src/tree_schema.c
```diff
@@ -381,7 +381,7 @@ lys_find_expr_atoms(const struct lysc_node *ctx_node, const struct lys_module *c
     (*set)->size = xp_set.used;
 
     for (i = 0; i < xp_set.used; ++i) {
-        if ((xp_set.val.scnodes[i].type == LYXP_NODE_ELEM) && (xp_set.val.scnodes[i].in_ctx == LYXP_SET_SCNODE_ATOM_CTX)) {
+        if ((xp_set.val.scnodes[i].type == LYXP_NODE_ELEM) && (xp_set.val.scnodes[i].in_ctx > -1)) {
```
rkrejci, line 384: Please use the named value.

rkrejci (changes requested):

src/tree_schema.c
michalvasko: Hm, I have automatically used a number because it seemed that using the named value would be more confusing since an interval of values is checked here, not a single one.

rkrejci: which means that the values are somehow ordered. You can describe how that are ordered at their definition and using `>` won't be confusing.

michalvasko: Fixed.

PR #1320, opened 2020-12-24
Re-enable LY_ENABLED_LYD_PRIV functionality

### LINKED ISSUE #1319: ly2: bring back priv in libyang2
The CMAKE define ENABLE_LYD_PRIV functionality has been commented out as well as the defining of LY_ENABLED_LYD_PRIV in libyang.h.in. Please restore this functionality, as it is used.

```
if(ENABLE_LYD_PRIV)
    set(LY_ENABLED_LYD_PRIV 1)
endif()
```

choppsv1:
fixes #1319


michalvasko:
Right, we wanted to have it always enabled for *libyang2* but it seems we forgot about it at some point. Should be fine now.

PR #1329, opened 2021-01-06
coverity fixes

rkrejci:


src/log.c
```diff
@@ -245,8 +245,14 @@ void
 ly_log_location(const struct ly_ctx *ctx, const struct lysc_node *scnode, const struct lyd_node *dnode,
         const char *path, const struct ly_in *in, uint64_t line, ly_bool reset)
 {
-    struct ly_log_location_s *loc = pthread_getspecific(ctx->log_location_key);
+    struct ly_log_location_s *loc;
 
+    if (!ctx) {
```
michalvasko, line 250: So it is a valid situation, if `ctx` is `NULL`? If I see this function being called, I would always expect the parameters to be used for logging.

src/log.c
```diff
@@ -290,9 +296,14 @@ void
 ly_log_location_revert(const struct ly_ctx *ctx, uint32_t scnode_steps, uint32_t dnode_steps,
         uint32_t path_steps, uint32_t in_steps)
 {
+    struct ly_log_location_s *loc;
 
-    struct ly_log_location_s *loc = pthread_getspecific(ctx->log_location_key);
+    if (!ctx) {
```
michalvasko, line 301: Same as previous comment.

src/log.c
rkrejci: The logging location is connected with the context, so if the context is not available (it is not necessarily mandatory at all places), I believe that we need this location just do nothing. Eventually, the check for ctx can be moved into the macro we actually use, but I don't see any advantage in it.

The NULL context is used also to ignore the reverting location - the provided context is set only when the location set is done, so there is no need to maintain a counter for set arguments.

michalvasko: I would prefer to have something like
```C
if (ctx) {
  ly_log_location(ctx, ...);
}
```
than for this logic to be hidden in the function. If there are even use-cases for this, I suppose there are but I do not really like it and would expect that at least on the level of functions, every one of them either always has the context or never has.

rkrejci: I'm not going to check every call of the location functions to get know if it is the case. I know about one reported by coverity, but it doesn't mean it is the only one. I can put the condition into the macro wrapping the location function, but I don't see added value in it.

I see the usage similar to `free()`, it also silently ignores NULL pointers so you don't have to wrap it by `if` conditions.

michalvasko: The added value is that in the function I always know what information will be used for logging. Is it really okay for a function to use whatever logging information is already set (context is not available) in one case and set some additional information (context is available) in another? It is true that in the former case no logging information should be used at all since the context is not available but is that really true for all the cases?

That leads to the question why is the context needed for this at all, ideally it should only determine whether the message will be stored in the context or not. For that the simple `THREAD_LOCAL` macro could be used instead of bothering with pthread keys.

Also, I believe it should be clear in every function whether the context is available or not, I think that is quite important information in general. I would not mind using asserts to make sure of this, but simply make sure that eventually this information is correctly present in every function.

rkrejci: You already don't know, directly in the function, what information will be used for logging - the location data can be set in some of the previous callers down in the stack.

It is connected with the context because the error records are connected with the context. It seems dangerous to me to have the location storage common for all the contexts when it is being connected (inserted) into the records which are actually context-specific. Probably it could work - I'm not aware of the code causing problems now, but that doesn't mean it cannot be simply added. For example moving data between 2 contexts, which is the functionality in libyang 1.x, can be such a thing.

So far there was no such a strict requirement for checking context presence in all cases. There are functions using context just for logging and not all the use cases require or even provide the context.

I could fix just the one case reported by coverity and check the context or count the set location. But it is vulnerable to keep the location functions without checking for context and handling it somehow. I don't get assert() for this, because believing that all the cases will be covered by tests and the issues will appear in debug mode is just an illusion.

michalvasko: > You already don't know, directly in the function, what information will be used for logging - the location data can be set in some of the previous callers down in the stack.

Okay, what I meant is that you should know whether the logging information will be used as it is (not modified because of missing context) or you will change it (context available).

> It is connected with the context because the error records are connected with the context.

Right, my idea was not to connect the error information with the context, just with the thread. Every API function should start and end with empty logging information (if not, it is a bug) so I do not see a problem with this and it would solve the problem we are discussing.

> So far there was no such a strict requirement for checking context presence in all cases. There are functions using context just for logging and not all the use cases require or even provide the context.

I would simply expect to always be able to determine whether the context is available or not based on the parameters (what else), which should actually be already possible even though you seems to disagree. As long as there is a `ly_ctx` or `lysc_node`/`lyd_node` parameter, the context must be available. If some of the parameters are optional, you definitely must know that in a function.

rkrejci: > > You already don't know, directly in the function, what information will be used for logging - the location data can be set in some of the previous callers down in the stack.
> 
> Okay, what I meant is that you should know whether the logging information will be used as it is (not modified because of missing context) or you will change it (context available).
> 

but it is not always that important - log messages are also dependant on several variables in the background - logging level, callback function and the context (for storing records). The important thing here is to have the same context variable for the related location functions (set and reset).

> > It is connected with the context because the error records are connected with the context.
> 
> Right, my idea was not to connect the error information with the context, just with the thread. Every API function should start and end with empty logging information (if not, it is a bug) so I do not see a problem with this and it would solve the problem we are discussing.
> 

agree that the location information should be always empty outside libyang. And as I wrote, I believe that for now, it is ok. But I also pointed out that there were functions in libyang 1.x working with multiple contexts and I see a possibility to have similar functions even in libyang 2.x in the future. And then it could be a problem, so I believe that having it connected with the context is better than having it just thread-wide.

> > So far there was no such a strict requirement for checking context presence in all cases. There are functions using context just for logging and not all the use cases require or even provide the context.
> 
> I would simply expect to always be able to determine whether the context is available or not based on the parameters (what else), which should actually be already possible even though you seems to disagree. As long as there is a `ly_ctx` or `lysc_node`/`lyd_node` parameter, the context must be available. If some of the parameters are optional, you definitely must know that in a function.

I agree, I just don't want to force using the following construct in all that functions when it can be inside the function itself. 
```c
if (ctx) {
  LOG_LOCSET(ctx, ...);
}
```
You are able to get know if a variable is NULL before you call `free()`, but why would you have to?

michalvasko: > log messages are also dependant on several variables in the background - logging level, callback function and the context

What I mean for "logging information" is line number/schema node/data node/path. None of the things you mention affect the message itself, only how/where it will be printed/stored.

> ... And then it could be a problem

Why? If the logging information is not connected to a context, how can several contexts cause problems?

> You are able to get know if a variable is NULL before you call free(), but why would you have to?

I do not really consider this comparison relevant. The difference is that after you call free on a variable, it is always freed (whether the variable was `NULL` or properly allocated before) so it does not affect the code that follow. In this case it does affect the following code because the logging information may or may not be changed, I believe it should always be known if we want to have that information accurate.

rkrejci: There is a possibility to have the data from both contexts in different places when processing multiple contexts. But ok, this seems like a hypothetical situation. And moving the location information from the context would solve both points here. I'll look at how it is possible to handle freeing data of the `THREAD_LOCAL` variables and will try to use it.

michalvasko (approved):

Issue #317, opened 2017-05-04
Feature request: report multiple errors from yanglint (and NETCONF)


jktjkt:
A TL;DR version: can you please report all ``must`` failures when validating data? It seems that only the first one of them is reported by ``yanglint`` and I presume also via NETCONF -- at least for list nodes.

Here's my toy model which attempts to validate list nodes for mutual conflicts:

```yang
module test-jkt-1 {
  namespace "http://example.org/test/jkt/1";
  prefix "tj1";

  container seq {

    list item {
      ordered-by "user";
      key "center";
      must "count(/seq/item[(center > current()/center) and (center - (width div 2) < current()/center + current()/width div 2)]) = 0" {
        error-message "Another item with a higher center overlaps with this one";
      }
      /*must "count(/seq/item[(center < current()/center) and (center + (width div 2) > current()/center - current()/width div 2)]) = 0" {
        error-message "Another item with a lower center overlaps with this one";
      }*/
      leaf center {
        type uint8;
        mandatory true;
      }
      leaf width {
        type uint8;
        mandatory true;
      }
    }

  }
}
```
```xml
<?xml version='1.0' encoding='UTF-8'?>
<seq xmlns="http://example.org/test/jkt/1">
    <item>
      <center>10</center>
      <width>2</width>
    </item>
    <item>
      <center>12</center>
      <width>2</width>
    </item>
    <item>
      <center>13</center>
      <width>1</width>
    </item>
    <item>
      <center>14</center>
      <width>1</width>
    </item>
</seq>
```

The check is implemented at the level of each individual list item so that e.g. ``yanglint`` can point to an item which doesn't conform to the validation rules. However, from the application domain's point of view, no single item is at fault here because it takes at least a pair to have a conflict due to an overlap within that pair.

The existing ``must`` checks just the items that are "higher" than the current one. I would like to add a complementary rule which checks the "lower" items for an overlap. When I uncomment the other ``must``, though, I'm still getting just the first error.

PR #1337, opened 2021-01-08
feature: support multiple validation errors

fanchanghu:
Hi,
I need multiple validation errors feature in my project, and I found an issue in libyang #317, so I have try to implement this feature myself, could you please check this PR and merge it to libyang?

thanks.


fanchanghu:
Hi,
the building for ARM failed with log like `ssh: connect to host 192.168.121.199 port 202: Connection refused`, it seems it has nothing to do with my PR.

michalvasko:
Okay, this feature was forgotten and can be easily added. However, the flag seems redundant and it would be much simpler without it, so please change it that way, so that always all unsatisfied `must` expression errors are printed. Also, change the branch to `devel` for now, it should not be long to the new release.

PR #1341, opened 2021-01-09
feature: support multiple validation errors to devel branch

fanchanghu:
Hi,
This is the same PR with #1337 to devel branch.

I also removed the option `LYD_OPT_MULTI_ERRORS` according to your request, please check it, thanks.

### DIFF
--- a/src/resolve.c
+++ b/src/resolve.c
@@ -6219,17 +6219,20 @@ resolve_list_keys(struct lys_node_list *list, const char *keys_str)
  *
  * @param[in] node Data node with optional must statements.
  * @param[in] inout_parent If set, must in input or output parent of node->schema will be resolved.
+ * @param[in] ignore_fail 0 - no, 1 - yes, 2 - yes, but only for external dependencies.
+ * @param[in] multi_error 0 - no, 1 - yes.
  *
  * @return EXIT_SUCCESS on pass, EXIT_FAILURE on fail, -1 on error.
  */
 static int
-resolve_must(struct lyd_node *node, int inout_parent, int ignore_fail)
+resolve_must(struct lyd_node *node, int inout_parent, int ignore_fail, int multi_error)
 {
     uint8_t i, must_size;
     struct lys_node *schema;
     struct lys_restr *must;
     struct lyxp_set set;
     struct ly_ctx *ctx = node->schema->module->ctx;
+    int rc = EXIT_SUCCESS;
 
     assert(node);
     memset(&set, 0, sizeof set);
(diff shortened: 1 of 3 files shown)


michalvasko:
Thanks, but the point of my request to make it all much simpler without useless arguments. Ignore whether all the errors can be stored, they will all be printed in any case. I was hoping you can simply change the `resolve_must` function and that will be it, no new parameters anywhere.

fanchanghu:
Hi,
I think it's better to use the log option, because:

1. If the log option `LY_LOSTORE_LAST` is set, the user can only get one error even if multiple errors are detected. 
2. Detecting multiple errors is inefficient when the user just wants to return from the first error. 
3. I think you may want to keep the api compatible for libyang1, Detecting multiple errors as default may breakdown some users. (return the last error instand of the first one.)

michalvasko:
Okay, we have discussed this change properly and have decided that the new option is the right solution after all, sorry about that. Also, please try to keep the code reasonably clean. For example, use 2 return values in `resolve_data_unres` instead of performing `|` on a single one. You can make these changes in this PR, even force-pushing a completely new commit is fine.

src/resolve.c
```diff
@@ -8622,18 +8634,21 @@ resolve_unres_data(struct ly_ctx *ctx, struct unres_data *unres, struct lyd_node
         }
         assert(!(options & LYD_OPT_TRUSTED) || ((unres->type[i] != UNRES_MUST) && (unres->type[i] != UNRES_MUST_INOUT)));
 
-        rc = resolve_unres_data_item(unres->node[i], unres->type[i], ignore_fail, NULL);
-        if (rc) {
-            /* since when was already resolved, a forward reference is an error */
-            return -1;
+        rc2 = resolve_unres_data_item(unres->node[i], unres->type[i], ignore_fail, multi_error, NULL);
+        if (rc2) {
+            if(!multi_error) {
```
michalvasko, line 8639: Space after `if`.

src/resolve.c
```diff
@@ -6303,12 +6306,15 @@ resolve_must(struct lyd_node *node, int inout_parent, int ignore_fail)
                 if (must[i].eapptag) {
                     ly_err_last_set_apptag(ctx, must[i].eapptag);
                 }
-                return 1;
+                rc = 1;
+                if(!multi_error) {
```
michalvasko, line 6310: Space after `if`.

michalvasko (changes requested):
Okay, except for minor formatting errors it seems fine.

michalvasko (approved):

Issue #1094, opened 2020-05-27
Performance issue while building data tree with operations create and destroy.


vishaldhingra:
[libyang_scale.zip](https://github.com/CESNET/libyang/files/4689235/libyang_scale.zip)
Hi,

We are currently facing performance issues on FRR where libyang is used for configuration processing.
I have noticed API lyxp_eval() consumes lot of cycles in lyd_find_path() in a scaled environment.
I have attached the test code, to reproduce this issue.

Our Requirement is to create a scaled data tree with given set of configurations
which have operations create and destroy.

Steps are done to reproduce the issue :
1. Created a context.
2. Loaded a module frr-scale.yang
3. Building the data tree via lyd_new_path/lyd_find_path APIs.

At present we tries to configure 10K routes in one commit.
In order to configure 1 route.
We have 4 X-Paths, 3 with operation NB_OP_CREATE and one with operation NB_OP_DESTROY.

Steps to use test code (It will create a data tree with 10K entries)
1. bin folder : make clean; make all
                It will give you executable libyang_scale
                run it with ./libyang_scale
    The yang module is also placed under this bin folder only name : frr-scale.yang
2. src folder : yang.c contains the code to create context, load module and create entries in data tree.
                ROUTE_COUNT Macro controls the number of route entries to create, At present it sets to 10K.

In code these xpaths are mapped to Macros:
Opertaion Create : XPATH_PREFIX, XPATH_PATH, XPATH_NEXTHOP
Operation delete : XPATH_MPLS

Libyang version : Latest master at commit ID : 773653e17440ef8d00fd1dbf9a06560b73f4361d

Observations : 
RUN 1: When i execute this binary, it would take around 4 min.
time ./libyang_scale 
log_time: Data Tree start = 05-27-2020  04:39:23.207444
log_time: Data Tree end = 05-27-2020  04:43:48.662590

real    4m25.469s
user    4m20.463s
sys     0m1.608s

RUN 2: When i execute this binary after commenting out the lyxp_eval() in lyd_find_path(), then it
       takes around .618 seconds
time ./libyang_scale 
log_time: Data Tree start = 05-27-2020  06:57:30.713136
log_time: Data Tree end = 05-27-2020  06:57:31.323144

real    0m0.618s
user    0m0.618s
sys     0m0.012s

Please help us, to resolve this issue. 
If any other information is required then let us know.

michalvasko:
Hi,
[...]

PR #1342, opened 2021-01-10
data tree: Added the API to find the dnode with simple xpath.

vishaldhingra:
This API is a wrapper over the lyd_find_sibling and takes the
input as simple xpath.

Issue #1094

Signed-off-by: vishaldhingra <vdhingra@vmware.com>

### DIFF
--- a/src/tree_data.h
+++ b/src/tree_data.h
@@ -1440,6 +1440,23 @@ int lyd_lyb_data_length(const char *data);
  */
 void *lyd_set_private(const struct lyd_node *node, void *priv);
 
+/**
+ * @brief Get the data node based on a simple XPath.
+ * This API returns the closest parent of the node (or the node itself)
+ * identified by the nodeid (path). This API uses lyd_find_sibling() If cache is
+ * enabled and the siblings are NOT top-level nodes, this function finds the
+ * node in a constant time.
+ * @param[in] data_tree Existing data tree to get.
+ * @param[in] ctx Context to use.
+ * @param[in] path Simple data path (see @ref howtoxpath).
+ * @param[in] output possible values are 1 and 0. output value maps to pathoption.
+ * Only valid pathoption in this API context is LYD_PATH_OPT_OUTPUT.
+ * Pass 1 if pathotion LYD_PATH_OPT_OUTPUT to use, else pass 0.
+ * NULL If no lyd_node exists for the given xpath.
+ */
+struct lyd_node * lyd_find_path_hash_based(struct lyd_node *data_tree, const struct ly_ctx *ctx, const char *path,
+										   int output);
+
 #endif
 
 /**@} */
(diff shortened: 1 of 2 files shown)


vishaldhingra:
We have a requirement to do a node lookup based on inbuilt libyang's hash tables. At present there is no API which takes xpath as the Input and returns the dnode based on hash table.

So, I have written the new API.
Please have a look.

src/tree_data.h
```diff
@@ -1440,6 +1440,28 @@ int lyd_lyb_data_length(const char *data);
...
+ * This API returns the closest parent of the node (or the node itself)
+ * identified by the nodeid (path). This API uses lyd_find_sibling() If cache is
+ * enabled and the siblings are NOT top-level nodes, this function finds the
+ * node in a constant time.
+ * @param[in] data_tree Existing data tree to get.
+ * @param[in] ctx Context to use.
+ * @param[in] path Simple data path (see @ref howtoxpath). List nodes can have predicates, one for each list key
+ * in the correct order and with its value as well or using specific instance position, leaves and leaf-lists
+ * can have predicates too that have preference over \p value. When specifying an identityref value in a predicate,
+ * you MUST use the module name as the value prefix!
```
michalvasko, line 1454: Just copying the description from `lyd_new_path()` will not do because this is a subset of what `lyd_new_path()` is doing. List predicates are always mandatory and I would say it should also be true for leaf-lists. Leaves do not require value at all, I think, but please verify that. If so, then you can actually remove `value` and `value_type` parameters.

src/tree_data.h
```diff
@@ -1440,6 +1440,28 @@ int lyd_lyb_data_length(const char *data);
...
+ * @param[in] value Value of the new leaf/lealf-list (const char*). If creating anydata or anyxml, the following
+ * \p value_type parameter is required to be specified correctly. If creating nodes of other types, the
+ * parameter is ignored.
+ * @param[in] value_type Type of the provided \p value parameter in case of creating anydata or anyxml node.
+ * @param[in] options Bitmask of options flags, see @ref pathoptions.
```
michalvasko, line 1459: The only relevant option is `LYD_PATH_OPT_OUTPUT` which you should specify explicitly. Even more, you could change the options to a single flag `int output` and then transform it internally into the flag.

michalvasko (changes requested):
Okay, I did not even realize that this functionality is there but used for something quite different. So fine, after addressing the comments it can be merged. Additionally, please change the target branch to `devel` (there should be a release soon) and indentation to 4 spaces to be coherent with the rest of the code.

src/tree_data.h
vishaldhingra: done

src/tree_data.h
vishaldhingra: Yes, you are correct. I have removed value and value_type parameters.
Please check the description again, If i am missing anything please let me know.

michalvasko:
Duplicate of #1346.

PR #1347, opened 2021-01-14
libyang2: json BUGFIX invalid free in lyjson_ctx_set_value

jvijtiuk:
Hello,

this PR fixes an invalid free issue in lyjson_ctx_set_value.

The original issue was discovered by fuzzing. The file below caused an invalid free when I ran it on the lyd_parse_mem_json fuzz harness.

```
{"200 -11-10T23:00:00Z": "hello w\rld"},
```

The issue seems to be caused by the check of the dynamic variable, which represents whether the JSON value that is yet to be set has been dynamically allocated, and then the old jsonctx->value is freed. It looks like jsonctx->dynamic should be checked instead as the old json value is beeing freed after all.

The fuzzer input triggered the crash because the first string wouldn't be a dynamic one, since it had no special characters, and then the second one had a line break ("\r") which caused it to be stored in a dynamic string, which triggered the case where the previous string wasn't dynamic, and the current one is, so the previous one is freed.

Looking at the code, a similar case exists in lyjson_number, where JSON numbers that have exponents in them are allocated dynamically. That case can be reproduced wiith the following input:

```
{"200 -11-10T23:00:00Z": 1E+2}
```

I've added both of those input files to the corpus for fuzz regression tests.

Regards,
Juraj


rkrejci (approved):

rkrejci:
Thanks, you are right!

PR #1349, opened 2021-01-15
explicitly set tmp to empty string to avoid buffer overflow of strlen…

gwendlan:
… if tmp happens to contain non-zero garbage

### DIFF
--- a/src/xpath.c
+++ b/src/xpath.c
@@ -172,6 +172,7 @@ print_expr_struct_debug(struct lyxp_expr *exp)
 {
     uint16_t i, j;
     char tmp[DBG_BUFF_SIZE];
+    tmp[0] = 0;
 
     if (!exp || (ly_log_level < LY_LLDBG)) {
         return;


gwendlan:
The following test program illustrates the problem:
/*
  simulate non-zero garbage in uninitialised buffer
  compile: gcc main.c
  call:    ./a.out -> prints "useful output"
           ./a.out 1 -> no zero byte in tmp -> crashes with buffer overflow/ stack smashing detected
           ./a.out 2 -> zero byte in tmp but not at its beginning -> causes garbage to be printed followed by what was intended
*/


#define DBG_BUFF_SIZE 8192
  char tmp[DBG_BUFF_SIZE];

  if (argc > 2) {
    /* simulate non-zere garbage with a zero byte in the middle of tmp */
    snprintf(tmp, DBG_BUFF_SIZE, "garbage from unitialised variable: ");
  }
  else if (argc > 1) {
    /* simulate non-zero garbage for the whole array */
    for (int i=0; i < DBG_BUFF_SIZE; i++) {
      tmp[i] = 'a';
    };
  }
  else {
    /* initialise properly */
    tmp[0]=0;
  };
  dbg_sprintf_append(tmp, DBG_BUFF_SIZE, "useful output");

  fprintf(stderr, "%s\n", tmp);
}

gwendlan:
Also note that, if strlen returns 8200, the expression buff_size - offset -1 wraps around:

(gdb) p (size_t)8192 - (size_t) 8200 - 1
$3 = 18446744073709551607

rkrejci (approved):

rkrejci:
Thanks!

PR #1354, opened 2021-01-17
schema compile BUGFIX fix infinite loop with include

choppsv1:
Don't reuse index variable for inner walk of arrays.


rkrejci (approved):

rkrejci:
Thanks!

Issue #1345, opened 2021-01-14
Skip parsing extension


fanchanghu:
Hi,
We need skip some extensions in our modules for saving memory. Then I found that the return value `2` of `lyext_check_position_clb` could be suitable for this case as decribed in its comment. However, this feature does not be supported now. So I want to do it myself in libyang1, is it OK or any suggestion?

rkrejci:
Well, it's a bug that the value `2` is not handled as described, but I'm not sure that using it for your purpose will help you. It just avoids instantiating the extension in the module, the plugin is still present. Wouldn't be the solution just to avoid loading the plugin at all? Or is the intention to avoid instantiating the extension only in specific places/modules, so there will be some places where the extension is supposed to be instantiated?

fanchanghu:
What we need is just avoiding instantiating the extension in the module. We have some extensions which are used to describe the node, such as chinese description, short name and so on, these extensions are only used by the NETCONF client. So we hope avoiding instantiating them in the server.

rkrejci:
ok, then I have no problem with accepting the fix to correctly handle the return values of this callback

PR #1356, opened 2021-01-19
support plugin to skip parsing special extension instance.

fanchanghu:
Hi,
This PR is for the issue #1345 , however, it is a little more complex than I expected.

Please have a look and give some suggestions, thanks.  @rkrejci 

I will still work on this to add some tests if there is no major problems in this code.

### DIFF
--- a/src/parser.h
+++ b/src/parser.h
@@ -86,7 +86,7 @@ enum LY_IDENT {
 };
 int lyp_yin_fill_ext(void *parent, LYEXT_PAR parent_type, LYEXT_SUBSTMT substmt, uint8_t substmt_index,
                      struct lys_module *module, struct lyxml_elem *yin, struct lys_ext_instance ***ext,
-                     uint8_t ext_index, struct unres_schema *unres);
+                     uint8_t *ext_size, struct unres_schema *unres);
 
 int lyp_yin_parse_complex_ext(struct lys_module *mod, struct lys_ext_instance_complex *ext,
                               struct lyxml_elem *yin, struct unres_schema *unres);
@@ -161,6 +161,20 @@ const char *lyp_get_yang_data_template_name(const struct lyd_node *node);
 const struct lys_node *lyp_get_yang_data_template(const struct lys_module *module, const char *yang_data_name, int yang_data_name_len);
 
 void lyp_ext_instance_rm(struct ly_ctx *ctx, struct lys_ext_instance ***ext, uint8_t *size, uint8_t index);
+
+/**
+ * @brief Get extension instances from given parent node
+ *
+ * @param[in] ctx Context with the modules
+ * @param[in] elem Parent of the extension instances
+ * @param[in] elem_type Parent type of the extension instances
+ * @param[out] ext_list Address of the extension instance pointer array
+ * @param[out] ext_size Address of size of the extension instance pointer array
+ * @param[out] stmt Short name of the extension instance parent
+ * @return 0 for success, -1 for failure
+ */
+int lyp_get_ext_list(struct ly_ctx *ctx, void *elem, LYEXT_PAR elem_type,
+        struct lys_ext_instance ****ext_list, uint8_t **ext_size, const char **stmt);
 
 /**
  * @brief Propagate imports and includes into the main module
--- a/src/parser.h
+++ b/src/parser.h
@@ -177,6 +177,16 @@ int lyp_get_ext_list(struct ly_ctx *ctx, void *elem, LYEXT_PAR elem_type,
         struct lys_ext_instance ****ext_list, uint8_t **ext_size, const char **stmt);
 
 /**
+ * @brief Reduce extension instance pointer array
+ *
+ * @param[in] ext Address of the extension instance pointer array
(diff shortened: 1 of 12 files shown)


src/parser.h
```diff
@@ -162,6 +162,9 @@ const struct lys_node *lyp_get_yang_data_template(const struct lys_module *modul
 
 void lyp_ext_instance_rm(struct ly_ctx *ctx, struct lys_ext_instance ***ext, uint8_t *size, uint8_t index);
 
+int lyp_get_ext_list(struct ly_ctx *ctx, void *elem, LYEXT_PAR elem_type,
```
rkrejci, line 165: Please add doxygen description with the short comment about the purpose of the function and its parameters.

src/parser_yang.c
```diff
@@ -3334,26 +3334,50 @@ free_yang_common(struct lys_module *module, struct lys_node *node)
...
+        info->substmt_index = ext_list[i]->insubstmt_index;
         info->ext_index = i;
-        if (unres_schema_add_node(module, unres, ext, UNRES_EXT, (struct lys_node *)info) == -1) {
+
+        rc = unres_schema_add_node(module, unres, ext, UNRES_EXT, (struct lys_node *)info);
+        if (rc == -1) {
             return EXIT_FAILURE;
         }
+        if (!rc && !ext_list[i]) {
+            /* this extension is skipped, move all extensions after it */
```
rkrejci, line 3362: what about using `memmove(3)`?

src/parser_yin.c
```diff
@@ -105,13 +106,16 @@ lyp_yin_fill_ext(void *parent, LYEXT_PAR parent_type, LYEXT_SUBSTMT substmt, uin
...
     info->substmt = substmt;
     info->substmt_index = substmt_index;
-    info->ext_index = ext_index;
+    info->ext_index = *ext_size;
 
-    if (unres_schema_add_node(module, unres, ext, UNRES_EXT, (struct lys_node *)info) == -1) {
-        return EXIT_FAILURE;
+    rc = unres_schema_add_node(module, unres, ext, UNRES_EXT, (struct lys_node *)info);
+    if (!rc && !(*ext)[*ext_size]) {
+        // extension instance is skipped
```
rkrejci, line 113: please use block comments `/* */`

src/resolve.c
```diff
@@ -7342,6 +7361,69 @@ print_unres_schema_item_fail(void *item, enum UNRES_ITEM type, void *str_node)
     }
 }
 
+static int
+save_ignored_ext_parent(int resolve_rc, struct ly_set *ext_parents, struct ly_set *ext_par_types,
```
rkrejci, line 7365: please add doxygen description comment about the function and its parameters

src/resolve.c
```diff
@@ -7342,6 +7361,69 @@ print_unres_schema_item_fail(void *item, enum UNRES_ITEM type, void *str_node)
...
+
+    finish:
+    if(!resolve_rc || resolve_rc == -1) {
+        free_ext_data(ctx, ext_data);
+    }
+    return rt;
+}
+
+static int
+compack_ext_list(struct ly_ctx *ctx, void *elem, LYEXT_PAR elem_type)
```
rkrejci, line 7390: please add doxygen description comment about the function and its parameters

src/resolve.c
```diff
@@ -7419,6 +7520,17 @@ resolve_unres_schema_types(struct unres_schema *unres, enum UNRES_ITEM types, st
...
 
+    if ((types & UNRES_EXT)) {
+        assert(ext_parents->number == ext_par_types->number);
+        for (i = 0; i < ext_parents->number; ++i) {
+            ret = compack_ext_list(ctx, ext_parents->set.g[i], (LYEXT_PAR)ext_parents->set.g[i]);
+            LY_CHECK_GOTO(ret == -1, finish);
+        }
+    }
+
+    finish:
```
rkrejci, line 7531: formatting - move the label to the column 0

rkrejci (changes requested):

rkrejci:
Could you please also, here or in the referenced issue, a little elaborate on why the change is more complex than we've expected? I agree that the test will be needed, but except for the minor issues I've pointed out, it seems fine to me.

fanchanghu:
Hi,
The review issues have been fixed, please have a look again, thanks.

I only thought of processing the skipped extension instance when parsing it at first, however, when I worked on this, I found that it also need be processed when resolved. And there could be more than 1 skipped extension instances when resolving, so I have to saving their parents in a set and compact the extension instance arrays later. That's why I said it was a little complex than I expected.

rkrejci:
ok, the code looks fine to me now, are you going to add some test cases?

michalvasko (approved):

fanchanghu:
Yes, I'm working on the test cases now, the test cases will be pushed later, with a memory leak issue (found by the test cases) patch.

fanchanghu:
Hi,
The test cases and memory leak issue patch have been committed, please check it, thanks.

rkrejci (approved):

PR #1357, opened 2021-01-20
data tree BUGFIX log error when data file is empty

fanchanghu:
Hi,
`lyd_parse_path` does not set `ly_errno` when data file is empty, this PR is used to fix it.


michalvasko:
This is intentional, an empty (XML) file is completely valid.

fanchanghu:
Oh, it's a little confused. We now use `NULL` to check invalid data. I'll fix it in our codes.

In additional, if we treat the empty file as valid data, maybe we should validate it as the API says, which may create some default nodes?

michalvasko:
Right, there was some confusion about empty data handling, should be fine now.

Issue #1353, opened 2021-01-17
ly2: duplicate includes in submodules causes problems


choppsv1:
This parses with libyang1, and fails with libyang2

main.yang:
```
module main {
  yang-version 1.1;
  namespace "http://example.com/main";
  prefix m;
  include common;
  include a;
}
```
common.yang:
```
submodule common {
  yang-version 1.1;
  belongs-to main { prefix m; }
}
```
a.yang
```
submodule a {
  yang-version 1.1;
  belongs-to main { prefix m; }
  include common;
}
```

```
$ yanglint main.yang
libyang err : Including "common" submodule into "main" failed.
libyang err : Including "a" submodule into "main" failed.
YANGLINT[E]: Processing schema module from main.yang failed.
```

choppsv1:
I've edited this issue to replace earlier text with simple repro details.

rkrejci:
Thanks for reporting and for the data to reproduce, the issue should be now fixed in libyang2 branch. Please test and confirm that it works for you.

choppsv1:
This immediate problem is fixed; however, when I run it on the actual models from FRR that found this I get another SIGSEGV

```
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (frr-bgp-common-structure) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (frr-bgp-common-multiprotocol) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (frr-bgp-neighbor) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (frr-bgp-neighbor) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (frr-bgp-peer-group) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (frr-bgp-peer-group) are not necessary.
libyang warn: YANG version 1.1 expects all includes in main module, includes in submodules (frr-bgp-bmp) are not necessary.
libyang warn: Container "ipv4-unicast" changed to presence because it has a meaning from its "when" condition.
libyang warn: Container "ipv6-unicast" changed to presence because it has a meaning from its "when" condition.
libyang warn: Container "ipv4-labeled-unicast" changed to presence because it has a meaning from its "when" condition.
[...]

PR #1358, opened 2021-01-20
BUGFIX includes in submodules

rkrejci:
Solves #1353

### DIFF
--- a/src/tree_schema_free.c
+++ b/src/tree_schema_free.c
@@ -69,16 +69,40 @@ lysp_import_free(struct ly_ctx *ctx, struct lysp_import *import)
     FREE_ARRAY(ctx, import->exts, lysp_ext_instance_free);
 }
 
-void
-lysp_include_free(struct ly_ctx *ctx, struct lysp_include *include)
+/**
+ * @brief Common function to erase include record in main module and submodule.
+ *
+ * There is a difference since the main module is expected to have the complete list if the included submodules and
+ * the parsed submodule is shared with any include in a submodule. Therefore, the referenced submodules in the include
+ * record are freed only from main module's records.
+ *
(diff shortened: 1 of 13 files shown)


src/tree_schema_free.c
```diff
@@ -81,6 +81,18 @@ lysp_include_free(struct ly_ctx *ctx, struct lysp_include *include)
...
+{
+    return lysp_include_free_(ctx, include, 0);
+}
+
+void
+lysp_include_free(struct ly_ctx *ctx, struct lysp_include *include)
+{
+    return lysp_include_free_(ctx, include, 1);
+}
+
```
michalvasko, line 95: Based on these code changes it seems to me these 2 functions are useless because a flag is used to determine which one is called and then they only pass a flag to `lysp_include_free_()`, so it can be called directly.

src/tree_schema_free.c
```diff
@@ -69,10 +69,10 @@ lysp_import_free(struct ly_ctx *ctx, struct lysp_import *import)
     FREE_ARRAY(ctx, import->exts, lysp_ext_instance_free);
 }
 
-void
-lysp_include_free(struct ly_ctx *ctx, struct lysp_include *include)
+static void
+lysp_include_free_(struct ly_ctx *ctx, struct lysp_include *include, ly_bool main_module)
```
michalvasko, line 73: Missing doxygen, or at least document what `main_module` flag does.

michalvasko (changes requested):

src/tree_schema_free.c
rkrejci: good point, fixed

michalvasko (approved):

PR #1363, opened 2021-01-21
schema REFACTOR avoid code duplication and improve code readability

### LINKED ISSUE #1352: ly2: confused about lysc_node_children() and lysc_node_children_full()
According to the documentation, the difference between these two functions is that the first skips RPC/Action input and output nodes while the second doesn't. Looking at the code, however, they seem to equivalent. Both have a `flags` parameter that is used to control the behavior for RPC/Action nodes, which aren't skipped in any case.

rkrejci:
lysc_node_children_p() and lysc_node_children_full() were wery similar,
refactoring should avoid code duplication and emphasis the differences
between the functions.

Fixes #1352

### DIFF
--- a/src/tree_schema_helpers.c
+++ b/src/tree_schema_helpers.c
@@ -1468,7 +1468,7 @@ lysc_node_children(const struct lysc_node *node, uint16_t flags)
         return NULL;
     }
 
-    children = lysc_node_children_p((struct lysc_node *)node, flags);
+    children = lysc_node_children_p(node, flags);
     if (children) {
         return *children;
     } else {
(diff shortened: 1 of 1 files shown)


src/tree_schema_helpers.c
```diff
@@ -1479,30 +1479,24 @@ lysc_node_children(const struct lysc_node *node, uint16_t flags)
...
-        return ((struct lysc_notif *)node)->data;
-    default:
-        return NULL;
+    } else {
+        struct lysc_node **children = lysc_node_children_p(node, flags);
+        if (children) {
+            return *children;
+        } else {
+            return NULL;
+        }
```
michalvasko, line 1499: Why not simply call `lysc_node_children()` instead of this code?

michalvasko (changes requested):

src/tree_schema_helpers.c
rkrejci: you are right, changed

michalvasko (approved):

PR #1364, opened 2021-01-21
tree printer FEATURE basic implementation of YANG tree format printer

lePici:


src/printer_tree.c
```diff
@@ -0,0 +1,2965 @@
...
+    va_start(ap, arg_count);
+    p->pf(p->out, arg_count, ap);
+    va_end(ap);
+}
+
+#define PRINT_N_TIMES_BUFFER_SIZE 16
+
+/** Print character n times. */
+static void
+trg_print_n_times(int32_t n, char c, struct trt_printing *p)
```
rkrejci, line 621: if using `ly_print()`, which wrapps `printf()`, the `"%*c", n, c` should do the job

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+
+/** Setting '|' symbol because node is divided or it is not last sibling. */
+static struct trt_wrapper
+trp_wrapper_set_mark(struct trt_wrapper wr)
+{
+    wr.bit_marks1 |= 1U << wr.actual_pos;
+    return trp_wrapper_set_shift(wr);
+}
+
+/** Setting ' ' symbol if node is last sibling otherwise set '|'. */
```
rkrejci, line 692: please always describe the function's parameters

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+            ly_print_(out, "%*c", TRD_INDENT_BTW_SIBLINGS, ' ');
+        }
+    }
+}
+
+/** Create struct trt_node_name as empty. */
+static struct trt_node_name
+trp_empty_node_name()
+{
+    struct trt_node_name ret;
```
rkrejci, line 749: please initiate the whole structure

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+trp_opts_keys_are_set(struct trt_node_name node_name)
+{
+    return node_name.type == TRD_NODE_KEYS;
+}
+
+/** Create empty struct trt_type. */
+static struct trt_type
+trp_empty_type()
+{
+    struct trt_type ret;
```
rkrejci, line 773: initiate whole structure

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+ *****************************************************************************/
+
+/**
+ * @brief List of features in node.
+ *
+ * The iffeature is just ly_bool because printing is not provided by the printer component (trp).
+ *
+ * See: trp_empty_iffeature, trp_iffeature_is_empty, trp_print_iffeatures
+ */
+struct trt_iffeature {
```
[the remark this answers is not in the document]
rkrejci: ok, probably I was wrong, it is done via `trp_iffeature_is_empty`, anyway I don't think it is necessary since it is just a presence flag

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+            trp_print_divided_node(ind_node2.node, ppck, tmp, mll, out);
+        }
+    } else if (ind_node1.indent.type == TRD_INDENT_IN_NODE_FAILED) {
+        /* node name is too long */
+        trp_print_line_up_to_node_name(node, ipck.wrapper, out);
+        if (trp_node_body_is_empty(node)) {
+            return;
+        } else {
+            ly_print_(out, "\n");
+            struct trt_pair_indent_node ind_node2 = trp_second_half_node(node, ind_node1.indent);
```
rkrejci, line 1607: please keep the variables declarations at the beginning of the block

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+/******************************************************************************
+ * Definition of Tro reading functions
+ *****************************************************************************/
+
+/** Return trt_parent_cache filled with default values. */
+static struct trt_parent_cache
+tro_empty_parent_cache()
+{
+    return (struct trt_parent_cache)
+           {
```
rkrejci, line 1625: no newline for `{` if it is not the function's body

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+
+    if (tc.pn->nodetype & (LYS_RPC | LYS_ACTION)) {
+        if (tc.tpn == tc.pn) {
+            /* just go to the top-node's sibling */
+            tc.pn = tc.pn->next;
+            tc.tpn = tc.pn;
+        } else {
+            /* try to go to the notif node as sibling */
+            if (!tc.pn->next) {
+                tc.pn = (const struct lysp_node *) lysp_node_notifs(tc.pn->parent);
```
rkrejci, line 1683: no spaces between cast's parentheses and cast entity

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+tro_read_node(struct trt_parent_cache ca, const struct trt_tree_ctx *tc)
+{
+    assert(tc != NULL && tc->pn != NULL);
+    const struct lysp_node *pn = tc->pn;
+
+    assert(pn->nodetype != LYS_UNKNOWN);
+
+    struct trt_node ret = trp_empty_node();
+
+    /* remember:
```
rkrejci, line 1933: this is not true after the latest changes

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+    /* TODO: TRD_NODE_TOP_LEVEL2 aka '@' is not supported right now. */
+    /* set type of the node */
+    ret.name.type = tro_resolve_node_type(pn, ca.last_list);
+
+    /* TODO: ret.name.module_prefix is not supported right now. */
+    ret.name.module_prefix = NULL;
+
+    /* set node's name */
+    ret.name.str =
+            pn->nodetype & (LYS_INPUT) ? "input" :
```
rkrejci, line 1955: it is not necessary to handle this now, the nodes have the "input"/"output" name now (if not, it is a bug)

src/printer_tree.c
```diff
@@ -0,0 +1,2757 @@
...
+    assert(tc != NULL && tc->module != NULL && tc->module->parsed != NULL);
+    const struct lysp_node_augment *augs;
+
+    /* if next_augment func was called for the first time */
+    if (tc->section != TRD_SECT_AUGMENT) {
+        tc->section = TRD_SECT_AUGMENT;
+        augs = tc->module->parsed->augments;
+    } else {
+        /* get augment sibling from top-node pointer */
+        augs = (const struct lysp_node_augment *) tc->tpn->parent->next;
```
rkrejci, line 2136: no space before the cast variable

rkrejci (changes requested):

src/printer_internal.h
```diff
@@ -122,6 +122,18 @@ LY_ERR yin_print_parsed_module(struct ly_out *out, const struct lys_module *modu
...
+/**
+ * @brief YANG Tree Diagram printer of the parsed submodule. Full YANG Tree printer.
+ *
+ * @param[in] out Output specification.
+ * @param[in] module Main module.
+ * @param[in] options Schema output options (see @ref schemaprinterflags).
+ * @param[in] line_length Maximum characters to be printed on a line, 0 for unlimited. Only for ::LYS_OUT_TREE printer.
+ * @return LY_ERR value, number of the printed bytes is updated in ::ly_out.printed.
+ */
+LY_ERR tree_print_parsed_module(struct ly_out *out, const struct lys_module *module, uint32_t options,
```
michalvasko, line 134: I would argue that the compiled module is printed more than the parsed module. Maybe omit `parsed`?

src/printer_tree.c
```diff
@@ -0,0 +1,3107 @@
...
+    TRD_SECT_MODULE = 0,    /**< The node belongs to the "module: <module_name>:" label. */
+    TRD_SECT_AUGMENT,       /**< The node belongs to some "augment <target-node>:" label. */
+    TRD_SECT_RPCS,          /**< The node belongs to the "rpcs:" label. */
+    TRD_SECT_NOTIF,         /**< The node belongs to the "notifications:" label. */
+    TRD_SECT_GROUPING,      /**< The node belongs to some "grouping <grouping-name>:" label. */
+    TRD_SECT_YANG_DATA      /**< The node belongs to some "yang-data <yang-data-name>:" label. */
+} trt_actual_section;
+
+/**
+ * @brief Ancestors who have a strong influence on their children
```
michalvasko, line 472: This comment confused me for a second, had to remind myself what it is I am reviewing. (ignore :D )

src/printer_tree.c
```diff
@@ -0,0 +1,3107 @@
...
+ */
+typedef enum {
+    TRD_ANCESTOR_ELSE = 0,
+    TRD_ANCESTOR_RPC_INPUT,
+    TRD_ANCESTOR_RPC_OUTPUT,
+    TRD_ANCESTOR_NOTIF
+} trt_ancestor_type;
```
michalvasko, line 479: I just want to add that you can detect all of these ancestors directly from `struct lysc_node` (flags `LYS_IS_INPUT`, `LYS_IS_OUTPUT`, `LYS_IS_NOTIF`). But not from `struct lysp_node` so I am not sure whether that helps in any way.

src/printer_tree.c
```diff
@@ -0,0 +1,3107 @@
...
+        return 0;
+    }
+
+    const char *hit = strstr(src, word);
+
+    if (hit) {
+        /* word was founded at the begin of src
+         * OR it match somewhere after delim
+         */
+        if ((hit == src) || (*(hit - 1) == delim)) {
```
michalvasko, line 584: Please use `[]` when accessing `char *` array (`hit[-1]`). You can use this for all other cases but we generally prefer to use `[]` for all array types.

src/printer_tree.c
```diff
@@ -0,0 +1,3107 @@
...
+            char delim_or_end = (hit + strlen(word))[0];
+            if ((delim_or_end == '\0') || (delim_or_end == delim)) {
+                return 1;
+            }
+        }
+        /* else -> hit is just substr and it's not the whole word */
+        /* jump to the next word */
+        for ( ; src[0] != '\0' && src[0] != delim; src++) {}
+        /* skip delim */
+        src = src[0] == '\0' ? src : src + 1;
```
michalvasko, line 597: Boolean operations should be in parentheses, `uncrustify` does not do this? I think it worked quite reliably.

src/printer_tree.c
```diff
@@ -0,0 +1,3107 @@
...
+ * Bit indicates where the sibling symbol must be printed.
+ * This place is in multiples of TRD_INDENT_BTW_SIBLINGS.
+ *
+ * See: trp_init_wrapper_top, trp_init_wrapper_body, trp_wrapper_set_mark,
+ * trp_wrapper_set_shift, trp_wrapper_if_last_sibling, trp_wrapper_eq,
+ * trp_print_wrapper
+ */
+struct trt_wrapper {
+    trd_wrapper_type type;  /**< Location of the wrapper. */
+    uint64_t bit_marks1;    /**< The set bits indicate where the '|' character is to be printed. */
```
michalvasko, line 199: Limits the printed schema tree depth to 64 bits. It is probably okay (not sure if even the current *libyang* did not have this limitation) but perhaps at least mention it somewhere. Also, please put asserts or internal errors on places where this could overflow (in case the depth is more than 64).

src/printer_tree.c
```diff
@@ -0,0 +1,3107 @@
...
+    return node_name.str == NULL;
+}
+
+/**
+ * @brief Create empty struct trt_type.
+ */
+static struct trt_type
+trp_empty_type()
+{
+    return (struct trt_type) {
```
michalvasko, line 823: An interesting construct but I do not like it. It is probably just a personal preference, but I would expect this to simply be a macro (the code in `{}`). Just like `PTHREAD_MUTEX_INITIALIZER`, for example, any static initialization.

src/printer_tree.c
```diff
@@ -0,0 +1,3107 @@
...
+ * @param[in] mll is max line length.
+ * @param[in,out] out is output handler.
+ * @return .type == TRD_INDENT_IN_NODE_DIVIDED - the node does not fit in the line, some indent variable has negative value as a line break sign.
+ * @return .type == TRD_INDENT_IN_NODE_NORMAL - the node fits into the line, all indent variables values has non-negative number.
+ * @return .type == TRD_INDENT_IN_NODE_FAILED - the node does not fit into the line, all indent variables has negative or zero values, function failed.
+ */
+static struct trt_pair_indent_node
+trp_try_normal_indent_in_node(struct trt_node node, struct trt_pck_print pck, struct trt_pck_indent indent, size_t mll, struct ly_out *out)
+{
+    struct trt_pair_indent_node ret = {indent.in_node, node};
```
michalvasko, line 1666: Would prefer for this to be `struct trt_pair_indent_node ret = {.indent = indent.in_node, .node = node};` (and similarly on other places).

src/printer_tree.c
```diff
@@ -0,0 +1,3107 @@
...
+ * @param[in] origin_tc is context of the tree.
+ * @return A patch structure that has prepared pointers for updating struct trt_tree_ctx.
+ *  If sibling exists then .pn points to him, otherwise is set to NULL.
+ *  The .tpn points to its sibling if it exists and if .pn points to the same node as .tpn,
+ *  otherwise .tpn value from origin_tc is copied.
+ */
+static struct trt_tree_ctx_node_patch
+tro_read_next_sibling(const struct trt_tree_ctx *origin_tc)
+{
+    assert(origin_tc != NULL && origin_tc->pn != NULL);
```
michalvasko, line 1833: Generally, you can leave out `!= NULL`, it is often used that way for checking parameters. But you can use it for any NULL pointer comparison.

src/printer_tree.c
```diff
@@ -0,0 +1,3107 @@
...
+        tro_read_node,
+        tro_read_if_sibling_exists
+    };
+
+    struct trt_fp_print print_fp = {
+        tro_print_features_names,
+        tro_print_keys
+    };
+
+    return (struct trt_printer_ctx) {
```
michalvasko, line 3034: Connected to a comment before, why not declare a static variable of this type, fill it properly, and return it? Better yet, set return type to `void` and make `struct trt_printer_ctx *` a parameter that you simply fill correctly. That is how we usually write dynamic initializers, avoids allocating the whole structure as return value and then the copy into the local structure.

src/printer_internal.h
rkrejci: I think that right now it prints information from the parsed tree. It is prepared to add also the compiled tree, but the format specification requires things like augments or groupings. So, the decision was to use the parsed trees for now in cost of for example not expanding uses (which is allowed by format specs)

rkrejci (changes requested):

src/printer_internal.h
michalvasko: Right, makes sense then.

src/printer_tree.c
```diff
@@ -97,6 +85,12 @@ struct ly_out_clb_arg {
     LY_ERR last_error;                  /**< The last error that occurred. If no error has occurred, it will be LY_SUCCESS. */
 };
 
+/**
+ * @brief Initialize struct ly_out_clb_arg with default settings.
+ */
+#define TRP_DEFAULT_LY_OUT_CLB_ARG \
+    (struct ly_out_clb_arg){.mode = TRD_PRINT, .out = out, .counter = 0, .last_error = LY_SUCCESS}
```
rkrejci, line 92: When the macro is not defined just temporarily inside a function, do not use variables directly, define that variable `OUT` for the macro and use it in initialization)

src/printer_tree.c
```diff
@@ -135,6 +129,12 @@ struct trt_pck_print {
     struct trt_fp_print fps;                /**< Print function. */
 };
 
+/**
+ * @brief Initialize struct trt_pck_print by parameters.
+ */
+#define TRP_INIT_PCK_PRINT(tree_ctx_ptr, fp_print) \
+    (struct trt_pck_print){.tree_ctx = tree_ctx_ptr, .fps = fp_print}
```
rkrejci, line 136: Use upper case for the macro parameters, it avoids possible confusion

src/printer_tree.c
```diff
@@ -208,6 +239,12 @@ struct trt_pck_indent {
     struct trt_indent_in_node in_node;  /**< Indent in node. */
 };
 
+/**
+ * @brief Initialize struct trt_pck_indent by parameters.
+ */
+#define TRP_INIT_PCK_INDENT(wr, indent_in_node) \
```
rkrejci, line 245: use upper case for macro parameters

src/printer_tree.c
```diff
@@ -283,6 +320,18 @@ struct trt_node_name {
...
+/**
+ * @brief Create struct trt_node_name as empty.
+ */
+#define TRP_EMPTY_NODE_NAME \
+    (struct trt_node_name){.type = TRD_NODE_ELSE, .module_prefix = NULL, .str = NULL}
+
+/**
+ * @brief Check if struct trt_node_name is empty.
+ */
+#define TRP_NODE_NAME_IS_EMPTY(node_name) \
```
rkrejci, line 332: upper case for macro parameters (and there are some more cases of this below)

rkrejci (changes requested):
almost ready :)

michalvasko (approved):

rkrejci (approved):

PR #1365, opened 2021-01-21
metadata plugin BUGFIX make the used global variable thread safe

### LINKED ISSUE #1344: Sporadic crashes when running libyang2 on multiple pthreads
Code to reproduce the problem: https://pastebin.com/raw/5mFVYx0M

In this code, a thousand pthreads are created in a loop. A separate libyang context is created for each pthread and YANG modules are loaded into them in parallel. Running this code causes libyang to crash more often than not with backtraces like this: https://pastebin.com/raw/Ssc2eeQN

Those crashes don't happen if `thread_main()` is called sequentially instead of in parallel (using pthreads). I suspect there's a data race going on somewhere which might explain the unpredictable behavior with pthreads.

@michalvasko Opening this issue as per your request so that we can keep track of this problem.

rkrejci:
Fixes #1344

### DIFF
--- a/src/plugins_exts_metadata.c
+++ b/src/plugins_exts_metadata.c
@@ -33,15 +33,15 @@ LYEXT_VERSION_CHECK
 #define ANNOTATION_SUBSTMT_DSC     4
 #define ANNOTATION_SUBSTMT_REF     5
 
-struct lysc_ext_substmt annotation_substmt[] = {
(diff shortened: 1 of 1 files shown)


rkrejci:
helgrind still reports some issue in `mktime()`, but it should be thread-safe according to specification, so I believe that it is false positive. The test code also doesn't crash now.

michalvasko (approved):

Issue #1351, opened 2021-01-15
ly2: issue with lysc_module_dfs_full()


rwestphal:
I've just done a simple test with the new `lysc_module_dfs_full()` function but it doesn't seem to be working properly, at least not with the "ietf-isis" YANG module. The issue is that `lysc_module_dfs_full()` sometimes calls its callback providing an invalid schema node pointer.

Code to reproduce the problem (should crash in the `snprintf()` line while trying to derefence the `snode` pointer):
```c
#include <stdlib.h>
#include <stdio.h>
#include <err.h>

#include <libyang/libyang.h>

#define YANG_SEARCH_PATH "/usr/local/share/yang/"

LY_ERR snode_print_config_flag(struct lysc_node *snode, void *data,
			       ly_bool *dfs_continue)
{
	char buf[BUFSIZ];

	//lysc_path(snode, LYSC_PATH_LOG, buf, sizeof(buf));
	snprintf(buf, sizeof(buf), snode->name);
	printf("%s: %u", buf, snode->flags & LYS_CONFIG_W);

	return LY_SUCCESS;
}

int main(int argc, char **argv)
{
	struct ly_ctx *ly_ctx;
	const struct lys_module *module;

	/* initialization */
	if (ly_ctx_new(YANG_SEARCH_PATH, 0, &ly_ctx) != LY_SUCCESS)
		errx(1, "ly_ctx_new(): %s", ly_errmsg(ly_ctx));

	module = ly_ctx_load_module(ly_ctx, "ietf-isis", NULL, NULL);
	if (module == NULL)
		errx(1, "ly_ctx_load_module: %s", ly_errmsg(ly_ctx));

	if (lysc_module_dfs_full(module, snode_print_config_flag, NULL)
	    != LY_SUCCESS)
		errx(1, "lysc_module_dfs_full(): %s", ly_errmsg(ly_ctx));

	/* cleanup */
	ly_ctx_destroy(ly_ctx, NULL);

	return 0;
}
```

rkrejci:
the callback receives all the nodes in the tree and the problem is they are not `struct lysc_node` in all cases - the input/output structures differ. The common part is always the `nodetype`. I see 2 possible options here:
1. we will live with `lysc_node` here and will improve documentation for the callbacks
2. we will add a new structure with only the `nodetype` member for these cases to be clear that such nodes cannot be used without a proper cast

[...]

PR #1366, opened 2021-01-21
schema CHANGE move priv field back to common location

choppsv1:
There's a setter but not a getter for private data, which makes it ugly
to get the value in client code. But beyond that there's existing code
that assumes this pointer is going to be in the same location (i.e., one
can pull it from a common lys_node, now lysc_node), b/c that's how it
worked in libyang1. Let's not make extra work and debugging for everyone
and leave the priv accessible from the same memory offset as was the
case with libyang1.

### DIFF
--- a/src/tree_schema.h
+++ b/src/tree_schema.h
@@ -1512,6 +1512,7 @@ struct lysc_action {
     uint8_t hash[LYS_NODE_HASH_COUNT]; /**< schema hash required for LYB printer/parser */
     struct lys_module *module;       /**< module structure */
     struct lysc_node *parent;        /**< parent node (NULL in case of top level node - RPC) */
+    void *priv;                      /** private arbitrary user data, not used by libyang */
 
     struct lysc_ext_instance *input_exts;  /**< list of the extension instances of input ([sized array](@ref sizedarrays)) */
     struct lysc_ext_instance *output_exts; /**< list of the extension instances of outpu ([sized array](@ref sizedarrays)) */
@@ -1524,8 +1525,6 @@ struct lysc_action {
     struct lysc_action_inout input;  /**< RPC's/action's input */
     struct lysc_action_inout output; /**< RPC's/action's output */
     struct lysc_when **when;         /**< list of pointers to when statements ([sized array](@ref sizedarrays)) */
-    void *priv;                      /** private arbitrary user data, not used by libyang */
-
 };
 
 struct lysc_notif {
@@ -1534,6 +1533,7 @@ struct lysc_notif {
     uint8_t hash[LYS_NODE_HASH_COUNT]; /**< schema hash required for LYB printer/parser */
     struct lys_module *module;       /**< module structure */
     struct lysc_node *parent;        /**< parent node (NULL in case of top level node) */
+    void *priv;                      /** private arbitrary user data, not used by libyang */
 
     struct lysc_node *data;          /**< first child node (linked list) */
     struct lysc_must *musts;         /**< list of must restrictions ([sized array](@ref sizedarrays)) */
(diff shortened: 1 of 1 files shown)


rkrejci:
I'm actually not strongly against the change, but could you explain the motivation a little bit more? From the ABI point of view, libyang 2.0 is completely different from libyang 1.0, so you will have to recompile (and most probably modify since also API changed) all the code working with libyang data. And you definitely should not access the data directly by some magic number, you can use `offsetof(3)` and again, recompilation should make it work.

rkrejci:
maybe some code snippet would help understand what problem you are solving

choppsv1:
Sure, this fails if the node indicated by the path is a lysc_action instead of lysc_node.

```
struct nb_node *nb_node_find(const char *path)
{
	const struct lysc_node *snode;
	snode = lys_find_path(ly_native_ctx, NULL, path, 0);
	if (!snode)
		return NULL;
	return snode->priv;
}
```

choppsv1:
FWIW, I am aware that I could write code that did the right thing, it would be through writing my own getter routine. If you really want to hide the priv field then a getter should be supplied to pair with the setter. However, there's also the aspect of not causing damage or porting hassles to existing code w/o justification. The code snippet above (continues) to work when one does the simple lys_node -> lysc_node conversion if priv is in the same location (shared) between all subtypes. There's still plenty of casting going on in libyang (internally and externally) and it's just safer to have the priv field be in the same location. This is just what one has to deal with given C and the choices made in libyang (e.g., not using a shared common structure to avoid casting).

rkrejci:
ok, but that's not the problem just for the `priv` member. The root cause of the problem is the same as in #1351 - not all structures that can appear behind `struct lysc_node *` pointer are actually compatible. Our first idea when #1351 appeared was to use a narrower structure just with `nodetype` member to make the caller cast the structure correctly. But now it seems that such a situation could be quite common and always checking nodetype would be a pain.

So we would like to redesign a little bit the `lysc_notif` and `lysc_action` with its input/output to make them compatible with `lysc_node` at the cost of several wasted bytes for members that are not necessary. So I'll not merge this PR, but I'm going to prepare another PR which should cover even the `priv` member. Then I would appreciate your feedback on this PR.

choppsv1:
The optimal memory solution is to use an embedded named common structure at the top of the object in all the type specific structs to hold the common fields. This allows avoiding almost all casting as the common functions would operate on the common first field. E.g.,

```
struct lysc_node {
    uint16_t nodetype;               /**< LYS_CONTAINER */
    uint16_t flags;                  /**< [schema node flags](@ref snodeflags) */
 ...
}
struct lysc_node_container {
    struct lysc_node node; /* <--- COMMON node data */
    struct lysc_node *child;         /**< first child node (linked list) */
    struct lysc_must *musts;         /**< list of must restrictions ([sized array](@ref sizedarrays)) */
    ...
};
```

And then if you have ```struct lysc_node_container *p``` and you need to pass it to something that operates on a ```struct lysc_node *``` you just use ```&p->node```.

michalvasko:
> This allows avoiding almost all casting as the common functions would operate on the common first field.

Yes, this avoids casting but no matter what node you have, if you want to access the common attributes, you must always use `specific_node->common_node->attr` instead of `specific_node->attr`. So, making all the nodes to have the common part seems like the best solution.

choppsv1:
> > This allows avoiding almost all casting as the common functions would operate on the common first field.
> 
> Yes, this avoids casting but no matter what node you have, if you want to access the common attributes, you must always use `specific_node->common_node->attr` instead of `specific_node->attr`. So, making all the nodes to have the common part seems like the best solution.

I'm not sure I agree saving a couple keystrokes is worth losing static type checking.

Most of the APIs would work with the common structure not specific nodes. So it would be ```common->attr```rather than ```specific->common.attr```. For the few type specific functions are you saying that it's worth saving a couple keystrokes to lose static type checking? (the field name doesn't have to be "common" it could be ```c```)? If that's really important one could use a local variable to save the 2 keystrokes (i.e., ```struct common *c = &specific->c; c->attr1; c->attr2;``` instead of ```specific->c.attr1; specific->c.attr2```. C isn't an OOP so it does required some extra work to use those concepts. :)

rkrejci:
Hi, the problem should be solved by #1380, so I'm going to close this PR.

Regarding the casting member - in my PR I've tried to connect pros from both approaches using an anonymous union. The definition might seem a little weird, but I believe that it improves usability as well as the correctness of use with the cast member as you proposed.

PR #1368, opened 2021-01-22
xml printer BUGFIX print namespace of custom attribute in netconf filter

fanchanghu:
Hi,
I'm using a custom attribute (defined as yang annotation) in netconf filter, like following:

```
<get-config xmlns="urn:ietf:params:xml:ns:netconf:base:1.0">
  <source>
    <running/>
  </source>
  <filter type="xpath" xmlns:a="urn:a" select="/a:P" xmlns:b="urn:b" b:testAttr2="222"/>
</get-config>
```

libyang can parse it successfully, but miss the namespace of the attribute when printing it as XML.

This PR is used to fix this issue.

### DIFF
--- a/src/printer_xml.c
+++ b/src/printer_xml.c
@@ -72,6 +72,20 @@ free_mlist(struct mlist **mlist) {
     }
 }
 
+inline static int
+is_netconf_filter(const struct lyd_node *node)
+{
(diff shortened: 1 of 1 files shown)


michalvasko:
Okay, thanks, the fix is correct, I am just merging it to `devel`.

PR #1376, opened 2021-01-26
basic type test

radek-isa:


tests/utests/utests.h
```diff
@@ -648,6 +758,7 @@ struct utest_context {
  *                     CHECK_LYD_VALUE_ ## TYPE_VAL.
  */
 #define CHECK_LYD_VALUE(NODE, TYPE_VAL, ...) \
+    assert_ptr_equal((NODE).realtype->plugin, &(ly_builtin_type_plugins[LY_TYPE_ ## TYPE_VAL]));\
```
michalvasko, line 761: May work for now, but once plugins are properly implemented, it most likely will not.

tests/utests/types/int8.c
```diff
@@ -0,0 +1,1758 @@
...
+    assert_int_equal(range->parts[0].max_64, 127);
+    lysp_leaf = (void *) mod->parsed->data;
+    CHECK_LYSP_NODE_LEAF(lysp_leaf, NULL, 0, 0x0, 0, "port", 0, 0, NULL, 0, 0, NULL, NULL);
+    CHECK_LYSP_TYPE(&(lysp_leaf->type), 0, 0, 0, 0, 0, 0x80, 0, 0, "int8", 0, 0, 1, 1, 0, 0);
+    CHECK_LYSP_RESTR(lysp_leaf->type.range, "min .. max", NULL, NULL, NULL, 0, NULL);
+
+
+    //TEST ERROR -60 .. 0 | 0 .. 127
+    schema = MODULE_CREATE_YANG("ERR0", "leaf port {type int8 {range \"-60 .. 0 | 0 .. 127\";}}");
+    UTEST_ADD_MODULE(schema, LYS_IN_YANG, NULL, LY_EVALID, &mod);
```
michalvasko, line 270: You are not adding a module but using `UTEST_ADD_MODULE` macro? Instead of rewriting the macro used everywhere, write a new one that will expect the module to fail to load.

michalvasko (commented):
It would be helpful if you run `uncrustify` on your test, but you will have to do so manually since tests are not formatted automatically yet.

michalvasko (commented):
Except for some minor formatting problems and using a type I did not even know existed (`int unsigned`), it seems fine. Also, I would probably rename `UTEST_ADD_MODULE_FAILED` to something like `UTEST_INVALID_MODULE` but does not matter much.

michalvasko (approved):

rkrejci (changes requested):

tests/utests/types/int8.c
```diff
@@ -0,0 +1,1720 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#define _POSIX_C_SOURCE 200809L /* strdup */
+#include <ctype.h>
+
+#define  _UTEST_MAIN_
+#include "../utests.h"
```
rkrejci, line 19: this one include with the `_UTEST_MAIN_` is an exception to the order of includes, it is supposed to be always the first.
I see a problem in the `utests.h` regarding the ordering, but I'm going to fix it in standalone bugfix, the rule is that `utests.h` is included first in all tests

tests/utests/types/int8.c
```diff
@@ -0,0 +1,1720 @@
...
+    CHECK_LYD_VALUE(value, INT8, "-23", -23);
+    assert_ptr_equal(value.realtype, lysc_type);
+    type->free(UTEST_LYCTX, &value);
+
+    /*
+     * minor tests
+     * dynamic alocated input text
+     */
+    val_text = "0xa";
+    char *alloc = (char *) malloc(strlen(val_text) + 1);
```
rkrejci, line 1450: variable definitions are expected at the beginning of a block (there are more cases of this) and there is no space expected between the cast's parentheses and the variable/function.

rkrejci (changes requested):

PR #1379, opened 2021-01-26
Fix various errors reported by UBSAN

jktjkt:
## Fix undefined behavior on memcpy()

libnetconf2's test `test_server_thread` eventually ends up in this stack trace:

```
libyang/src/tree_schema.c:3288:16: runtime error: null pointer passed as argument 1, which is declared to never be null
/nix/store/ypih4394q488ljr421x8jak55vmr0ckn-glibc-2.32-dev/include/string.h:44:28: note: nonnull attribute specified here
#0 0x7f565ebf8f6e in lys_node_dup_recursion libyang/src/tree_schema.c:3288:9
#1 0x7f565ebf3409 in lys_node_dup libyang/src/tree_schema.c:3609:14
#2 0x7f565ebbf62c in yang_check_deviation libyang/src/parser_yang.c:4646:26
#3 0x7f565eba6531 in yang_check_sub_module libyang/src/parser_yang.c:4784:13
#4 0x7f565eba3d1c in yang_read_module libyang/src/parser_yang.c:2705:13
#5 0x7f565ebd1077 in lys_parse_mem_ libyang/src/tree_schema.c:1083:15
#6 0x7f565ebd31ab in lys_parse_fd_ libyang/src/tree_schema.c:1271:14
#7 0x7f565e92f291 in ly_ctx_load_localfile libyang/src/context.c:916:39
#8 0x7f565e92c31b in ly_ctx_load_sub_module libyang/src/context.c:1063:19
#9 0x7f565e92fdcc in ly_ctx_load_module libyang/src/context.c:1102:12
#10 0x597422 in main libnetconf2/tests/test_server_thread.c:686:5
#11 0x7f565dc98dbc in __libc_start_main (/nix/store/kah5n342wz4i0s9lz9ka4bgz91xa2i94-glibc-2.32/lib/libc.so.6+0x23dbc)
#12 0x42a4d9 in _start /build/glibc-2.32/csu/../sysdeps/x86_64/start.S:120
```

In other words, it would call `memcpy` with the source parameter being NULL, which is [explicitly said to be undefined behavior](https://www.imperialviolet.org/2016/06/26/nonnull.html).

## Disable strict pointer aliasing
    
I'm not really a C-level language lawyer, but the way how the code routinely uses two differently-types pointers to access the same object is undefined behavior in C. Just having "the same layout" of individual struct members [does not make these types compatible](https://en.cppreference.com/w/c/language/type#Compatible_types), and that means that it is necessary to disable [strict aliasing](https://en.cppreference.com/w/c/language/object#Strict_aliasing) rules in the whole of libyang, as far as I can tell.

## Fix a warning from UBSAN
    
I won't pretend I understand what UBSAN is really trying to tell me
here, but that assignment can be rewritten in a way which does not cause
the following warning:
```
libyang/src/tree_data.c:5564:19: runtime error: member access within address 0x60600005c7e0 with insufficient space for an object of type 'struct lyd_node'
0x60600005c7e0: note: pointer points here
1c 00 00 26  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00
	   ^
```
Fixes: https://github.com/sysrepo/sysrepo/issues/2295

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE debug)
 endif()
 
-set(CMAKE_C_FLAGS         "${CMAKE_C_FLAGS} -Wall -Wextra -std=gnu99")
+set(CMAKE_C_FLAGS         "${CMAKE_C_FLAGS} -Wall -Wextra -std=gnu99 -fno-strict-aliasing")
 set(CMAKE_C_FLAGS_RELEASE "-O3 -DNDEBUG")
 set(CMAKE_C_FLAGS_PACKAGE "-g -O3 -DNDEBUG")
 set(CMAKE_C_FLAGS_DEBUG   "-g -Og")
--- a/src/tree_data.c
+++ b/src/tree_data.c
@@ -5561,7 +5561,7 @@ _lyd_dup_node(const struct lyd_node *node, const struct lys_node *schema, struct
         new_any = calloc(1, sizeof *new_any);
         new_node = (struct lyd_node *)new_any;
         LY_CHECK_ERR_GOTO(!new_node, LOGMEM(ctx), error);
-        new_node->schema = (struct lys_node *)schema;
+        new_any->schema = (struct lys_node *)schema;
 
         if (_lyd_dup_node_common(new_node, node, ctx, options)) {
             goto error;
--- a/src/tree_schema.c
+++ b/src/tree_schema.c
@@ -3285,7 +3285,9 @@ lys_node_dup_recursion(struct lys_module *module, struct lys_node *parent, const
             }
         }
     } else {
-        memcpy(retval->iffeature, node->iffeature, retval->iffeature_size * sizeof *retval->iffeature);
+        if (node->iffeature_size) {
+            memcpy(retval->iffeature, node->iffeature, retval->iffeature_size * sizeof *retval->iffeature);
+        }
     }
 
     /*


michalvasko:
Hmm, thanks for all the information, I think I have finally understood what strict aliasing is supposed to be (well, not from these formal texts but some other ones). Then I have several questions, not sure if you will be able to help me with all of them.

Why `tree_data.c:5497` is fine when `5564` is not? I suppose this may be caused by something hidden deep in the specifics or your compiler (like I said, for me UBSAN did not report anything). But if this really is a problem of strict aliasing, then the new compiler flag should forbid it and the error should be fixed, right? Without actually changing the code. I am not sure this is worth investigating more but based on what I understood, the problem with strict aliasing manifests by having unexpected values in a variable because it was written to using another incompatible variable/alias (specifically, variable with incompatible type with the original one). I do not think this really fits here.

Lastly, you are right about the `memcpy` fix, that was wrong without anything to add.

jktjkt:
Yeah, I also suspect that UBSAN only catches *some* cases of UB. Back in 2017 there were patches to llvm/clang introducing TYSAN, a special sanitizer exactly for this purpose, unfortunately, they have not been merged yet. Today UBSAN is therefore the best tool that we have available, I'm afraid.

It is important to note that even though `-fno-strict-aliasing` disables some optimizations, the code that is present in the repo is still, technically speaking, hitting that dreaded Undefined Behavior, and therefore "ill-formed". My understanding of `-fno-strict-aliasing` is that it is meant as a workaround rather than a fix, and that's (probably) why UBSAN keeps flagging (some) behavior that it can identify as UB as such.

Are you going to merge this, please? With this PR merged, all of libyang+sysrepo+libnetconf2+netopeer2 test suite as well as the test suite of our internal CzechLight SW and the [velia](https://github.com/CESNET/velia) system manager/monitor and the [netconf-cli](https://github.com/CESNET/netconf-cli) pass without hitting UBSAN warnings -- at least on my clang 11. That makes it, in turn, possible to enable UBSAN coverage for our internal code, which is a big bonus to us.

michalvasko:
Based on what I read, there are 2 different problems mixed. One is the problem of struct aliasing. It should be an optimization that requires specific accessing to variables (using struct aliases). If you do not conform to this requirement, there may be unexpected behavior of your application. If you disable it, it should not be possible for this to cause any problems.

Then there seem to be certain requirements on compatible types in the C standard, which left all other cases as undefined behavior. My guess is that this may have to do with alignment of the members, for which there are no rules for 2 structures with different members. Whether this can cause any real problems I do not know. But I agree we should try to conform to if possible.

So yes, I can merge this. For *libyang* version 2 there should not be these kinds of problems after the currently discussed PR #1380 is merged.

PR #1380, opened 2021-01-26
unify schema node structures

### LINKED ISSUE #1351: ly2: issue with lysc_module_dfs_full()

rkrejci:
Not all the structures that can appear behind `lysc_node *` or `lysp_node *` pointers were actually compatible with these structures, so the `nodetype` had to be carefully handled. This change unifies all the relevant structures keeping directly available basic members even in the derivated structures but also adding new `node` member usable as a generic `lysc_node`/`lysp_node` alias for the structure (instead of explicit cast).

This change affects API!

Fixes #1351 
Fixes #1366


rkrejci:
@rwestphal , @choppsv1 - I would appreciate if you could review the changes (mainly the `tree_schema.h`), they are based on your previous feedback

src/tree_schema.h
```diff
@@ -1801,7 +1906,7 @@ const struct lysc_when *lysc_has_when(const struct lysc_node *node);
  * @param[in] node Node to examine.
  * @return The node's groupings sized array if any, NULL otherwise.
```
michalvasko, line 1907: s/sized array/linked list

src/tree_schema.h
```diff
@@ -1817,15 +1922,15 @@ const struct lysp_tpdf *lysp_node_typedefs(const struct lysp_node *node);
  * @param[in] node Node to examine.
  * @return The node's actions/RPCs sized array if any, NULL otherwise.
```
michalvasko, line 1923: 2x s/sized array/linked list

src/tree_schema.h
```diff
@@ -1817,15 +1922,15 @@ const struct lysp_tpdf *lysp_node_typedefs(const struct lysp_node *node);
...
  */
-const struct lysp_action *lysp_node_actions(const struct lysp_node *node);
+const struct lysp_node_action *lysp_node_actions(const struct lysp_node *node);
 
 /**
  * @brief Get the Notifications sized array of the given (parsed) schema node.
  * Decides the node's type and in case it has a Notifications array, returns it.
  * @param[in] node Node to examine.
  * @return The node's Notifications sized array if any, NULL otherwise.
```
michalvasko, line 1931: 2x s/sized array/linked list

src/tree_schema.h
```diff
@@ -1841,47 +1946,25 @@ const struct lysp_node *lysp_node_children(const struct lysp_node *node);
  * @param[in] node Node to examine.
  * @return The node's actions/RPCs sized array if any, NULL otherwise.
```
michalvasko, line 1947: 2x s/sized array/linked list

src/tree_schema.h
```diff
@@ -1841,47 +1946,25 @@ const struct lysp_node *lysp_node_children(const struct lysp_node *node);
...
  */
-const struct lysc_action *lysc_node_actions(const struct lysc_node *node);
+const struct lysc_node_action *lysc_node_actions(const struct lysc_node *node);
 
 /**
  * @brief Get the Notifications sized array of the given (compiled) schema node.
  * Decides the node's type and in case it has a Notifications array, returns it.
  * @param[in] node Node to examine.
  * @return The node's Notifications sized array if any, NULL otherwise.
```
michalvasko, line 1955: 2x s/sized array/linked list

michalvasko (changes requested):
No real problem, looks good.

src/tree_schema.h
rkrejci: fixed

src/tree_schema.h
rkrejci: fixed

src/tree_schema.h
rkrejci: fixed

src/tree_schema.h
rkrejci: fixed

src/tree_schema.h
rkrejci: fixed

src/tree_schema.h
```diff
@@ -1568,69 +1606,166 @@ struct lysc_node {
...
+            const char *name;        /**< Notification name (mandatory) */
+            const char *dsc;         /**< description */
+            const char *ref;         /**< reference */
+            struct lysc_ext_instance *exts; /**< list of the extension instances ([sized array](@ref sizedarrays)) */
+            struct lysc_when **when; /**< list of pointers to when statements ([sized array](@ref sizedarrays)) */
+            void *priv;              /** private arbitrary user data, not used by libyang */
+        };
+    };
+
+    struct lysc_node *data;          /**< first child node (linked list) */
```
michalvasko, line 1632: Rename to `child` to unfiy with the rest of the nodes?

src/tree_schema.h
```diff
@@ -1568,69 +1606,166 @@ struct lysc_node {
...
```
michalvasko, line 1687: Also, rename to `child`?

src/tree_schema.h
rkrejci: yes, make sense not only in compiled nodes but also in the parsed nodes

michalvasko (approved):

src/tree_schema_helpers.c
```diff
@@ -1471,50 +1380,20 @@ lysc_node_children(const struct lysc_node *node, uint16_t flags)
...
-    if (node->nodetype == LYS_RPC || node->nodetype == LYS_ACTION) {
+    if ((node->nodetype == LYS_RPC) || (node->nodetype == LYS_ACTION)) {
         if (flags & LYS_CONFIG_R) {
-            return (struct lysc_node *)&((struct lysc_action *)node)->output;
+            return (struct lysc_node *)&((struct lysc_node_action *)node)->output;
         } else {
             /* LYS_CONFIG_W, but also the default case */
-            return (struct lysc_node *)&((struct lysc_action *)node)->input;
+            return (struct lysc_node *)&((struct lysc_node_action *)node)->input;
         }
```
choppsv1, line 1389: Also I suggest creating type conversion inline functions to avoid these casts in multiple places. This allows dropping in an assert as well e.g.,
```
static inline struct lysc_node_action *
lysc_node_action(struct lysc_node *n)
{
    assert(n->nodetype == LYS_ACTION);
    return (struct lysc_node_action *)n;
}
```
then the above code becomes:
`return &lysc_node_action(node)->input.node;`
which eliminates all casting.

choppsv1 (changes requested):

src/tree_schema_helpers.c
```diff
@@ -1471,50 +1380,20 @@ lysc_node_children(const struct lysc_node *node, uint16_t flags)
...
-{
-    if (!node) {
-        return NULL;
-    }
-
-    if (node->nodetype == LYS_RPC || node->nodetype == LYS_ACTION) {
+    if ((node->nodetype == LYS_RPC) || (node->nodetype == LYS_ACTION)) {
         if (flags & LYS_CONFIG_R) {
-            return (struct lysc_node *)&((struct lysc_action *)node)->output;
+            return (struct lysc_node *)&((struct lysc_node_action *)node)->output;
```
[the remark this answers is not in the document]
rkrejci: yes, this is planned as a subsequent step, there are definitely much more places where the `node` is supposed to be used instead of cast.

src/tree_schema_helpers.c
rkrejci: good point, I like it

src/tree_schema.h
```diff
@@ -1568,69 +1606,166 @@ struct lysc_node {
...
+            const char *ref;         /**< reference */
+            struct lysc_ext_instance *exts; /**< list of the extension instances ([sized array](@ref sizedarrays)) */
+            struct lysc_when **when; /**< list of pointers to when statements ([sized array](@ref sizedarrays)) */
+            void *priv;              /** private arbitrary user data, not used by libyang */
+        };
+    };
+
+    struct lysc_node *child;          /**< first child node (linked list) */
+    struct lysc_must *musts;         /**< list of must restrictions ([sized array](@ref sizedarrays)) */
+};
```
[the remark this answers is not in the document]
rkrejci: The `lysc_node` and `lysp_node` structures are part of the API, so there shouldn't be any changes (each change here means moving to a new SONAME and losing backward compatibility).
Having the common members explicitly even in the derived structures allows providing more precise documentation regarding the content of each member in a specific node type (expected flags, not used members, etc.). The `name` in the augment statement has actually the meaning of `nodeid` (target), which can be explained this way.
I believe that this mess in the definition is worthwhile.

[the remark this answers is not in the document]
choppsv1: Has libyang2 actually been released yet though? That's what I meant about now being the time to change. To your second point about documenting different meanings, which I'm more sympathetic to, if they have different meanings though, they maybe shouldn't be in the common node structure.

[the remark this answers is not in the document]
rkrejci: No, it wasn't released yet, that's why we are able to do this kind of change.

Regarding the different meanings of the members - it is not a common situation, but there are few such cases. The reason why the structures are supposed to be somehow compatible is that they all can appear in schema tree (parsed or compiled), so they are hidden behind the `parent` or `child` pointers. But you are right, that now after unifying all the relevant structures, we should again look at the members if they are really necessary to be in the common structure.

I still believe that better documentation and easier use of the members is sufficient justification for this solution. It seems that there are benefits (more freedom regarding accessing the members, better understandability thanks to docs) for users and the only problem is a little bit more challenging maintenance in case of structure change for us (libyang developers). The first case is supposed to be common with a larger group of affected people, while the second case is rare with only a few affected people.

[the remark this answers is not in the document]
choppsv1: I think you will find that most people are confused by the union. It's sort of abusing the language to make it 2 keystrokes less to type. The documentation for the common fields goes on the common structure and that's where I would look for it. Again, I think you have a workable solution, but I think KISS and "do the obvious thing" would not use a union.

src/schema_compile_amend.c
```diff
@@ -903,11 +870,11 @@ lys_apply_deviate_add(struct lysc_ctx *ctx, struct lysp_deviate_add *d, struct l
             musts = &((struct lysp_node_container *)target)->musts;
             break;
         case LYS_NOTIF:
-            musts = &((struct lysp_notif *)target)->musts;
+            musts = &((struct lysp_node_notif *)target)->musts;
             break;
         case LYS_INPUT:
         case LYS_OUTPUT:
-            musts = &((struct lysp_action_inout *)target)->musts;
+            musts = &((struct lysp_node_action_inout *)target)->musts;
```
michalvasko, line 877: Use the helper function.

src/schema_compile_amend.c
```diff
@@ -1146,11 +1113,11 @@ lys_apply_deviate_delete(struct lysc_ctx *ctx, struct lysp_deviate_del *d, struc
...
```
michalvasko, line 1120: Use the helper function.

src/schema_compile_amend.c
```diff
@@ -1435,7 +1402,7 @@ lysp_schema_nodeid_match_pnode(const struct lysp_node *pnode, const struct lys_m
 
     /* compare names */
     if (pnode->nodetype & (LYS_ACTION | LYS_RPC)) {
-        pname = ((struct lysp_action *)pnode)->name;
+        pname = ((struct lysp_node_action *)pnode)->name;
```
michalvasko, line 1405: The whole `if` construct now redundant.

src/tree_data.c
```diff
@@ -1534,11 +1534,11 @@ lyd_new_implicit_r(struct lyd_node *parent, struct lyd_node **first, const struc
             if (!(iter->flags & LYS_PRESENCE) && lyd_find_sibling_val(*first, iter, NULL, 0, NULL)) {
                 /* create default NP container */
                 LY_CHECK_RET(lyd_create_inner(iter, &node));
-                node->flags = LYD_DEFAULT | (node->schema->when ? LYD_WHEN_TRUE : 0);
+                node->flags = LYD_DEFAULT | (lysc_node_container(node->schema)->when ? LYD_WHEN_TRUE : 0);
```
michalvasko, line 1537: Use helper function instead.

src/tree_data.c
```diff
@@ -1534,11 +1534,11 @@ lyd_new_implicit_r(struct lyd_node *parent, struct lyd_node **first, const struc
...
                 lyd_insert_node(parent, first, node);
 
                 /* cannot be a NP container with when */
-                assert(!iter->when);
+                assert(!lysc_node_container(iter)->when);
```
michalvasko, line 1541: Helper function.

src/tree_data.c
```diff
@@ -1563,10 +1563,10 @@ lyd_new_implicit_r(struct lyd_node *parent, struct lyd_node **first, const struc
                 } else if (ret) {
                     return ret;
                 }
-                node->flags = LYD_DEFAULT | (node->schema->when ? LYD_WHEN_TRUE : 0);
+                node->flags = LYD_DEFAULT | (lysc_node_leaf(node->schema)->when ? LYD_WHEN_TRUE : 0);
                 lyd_insert_node(parent, first, node);
 
-                if (iter->when && node_when) {
+                if (lysc_node_leaf(iter)->when && node_when) {
```
michalvasko, line 1569: Helper function.

src/tree_data.c
```diff
@@ -1591,10 +1591,10 @@ lyd_new_implicit_r(struct lyd_node *parent, struct lyd_node **first, const struc
                     } else if (ret) {
                         return ret;
                     }
-                    node->flags = LYD_DEFAULT | (node->schema->when ? LYD_WHEN_TRUE : 0);
+                    node->flags = LYD_DEFAULT | (lysc_node_leaflist(node->schema)->when ? LYD_WHEN_TRUE : 0);
                     lyd_insert_node(parent, first, node);
 
-                    if (iter->when && node_when) {
+                    if (lysc_node_leaflist(iter)->when && node_when) {
```
michalvasko, line 1597: Helper function.

src/validation.c
```diff
@@ -1178,14 +1179,14 @@ lyd_validate_must(const struct lyd_node *node, LYD_VALIDATE_OP op)
...
+        musts = ((struct lysc_node_notif *)node->schema)->musts;
         break;
     case LYS_RPC:
     case LYS_ACTION:
         if (op == LYD_VALIDATE_OP_RPC) {
-            musts = ((struct lysc_action *)node->schema)->input.musts;
+            musts = ((struct lysc_node_action *)node->schema)->input.musts;
         } else if (op == LYD_VALIDATE_OP_REPLY) {
-            musts = ((struct lysc_action *)node->schema)->output.musts;
+            musts = ((struct lysc_node_action *)node->schema)->output.musts;
```
michalvasko, line 1189: Helper function.

michalvasko (changes requested):

michalvasko (approved):

Issue #1391, opened 2021-01-30
Doubt about why continue to parse xpath expression if parsing xpath subexpressions failed 


fredgan:
Hi Michal,

when the functions `_transform_json2xml_subexp` and `transform_xml2json_subexp` (as follows) fails, the current processing is continue to  parse xpath, why not handle  like `goto error`? what is the intention?  

1) https://github.com/CESNET/libyang/blob/6099310f555093e5be7262df6a7f6afda76ba075/src/common.c#L375-L378
2) https://github.com/CESNET/libyang/blob/6099310f555093e5be7262df6a7f6afda76ba075/src/common.c#L552-L555

michalvasko:
It is intentional. Without going into too many details, *libyang* version 1 is not able to always determine whether a `str:str2` is supposed to be interpreted as a value with a prefix or a value with a colon. So it tries to interpret is as the former, if it fails, it goes with the latter. In *libyang* version 2 it is implemented better without the need for such code.

PR #1392, opened 2021-01-30
Bugfix #1391: failure handling if xpath subexpressions parsing failed

fredgan:
test cases may fail, it will be supplemented and modified later


michalvasko:
Explained in #1391.

PR #1412, opened 2021-02-06
data tree BUGFIX handle when in mandatory true node

choppsv1:
when validation creates opaque context node with no schema, then does an lyxp_eval on it which doesn't check for NULL schema. Check for this. Add UT which exposes the problem (and shows fixed).


michalvasko:
Okay, no problem, thanks for the test.

PR #1416, opened 2021-02-08
libyang BUGFIX fix inverted pattern when string data is parsing

radek-isa:


michalvasko (commented):
So the code was completely missing support for inverted match? If so, I would leave that up to @rkrejci. Other than that, there is a really high number of formatting errors considering that so few lines were modified.

src/plugins_types.c
```diff
@@ -506,33 +505,34 @@ ly_type_validate_patterns(struct lysc_pattern **patterns, const char *str, size_
         }
 
         rc = pcre2_match(patterns[u]->code, (PCRE2_SPTR)str, str_len, 0, PCRE2_ANCHORED | PCRE2_ENDANCHORED, match_data, NULL);
-        if (rc == PCRE2_ERROR_NOMATCH) {
-            if (asprintf(&errmsg, "String \"%.*s\" does not conform to the pattern \"%s\".", (int)str_len, str, patterns[u]->expr) == -1) {
+        pcre2_match_data_free(match_data);
+
+        /* ERROR CHECK */
+        /* !(rc == PCRE2_ERROR_NOMATCH || rc >= 0) */
```
rkrejci, line 511: Why don't you reverse the if's branches? Start with a positive result/branch.

src/plugins_types.c
```diff
@@ -506,33 +505,34 @@ ly_type_validate_patterns(struct lysc_pattern **patterns, const char *str, size_
...
+        if ((rc != PCRE2_ERROR_NOMATCH) && (rc < 0)) {
+            PCRE2_UCHAR pcre2_errmsg[LY_PCRE2_MSG_LIMIT] = {0};
+            pcre2_get_error_message(rc, pcre2_errmsg, LY_PCRE2_MSG_LIMIT);
+
+            *err = ly_err_new(LY_LLERR, LY_ESYS, 0, strdup((const char *)pcre2_errmsg), NULL, NULL);
+            return LY_ESYS;
+        }
+        /* MATCH CHECK   */
```
rkrejci, line 519: do not split `} else if ...`

src/plugins_types.c
```diff
@@ -506,33 +505,34 @@ ly_type_validate_patterns(struct lysc_pattern **patterns, const char *str, size_
...
+        else if (((rc == PCRE2_ERROR_NOMATCH) && (patterns[u]->inverted == 0)) ||
+                ((rc != PCRE2_ERROR_NOMATCH) && (patterns[u]->inverted == 1))) {
+            LY_ERR ret = 0;
+            const char *inverted = patterns[u]->inverted == 0 ? " " : " inverted ";
```
rkrejci, line 523: Do you feel that it is necessary to keep it in a variable? I'm not strictly against it, but in such a simple case it seems to be better to put the conditional evaluation just as a parameter of the `asprintf()`.

tests/utests/data/test_types.c
```diff
@@ -253,6 +254,17 @@ test_string(void **state)
     TEST_PATTERN_1(tree, "str", 1, STRING, "teststring");
     lyd_free_all(tree);
 
+    /* inverted match */
+    // str-invert {type string {pattern '[a-z ]*' {modifier invert-match;}}}"
```
rkrejci, line 258: do not use `//` comments

rkrejci (changes requested):

rkrejci:
And please reword the commit message - follow the format of other commits

src/plugins_types.c
radek-isa: Because the next condition (rc != PCRE2_ERROR_NOMATCH) && (patterns[u]->inverted == 1) would have to be little more complicated in that case. It would have to be changed to  (rc >= 0) && (patterns[u]->inverted == 1)

src/plugins_types.c
radek-isa: function asprintf take two lines already.  So I wanted expand that command as less as possible.

rkrejci: so you have added the extra line somewhere else to hide a little more the string to be printed :)

src/plugins_types.c
```diff
@@ -506,33 +505,33 @@ ly_type_validate_patterns(struct lysc_pattern **patterns, const char *str, size_
...
+        /* !(rc == PCRE2_ERROR_NOMATCH || rc >= 0) */
+        if ((rc != PCRE2_ERROR_NOMATCH) && (rc < 0)) {
+            PCRE2_UCHAR pcre2_errmsg[LY_PCRE2_MSG_LIMIT] = {0};
+            pcre2_get_error_message(rc, pcre2_errmsg, LY_PCRE2_MSG_LIMIT);
+
+            *err = ly_err_new(LY_LLERR, LY_ESYS, 0, strdup((const char *)pcre2_errmsg), NULL, NULL);
+            return LY_ESYS;
+        } else if (((rc == PCRE2_ERROR_NOMATCH) && (patterns[u]->inverted == 0)) ||
+                  ((rc != PCRE2_ERROR_NOMATCH) && (patterns[u]->inverted == 1))) {
+            /* MATCH CHECK   */
```
rkrejci, line 520: I don't know - isn't it confusing? It is not the check, it is already known that the pattern doesn't match. Actually both the `.. CHECK */` comments seem like
```
i++; /* increment counter */
```

src/plugins_types.c
```diff
@@ -506,33 +505,33 @@ ly_type_validate_patterns(struct lysc_pattern **patterns, const char *str, size_
         }
 
         rc = pcre2_match(patterns[u]->code, (PCRE2_SPTR)str, str_len, 0, PCRE2_ANCHORED | PCRE2_ENDANCHORED, match_data, NULL);
-        if (rc == PCRE2_ERROR_NOMATCH) {
-            if (asprintf(&errmsg, "String \"%.*s\" does not conform to the pattern \"%s\".", (int)str_len, str, patterns[u]->expr) == -1) {
+        pcre2_match_data_free(match_data);
+
+        /* ERROR CHECK */
+        /* !(rc == PCRE2_ERROR_NOMATCH || rc >= 0) */
```
rkrejci, line 511: Why do you have this comment here? If you think that this is a more readable representation, use it (don't try to outfox the compiler, it will optimize it). Why there should be both conditions?

rkrejci (changes requested):

rkrejci (approved):

PR #1417, opened 2021-02-08
schema REFACTOR compiling nodes' flags

rkrejci:
Avoid mixing compilation (setting up) of different node's flags and
clarify the purpose and use of several compilation options.

New compilation flag LYS_COMPILE_NO_CONFIG was added to clearly avoid
processing the config flags. So far it was indirectly deduced from
getting know it is compiling Notification/Action/RPC. However it will
be useful to switch off the config flags processing also in other
situations (YANG extensions), so keep it separated.

### DIFF
--- a/src/schema_compile.h
+++ b/src/schema_compile.h
@@ -37,17 +37,19 @@ struct lyxp_expr;
  *
  * @{
  */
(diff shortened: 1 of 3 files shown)


michalvasko (approved):
Just one general question, should `config` be always ignored in an extension?

rkrejci:
no, but e.g. yang-data requires it, so I will have to figure out the way to set the compilation flag for it

PR #1421, opened 2021-02-16
Libyang2 BUGFIX unification of relasing values in plugins store function

radek-isa:


src/plugins_types.h
```diff
@@ -181,6 +181,9 @@ void lysc_prefixes_free(struct lysc_prefix *prefixes);
  * must function as expected.
  *
  * Note that the \p value string is not necessarily zero-terminated. The provided \p value_len is always correct.
+ * Note that dynamically allocated value can be added to dictonary by function zero copy. This means that all store function
```
michalvasko, line 184: Not sure if this note is really needed, basically the same text is in the description of `LY_TYPE_STORE_DYNAMIC`.

src/plugins_types.h
```diff
@@ -181,6 +181,9 @@ void lysc_prefixes_free(struct lysc_prefix *prefixes);
...
+ * have to free dynamically allocated value in case they dont used it. Also, in case of any error store function have to call free
```
michalvasko, line 185: I suppose this is up for discussion but I always try to stick to the rule that if a function fails, it should not have any side-effect, if possible. That is why inserting the value into dictionary should be the last thing they perform. But this should be mentioned in the description.

radek-isa: Then in original ligyang2 is two problems. First, if adding value to the dictionary by zero-copy function fails then the function release value "**https://github.com/CESNET/libyang/blob/libyang2/src/hash_table.c#L249-L253**". The second problem is some store function somewhere in the middle call function and add value to the dictionary. After that, they check if the value is valid. 

I'm not against it. It depends How we decide. In case when not functions are not going to releasing variable if an error occurs. I would have to rewrite plugins store functions.

src/plugins_types.c
```diff
@@ -914,18 +931,24 @@ ly_type_store_binary(const struct ly_ctx *ctx, const struct lysc_type *type, con
...
         struct ly_err_item **err)
 {
+    LY_ERR ret = LY_SUCCESS;
     size_t start = 0, stop = 0, count = 0, u, termination = 0;
     struct lysc_type_bin *type_bin = (struct lysc_type_bin *)type;
     char *errmsg;
     int rc = 0;
 
-    LY_CHECK_ARG_RET(ctx, value, LY_EINVAL);
+    if (value == NULL) {
```
rkrejci, line 940: This change doesn't make sense - if the `value` is NULL, you are goto to free it.

src/plugins_types.c
```diff
@@ -980,21 +1003,23 @@ ly_type_store_binary(const struct ly_ctx *ctx, const struct lysc_type *type, con
...
+        ret = lydict_insert(ctx, &value[start], stop + 1 - start, &storage->canonical);
     } else {
-        LY_CHECK_RET(lydict_insert(ctx, value_len ? value : "", value_len, &storage->canonical));
+        ret = lydict_insert(ctx, value_len ? value : "", value_len, &storage->canonical);
     }
+    LY_CHECK_GOTO(ret != LY_SUCCESS, clean_up_value);
     storage->ptr = NULL;
     storage->realtype = type;
 
     if (options & LY_TYPE_STORE_DYNAMIC) {
```
rkrejci, line 1019: `goto cleanup_value` (maybe with success comment) to avoid code duplication

src/plugins_types.c
```diff
@@ -1335,29 +1398,40 @@ ly_type_store_empty(const struct ly_ctx *ctx, const struct lysc_type *type, cons
...
+            ret = LY_EVALID;
         }
+        goto clean_up;
     }
 
     if (options & LY_TYPE_STORE_DYNAMIC) {
-        LY_CHECK_RET(lydict_insert_zc(ctx, (char *)value, &storage->canonical));
+        ret = lydict_insert_zc(ctx, (char *)value, &storage->canonical);
+        options &= ~LY_TYPE_STORE_DYNAMIC;
+        /* LY_CHECK_GOTO(ret != LY_SUCCESS, clean_up); */
```
rkrejci, line 1422: seem like it is not finished - move the rest of the `storage` filling before storing the value or uncomment those `LY_CHECK_GOTO`s

src/plugins_types.h
rkrejci: I wouldn't connect the note with the dictionary and its zero copy functionality, but the general note regarding the expected behavior of the callback seems useful to me.

src/plugins_types.h
rkrejci: I agree with @radek-isa and tend to agree with this change. Seems to me that only this way it is always clear what the caller should do with the value (it is unknown in which point the callback failed), so I would prefer to accept this change.

rkrejci (changes requested):

src/plugins_types.h
michalvasko: My idea was, of course, that is the callback fails, the value is unchanged, if it succeeds and the dynamic flag was used, the value was used and should not be freed (so the caller always knows what to do). I have just mentioned what is the best approach in my experience (I do not find freeing the value on error as the logical thing to do in general) but I do not really have anything against this solution, so fine.

src/parser_xml.c
```diff
@@ -469,6 +469,14 @@ lydxml_subtree_r(struct lyd_xml_ctx *lydctx, struct lyd_node *parent, struct lyd
         assert(lydctx->parse_opts & LYD_PARSE_OPAQ);
 
         if (xmlctx->ws_only) {
+            /* White string only value should be staticaly allocated.
```
michalvasko, line 472: Why? The whitespaces may be relevant, the flag is there just for cases when it is not, for a quick check. If you have a `string` value, for example, all the exact whitespace values must be preserved, this change just throws them all away.

src/plugins_types.c
```diff
@@ -914,97 +932,113 @@ ly_type_store_binary(const struct ly_ctx *ctx, const struct lysc_type *type, con
...
 
-    /* length of the encoded string */
+    /* check if value meets the type requirments */
     if (type_bin->length) {
-        char buf[LY_NUMBER_MAXLEN];
-        /* get correct length based on base64 encoding rules */
-        uint64_t len = ((count / 4) * 3) - termination;
-        snprintf(buf, LY_NUMBER_MAXLEN, "%" PRIu64, len);
-        LY_CHECK_RET(ly_type_validate_range(LY_TYPE_BINARY, type_bin->length, len, buf, err));
+        const unsigned value_length = ((base64_count + base64_terminated) / 4) * 3 - base64_terminated;
```
michalvasko, line 1007: Please do not use `unsigned` or `int unsigned` or even `unsigned int`, they have variable length which can cause problems. Just use `uint32_t` instead.

src/plugins_types.c
```diff
@@ -1263,23 +1314,36 @@ ly_type_store_enum(const struct ly_ctx *ctx, const struct lysc_type *type, const
...
-        LY_CHECK_RET(lydict_insert(ctx, value_len ? value : "", value_len, &storage->canonical));
+        ret = lydict_insert(ctx, value_len ? value : "", value_len, &storage->canonical);
+        LY_CHECK_GOTO(ret != LY_SUCCESS, cleanup);
     }
     storage->enum_item = &type_enum->enums[u];
     storage->realtype = type;
 
+    if (options & LY_TYPE_STORE_DYNAMIC) {
+        free((char *) value);
+    }
```
michalvasko, line 1329: I would prefer `goto cleanup;` here but that is just a minor issue.

src/plugins_types.c
```diff
@@ -1669,6 +1766,11 @@ ly_type_store_instanceid(const struct ly_ctx *ctx, const struct lysc_type *type,
             *err = ly_err_new(LY_LLERR, LY_EVALID, LYVE_DATA, errmsg, NULL, NULL);
         }
     }
+clean_value:
```
michalvasko, line 1769: Rename to `cleanup_value` to be consistent with the other name.

src/plugins_types.h
michalvasko: Yes, I was referring to the first sentence. The other one is fine.

michalvasko (changes requested):

src/parser_xml.c
radek-isa: The white spaces are irrelevant in this case because line 481 changes the value to the empty value anyway. Problems occur when a dynamically allocated string containing only white spaces. If such a string is created then for some reason double free happens. The reason of double free problem is changing the value in some functions to an empty string and don't expecting the value could be dynamically allocated. For example function
 **https://github.com/CESNET/libyang/blob/libyang2/src/tree_data.c#L725-L727**
 **https://github.com/CESNET/libyang/blob/libyang2/src/tree_data.c#L2405-L2407**
 Luckily this cannot happen now in libyang2 because strings with only white characters are statically allocated.
This was a more simple way how to prevent this potential bug.

michalvasko: Oh, sorry, I missed the whole context of this code. But I do not like comment, seems more misleading than helpful. If you just put all the code after the `/* ignore WS-only value */`, it should be fine.

src/plugins_types.c
```diff
@@ -1638,38 +1727,25 @@ ly_type_store_instanceid(const struct ly_ctx *ctx, const struct lysc_type *type,
...
-
+    LY_CHECK_GOTO(ret = lydict_insert_zc(ctx, str, &storage->canonical), cleanup);
     storage->realtype = type;
 
     /* cleanup */
+cleanup:
     lyxp_expr_free(ctx, exp);
     ly_path_free(ctx, path);
 
+clean_value:
```
michalvasko, line 1738: s/clean_value/cleanup_value

michalvasko (changes requested):

src/plugins_types.c
radek-isa: Thanks. I have forgotten about that when I made all changes.

michalvasko (approved):

rkrejci (approved):

PR #1423, opened 2021-02-17
Opaque fixes

michalvasko:


src/common.h
```diff
@@ -262,12 +262,18 @@ size_t LY_VCODE_INSTREXP_len(const char *str);
 
 #define LY_VCODE_DEV_NOT_PRESENT LYVE_REFERENCE, "Invalid deviation %s \"%s\" property \"%s\" which is not present."
 
-#define LY_VCODE_NOWHEN         LYVE_DATA, "When condition \"%s\" not satisfied."
+/* RFC 7950 section 15 errors (errmsg is used in type plugin validation callbacks) */
+#define LY_VCODE_NOUNIQ         LYVE_DATA, "Unique data leaf(s) \"%s\" not satisfied in \"%s\" and \"%s\"."
+#define LY_VCODE_NOMAX          LYVE_DATA, "Too many \"%s\" instances."
+#define LY_VCODE_NOMIN          LYVE_DATA, "Too few \"%s\" instances."
 #define LY_VCODE_NOMUST         LYVE_DATA, "Must condition \"%s\" not satisfied."
+#define LY_ERRMSG_NOLREF_VAL    /* LYVE_DATA */ "Invalid leafref value \"%s\" - no target instance \"%s\" with the same value."
```
rkrejci, line 270: in the end, I see the reason for this, but probably only because the PR highlights the changes for me. However, I'm not sure that this solution improves the readability of the code. Is it necessary to have all that error messages defined here? And mix them together?

src/tree_data.h
```diff
@@ -163,7 +163,9 @@ struct lysc_type;
  * - ::lyd_new_list2()
  * - ::lyd_new_any()
  * - ::lyd_new_opaq()
+ * - ::lyd_new_opaq2()
```
rkrejci, line 166: also the  transition manual needs an update

rkrejci (changes requested):

src/common.h
michalvasko: Well, it is not necessary but I do think it is a good idea :) I have done the work, looked into the specification and put all the errors here in the defined order, if it stays this way this work will not have to be repeated. So I would prefer to have all of these together but I am not saying it cannot be improved, I am just not sure how.

src/tree_data.h
michalvasko: Okay, it should be fixed now but I do not really get the added value of the changes I made.

src/common.h
rkrejci: what about putting them all there as `LY_ERRMSG_...` and keep `LY_VCODE_` outside and just use some of the messages? Just that mixing of different macros seems weird to me.

michalvasko: I would actually prefer to not have any `ERRMSG` because then the coupling of `LY_VCODE` and the message would be strictly required. If the problem is just that the two kinds are mixed, it should not be a problem to put all `LY_VCODE` first, then skip a line, and then put `LY_ERRMSG`.

PR #1427, opened 2021-02-22
adds fuzzer for oss-fuzz integration.

DavidKorczynski:
Hi Maintainers,

I have worked a bit on setting fuzzing up for libyang by way of OSS-Fuzz. Essentially, OSS-Fuzz is a free service run by Google that performs continuous fuzzing of important open source projects, and it would be great to have libyang integrated. The only expectation of integrating into OSS-Fuzz is that bugs will be fixed. This is not a "hard" requirement in that no one enforces this and the main point is if bugs are not fixed then it is a waste of resources to run the fuzzers, which we would like to avoid.

In this PR https://github.com/google/oss-fuzz/pull/5227 I have done exactly that, namely created the necessary logic from an OSS-Fuzz perspective to integrate libyang. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports, coverage reports and more stats. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

At the moment I stored the fuzzers in the Google OSS-Fuzz repository, but we can move them up here to make maintenance easier.

### DIFF
--- a/tests/fuzz/oss-fuzz/fuzz_xml_parse.c
+++ b/tests/fuzz/oss-fuzz/fuzz_xml_parse.c
@@ -1,0 +1,29 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "libyang.h"
+
+int LLVMFuzzerTestOneInput(const char *data, size_t size) {
+    struct ly_ctx *ctx = NULL;
+
+    char filename[256];
+    sprintf(filename, "/tmp/libfuzzer.%d", getpid());
+
+    FILE *fp = fopen(filename, "wb");
+    if (!fp) {
+        return 0;
+    }
+    fwrite(data, size, 1, fp);
+    fclose(fp);
+
+    ctx = ly_ctx_new(NULL, 0);
+    if (!ctx) {
+        return 0;
+    }
+    lyxml_parse_path(ctx, filename, LYS_IN_YANG);
+    ly_ctx_clean(ctx, NULL);
+    ly_ctx_destroy(ctx, NULL);
+
+    unlink(filename);
+    return 0;
+}


rkrejci:
Hi and thanks for your activity. However, since we are just before finishing libyang 2.0 (branch libyang2) it makes more sense to me to focus on new libyang with this and don't bother with it in libyang 1.x. @jvijtiuk, does it make sense to connect this somehow with your fuzzing tests in libyang2?

DavidKorczynski:
Ah, I wasn't aware of that. I am happy to connect the libyang 2.0 branch to OSS-Fuzz with the existing fuzzers you have in there! I think that would probably be the preferred way. Let me know.

jvijtiuk:
Hello,

>  @jvijtiuk, does it make sense to connect this somehow with your fuzzing tests in libyang2?

Sounds good to me. I think that the `parse_mem` harnesses would be the best ones to integrate with OSS-fuzz.
They are available [here](https://github.com/CESNET/libyang/tree/libyang2/tests/fuzz). I've been building them as part of libyang, but they should be buildable as standalone executables. There's also a README in there, and a `main.c` which wraps the harnesses so that they can be used with both AFL and LibFuzzer.

The other harnesses (`buf_add_char` and `yang_parse_module`) fuzz internal libyang functions, and can have too much false positives from my experience, so I don't think those would be worth integrating into OSS-fuzz.

DavidKorczynski:
> Hello,
> 
> > @jvijtiuk, does it make sense to connect this somehow with your fuzzing tests in libyang2?
> 
> Sounds good to me. I think that the `parse_mem` harnesses would be the best ones to integrate with OSS-fuzz.
> They are available [here](https://github.com/CESNET/libyang/tree/libyang2/tests/fuzz). I've been building them as part of libyang, but they should be buildable as standalone executables. There's also a README in there, and a `main.c` which wraps the harnesses so that they can be used with both AFL and LibFuzzer.
> 
> The other harnesses (`buf_add_char` and `yang_parse_module`) fuzz internal libyang functions, and can have too much false positives from my experience, so I don't think those would be worth integrating into OSS-fuzz.

Awesome! Which email(s) should I use for the bug reports then?

michalvasko:
> Which email(s) should I use for the bug reports then?

Just mine mvasko@cesnet.cz should be enough.

rkrejci:
So I'm closing the PR and the new one should target the `libyang2` branch, thanks.

PR #1432, opened 2021-02-24
data validation BUGFIX handling when in default nodes

rkrejci:
When parsing data, the when in implicitely added default nodes were not
taken into the validation process.


michalvasko (approved):
I don't particularly like the long line but there are not many ways of improving it.

PR #1433, opened 2021-02-24
Extension instance parsing

michalvasko:


src/parser_yin.c
```diff
@@ -3144,14 +3104,29 @@ yin_parse_element_generic(struct lys_yin_parser_ctx *ctx, enum ly_stmt parent, s
...
     LY_CHECK_ERR_GOTO(!(*element), LOGMEM(ctx->xmlctx->ctx); ret = LY_EMEM, cleanup);
-    (*element)->stmt = name2nsname(ctx, ctx->xmlctx->name, ctx->xmlctx->name_len, ctx->xmlctx->prefix, ctx->xmlctx->prefix_len);
-    LY_CHECK_ERR_GOTO(!(*element)->stmt, ret = LY_EMEM, cleanup);
+
+    /* store identifier */
+    if (ctx->xmlctx->prefix_len) {
+        if (asprintf(&id, "%.*s:%.*s", (int)ctx->xmlctx->prefix_len, ctx->xmlctx->prefix, (int)ctx->xmlctx->name_len,
+                ctx->xmlctx->name) == -1) {
+            LOGMEM(ctx->xmlctx->ctx);
+            return LY_EMEM;
```
rkrejci, line 3120: It would be nice to have it unified - if the rest of the function uses goto, then it's weird to use return here

src/plugins_types.c
```diff
@@ -101,7 +102,12 @@ ly_xml_resolve_prefix(const struct ly_ctx *ctx, const char *prefix, size_t prefi
         return NULL;
     }
 
-    return ly_ctx_get_module_implemented_ns(ctx, ns->uri);
+    mod = ly_ctx_get_module_implemented_ns(ctx, ns->uri);
+    if (!mod) {
```
rkrejci, line 106: is there a possibility to have it NULL even in another case than YIN extension prefix resolution? For example in case of some invalid data and using another way then would actually cause accepting the invalid input referencing some unexpected module here?

rkrejci (changes requested):

src/plugins_types.c
michalvasko: I believe the only new situation that can happen is that a non-implemented module is returned but since `ly_schema_resolve_prefix` can do that, too, it should not be a problem.

src/parser_yin.c
michalvasko: Should be fixed.

rkrejci (approved):

PR #1434, opened 2021-02-25
yang-data extension support

rkrejci:
The set of patches finishes the statement parser for schema and add support for yang-data extension defined in RFC 8040. 

There are also some changes in extensions API, but the API itself is still target to cleanup and finish in some future PR.

### DIFF
--- a/src/parser_data.h
+++ b/src/parser_data.h
@@ -58,8 +58,11 @@ struct ly_in;
  *     example the *:operational* datastore is not necessarily valid and results of the NETCONF's \<get\> or \<get-config\>
  *     oprations used with filters will be incomplete (and thus invalid). This can be allowed using ::LYD_PARSE_ONLY,
  *     the ::LYD_PARSE_NO_STATE should be used for the data returned by \<get-config\> operation.
+ * - ::lyd_parse_ext_data() is used for parsing configuration data trees defined inside extension instances, such as
+ *   instances of yang-data extension specified in [RFC 8040](http://tools.ietf.org/html/rfc8040).
  * - ::lyd_parse_op() is used for parsing RPCs/actions, replies, and notifications. Even NETCONF rpc, rpc-reply, and
- *     notification messages are supported.
+ *   notification messages are supported.
+ * - ::lyd_parse_ext_op() is used for parsing RPCs/actions, replies, and notifications defined inside extension instances.
  *
  * Further information regarding the processing input instance data can be found on the following pages.
  * - @subpage howtoDataValidation
@@ -71,7 +74,9 @@ struct ly_in;
  * - ::lyd_parse_data_mem()
  * - ::lyd_parse_data_fd()
  * - ::lyd_parse_data_path()
+ * - ::lyd_parse_ext_data()
  * - ::lyd_parse_op()
+ * - ::lyd_parse_ext_op()
  */
 
 /**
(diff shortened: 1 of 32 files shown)


src/tree_schema.h
```diff
@@ -250,38 +250,85 @@ struct lyxp_expr;
 
 #define LYS_NODETYPE_MASK 0xffff  /**< Mask for nodetypes, the value is limited for 16 bits */
 
+/**
+ * @brief Test for operation (RPC or Action) statements
+ */
+#define LY_STMT_IS_OP(STMT) (((STMT) == LY_STMT_ACTION) || ((STMT) == LY_STMT_RPC))
+
+/**
+ * @brief Test for schema node (anydata, anyxml, choice, container, leaf, leaf-list and list) statements.
```
michalvasko, line 259: I am missing `case` here and even after seeing the comment with it just below the reason is not much clearer. If this is supposed to be used only for top-level nodes, it should be mentioned. But then why there is no `uses`, for example?

src/tree_schema.c
```diff
@@ -215,12 +215,14 @@ lys_getnext_into_case(const struct lysc_node_case *first_case, const struct lysc
 }
 
 API const struct lysc_node *
-lys_getnext(const struct lysc_node *last, const struct lysc_node *parent, const struct lysc_module *module, uint32_t options)
+lys_getnext_(const struct lysc_node *last, const struct lysc_node *parent, const struct lysc_module *module,
```
michalvasko, line 218: I am guessing this should be a static function now.

src/plugins_exts_yangdata.h
```diff
@@ -0,0 +1,28 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* LY_PLUGINS_EXTS_YANGDATA_H_ */
```
michalvasko, line 28: There will be some future use for this, I guess? Otherwise I do not really understand the use.

src/tree_data.c
```diff
@@ -538,6 +551,26 @@ lyd_parse_op(const struct ly_ctx *ctx, struct lyd_node *parent, struct ly_in *in
...
+lyd_parse_op(const struct ly_ctx *ctx, struct lyd_node *parent, struct ly_in *in, LYD_FORMAT format,
+        enum lyd_type data_type, struct lyd_node **tree, struct lyd_node **op)
+{
+    LY_CHECK_ARG_RET(ctx, ctx || parent, in, data_type, parent || tree || op, LY_EINVAL);
+
+    return lyd_parse_op_(ctx, NULL, parent, in, format, data_type, tree, op);
+}
+
+API LY_ERR
+lyd_parse_ext_op(const struct lysc_ext_instance *ext, struct lyd_node *parent, struct ly_in *in, LYD_FORMAT format,
```
michalvasko, line 564: Missing in a public header if I did not miss it.

src/parser_data.h
```diff
@@ -251,6 +254,23 @@ LY_ERR lyd_parse_data_fd(const struct ly_ctx *ctx, int fd, LYD_FORMAT format, ui
...
+ * @param[in] parent Optional parent to connect the parsed nodes to.
+ * @param[in] in The input handle to provide the dumped data in the specified @p format to parse (and validate).
+ * @param[in] format Format of the input data to be parsed. Can be 0 to try to detect format from the input handler.
+ * @param[in] parse_options Options for parser, see @ref dataparseroptions.
+ * @param[in] validate_options Options for the validation phase, see @ref datavalidationoptions.
+ * @param[out] tree Full parsed data tree, note that NULL can be a valid tree. If @p parent is set, set to NULL.
+ * @return LY_SUCCESS in case of successful parsing (and validation).
+ * @return LY_ERR value in case of error. Additional error information can be obtained from the context using ly_err* functions.
+ */
+LY_ERR lyd_parse_ext_data(const struct lysc_ext_instance *ext, struct lyd_node *parent, struct ly_in *in, LYD_FORMAT format,
```
michalvasko, line 271: Just to clarify, you can use this function only for extensions that define strictly separate data trees, which need to be specified explicitly so as not to confuse the extension data with normal module data.

michalvasko (changes requested):

src/plugins_exts_yangdata.h
rkrejci: actually no, there was use, but later it became useless - removing it.

src/tree_schema.c
rkrejci: yes, and description is missing

src/tree_schema.h
rkrejci: It's because the current use of this macro, which comes from yang-data needs (no `uses` because it is limited to compiled statements, no `case` since yang-data does not allow it and it is not clear how to actually handle cases in top level of extension). You are right that it is supposed to be more generic so it should be reflected. I'll try to improve the macro as well as its description.

src/tree_data.c
rkrejci: good catch

src/parser_data.h
rkrejci: ok, I'll add it

michalvasko (approved):

PR #1437, opened 2021-02-26
yang parser BUGFIX inheriting the "config" value

### LINKED ISSUE #1429: deviations can produce a model where child nodes of config: false are config: true
I'm trying to modify an entire model to make it read-only, like this:
### `example.yang`
```yang
module example {
  namespace "x:example:example";
  prefix "ex";

  container a {
    leaf aa {
      type uint8;
    }
  }
}
```
### `aug.yang`
```yang
module aug {
  namespace "x:example:aug";
  prefix "aug";

  import example {
    prefix "ex";
  }

  deviation "/ex:a" {
    deviate add {
      config false;
    }
  }
}
```
However, when I look at the result, I see that `/ex:a/aa` is marked as a config node, while `/ex:a` is read-only.
```console-session
$ yanglint -f tree aug.yang example.yang 
module: aug

module: example
  +--ro a
     +--rw aa?   uint8
```
That is in violation of [RFC 7950](https://tools.ietf.org/html/rfc7950#page-135) which says:
> If a node has "config" set to "false", no node underneath it can have "config" set to "true".

Discovered by @peckato1.

lePici:
Inheriting the "config" value due to "deviate" now fixes #1429.

### DIFF
--- a/src/parser_yang.c
+++ b/src/parser_yang.c
@@ -4499,6 +4499,9 @@ yang_check_deviate(struct lys_module *module, struct unres_schema *unres, struct
     }
 
     if ((deviate->flags & LYS_CONFIG_MASK)) {
+        struct lys_node *parent;
+        const struct lys_node_list *node_list;
+
         /* cannot add if it was explicitly set */
         if ((deviate->mod == LY_DEVIATE_ADD) && (dev_target->flags & LYS_CONFIG_SET)) {
             LOGVAL(module->ctx, LYE_INSTMT, LY_VLOG_NONE, NULL, "config");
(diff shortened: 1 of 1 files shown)


src/parser_yang.c
```diff
@@ -4516,6 +4519,64 @@ yang_check_deviate(struct lys_module *module, struct unres_schema *unres, struct
...
+        /* "config" must be set exclusively to "yes" or "no" */
+        assert(!((dev_target->flags & LYS_CONFIG_W) && (dev_target->flags & LYS_CONFIG_R)));
+
+        /* from rfc7950#page-86: All key leafs in a list MUST have the same value
+         * for their "config" as the list itself.
+         */
+        if ((node_list = lys_is_key((const struct lys_node_leaf *)dev_target, NULL))) {
+            if ((node_list->flags & LYS_CONFIG_MASK) != (dev_target->flags & LYS_CONFIG_MASK)) {
+                LOGVAL(module->ctx, LYE_INSTMT, LY_VLOG_NONE, NULL, "config");
+                LOGVAL(module->ctx, LYE_SPEC, LY_VLOG_NONE, NULL, "The deviate causes a violation of \"config\" in key leaf.");
```
michalvasko, line 4535: I suppose you wanted to keep the message short so fine, but some details could be included. But at least make it `The deviate causes a violation of \"config\" rules in a key leaf.` so that it makes more sense. Similarly for the other cases.

src/parser_yang.c
```diff
@@ -4516,6 +4519,64 @@ yang_check_deviate(struct lys_module *module, struct unres_schema *unres, struct
 
         /* ... and replace it with the value specified in deviation */
         dev_target->flags |= deviate->flags & LYS_CONFIG_MASK;
+
+        /* "config" is explicitely set in the node */
+        dev_target->flags |= LYS_CONFIG_SET;
+
+        /* "config" must be set exclusively to "yes" or "no" */
+        assert(!((dev_target->flags & LYS_CONFIG_W) && (dev_target->flags & LYS_CONFIG_R)));
```
michalvasko, line 4527: You may wan to use the config mask to make this more readable.

michalvasko (changes requested):

src/parser_yang.c
```diff
@@ -4516,6 +4519,64 @@ yang_check_deviate(struct lys_module *module, struct unres_schema *unres, struct
...
+            struct lys_node *next, *elem;
+            LY_TREE_DFS_BEGIN(dev_target->child, next, elem) {
+                if (elem->flags & LYS_CONFIG_W) {
+                    if (elem->flags & LYS_CONFIG_SET) {
+                        LOGVAL(module->ctx, LYE_INSTMT, LY_VLOG_NONE, NULL, "config");
+                        LOGVAL(module->ctx, LYE_SPEC, LY_VLOG_NONE, NULL, "The deviate causes a violation of \"config\" in child node.");
+                        goto error;
+                    } else {
+                        /* remove current config value of the elem including LYS_CONFIG_SET */
+                        elem->flags &= ~(LYS_CONFIG_MASK | LYS_CONFIG_SET) ;
```
lePici, line 4566: Please pay attention here if the **LYS_CONFIG_SET** bit should really be cleared. The point is that if the **config** was explicitly set to **yes** in the node, and at the same time the inheritance from the ancestor was projected, then the **config** is switched to **no**, so such a **config** does not have the value set explicitly, but rather implicitly. Therefore, the flag is deleted.

jktjkt: I think it is a good habit to capture these possible gotchas as comments in the code rather that explanatory comments in the code review. The latter tends to get lost quicker.

src/parser_yang.c
```diff
@@ -4516,6 +4519,61 @@ yang_check_deviate(struct lys_module *module, struct unres_schema *unres, struct
...
+                    "The \"config\" value in the key leaf must be the same as in the list to which belongs.");
+                goto error;
+            }
+        }
+
+        /* from rfc7950#page-135: If the parent node is a case node,
+         * the value is the same as the case node's parent choice node.
+         */
+        if (dev_target->nodetype == LYS_CASE) {
+            if ((parent = lys_parent(dev_target)) && (parent->nodetype == LYS_CHOICE)) {
```
michalvasko, line 4546: I do not think this can ever be false, should probably be an assert and it will save one `if` nesting.

src/parser_yang.c
```diff
@@ -4516,6 +4519,61 @@ yang_check_deviate(struct lys_module *module, struct unres_schema *unres, struct
...
+            LY_TREE_DFS_BEGIN(dev_target->child, next, elem) {
+                if (elem->flags & LYS_CONFIG_W) {
+                    /* Remove current config value of the elem including LYS_CONFIG_SET.
+                     * Note that LYS_CONFIG_SET is cleared. The point is that if the config was explicitly set to yes in the node,
+                     * and at the same time the inheritance from the ancestor was projected, then the config is switched to no,
+                     * so such a config does not have the value set explicitly, but rather implicitly. Therefore, the flag is deleted.
+                     */
+                    elem->flags &= ~(LYS_CONFIG_MASK | LYS_CONFIG_SET) ;
+                }
+                /* assurance that the node will have "config" set to "no" */
```
michalvasko, line 4571: I do not think this is correct but if nothing else then some clarification would be helpful. If the node had no config set before (i think only RPC/action/notification do not have those), why do you want to set it now? Unless I am missing something, the reason it was not set before is still present so the subtree should be left the way it was.

michalvasko (changes requested):

michalvasko (approved):

PR #1438, opened 2021-03-01
libyang2: json BUGFIX out of bounds write in lyjson_number

jvijtiuk:
Hello,

this PR fixes a 1 byte out of bounds write in `lyjson_number`. 

It counts the base number minus in `num_len` when parsing negative JSON numbers with an exponent and no decimal point in `lyjson_number`.
The PR also adds a regression test case, with the original file found by fuzzing which caused the issue in `lyd_parse_mem_json`.

The issue appears when the `-689e-47` number in the fuzzer generated file is parsed. Positive numbers don't seem to have any issues, as `809e-47` is parsed fine.

I ran the tests after this fix and everything seems to be passing. After the fix valgrind doesn't report the oob anymore. I've also tested by adding a print statement after the number is parsed from JSON, and the parsed values look ok to me.


michalvasko:
Alright, thanks.

PR #1441, opened 2021-03-03
plugin types BUGFIX fix eventual double free when call lydict_insert_…

radek-isa:
…zc fail in store_bists function

### DIFF
--- a/src/plugins_types.c
+++ b/src/plugins_types.c
@@ -1199,9 +1199,9 @@ next:
 
         if (!ws_count && !lws_count && (options & LY_TYPE_STORE_DYNAMIC)) {
             ret = lydict_insert_zc(ctx, (char *)value, &can);
-            LY_CHECK_GOTO(ret, cleanup);
             value = NULL;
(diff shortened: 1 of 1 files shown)


src/plugins_types.c
```diff
@@ -54,15 +54,14 @@ ly_err_msg_create(LY_ERR *err_code, LY_VECODE err_vecode, const char *err_msg, .
...
     int  print_ret;
-
     va_list(print_args);
 
     if (err_msg != NULL) {
         va_start(print_args, err_msg);
         print_ret = vasprintf(&msg, err_msg, print_args);
     }
 
-    if ((print_ret < 0) || (err_msg == NULL)) {
```
rkrejci, line 65: I guess this has nothing to do with the subject. It is another bug - the `print_ret` is uninitiated when `err_msg` is not NULL.

rkrejci (changes requested):

rkrejci (approved):

Issue #1336, opened 2021-01-07
Cleanup YANG extension plugins API


rkrejci:
Review plugins_exts.h and clarify its usage. The API of the internal extensions must be the same as for external plugins (what we need in internal types might need also developers of the external plugins).

PR #1443, opened 2021-03-04
extensions cleanup - part 1

rkrejci:
Several bugfixes and improvements regarding processing YANG extensions. There is still work to do, so this is just a first part of task #1336.


michalvasko (approved):
Seems there is a lot to improve, the changes I understood I liked.

PR #1450, opened 2021-03-06
resolve BUGFIX must validation raises 2 errors if there is an error-message defined in it

fanchanghu:
Hi,
I had created a PR to make libyang support multiple validation errors in #1341. However, I find that the `must` validation will raise 2 errors if there is an error-message defined in it. This PR is to fix this issue.

This issue could be reproduced with following module:
```
list T{
	key f1;
	leaf f1 {
		type string;
	}

	must "./f1='12'"{
		error-app-tag "ttttt";
		error-message "eeeee";
	}
}
```
current libyang raise following 2 errors, while expected is only the last one.
```
1: (null), Must condition "./f1='12'" not satisfied., /A:T[f1='1.0']
2: ttttt, eeeee, /A:T[f1='1.0']
```


michalvasko:
Okay, seems fine to me, merging.

michalvasko (approved):

PR #1456, opened 2021-03-08
plugin REFACTOR put error print message into function ly_err_msg_create

radek-isa:


michalvasko (commented):
Right, this was one of the features I knew was missing, so good that you have added it and it seems fine. However, please also add a test for the apptag being specified in the YANG and returned by some type validation.

michalvasko (approved):

src/plugins_types.c
```diff
@@ -44,12 +44,14 @@
  *
  * @param [in,out] err_code return code. if thre is not enough memory for allocating error message then this code change to LY_EMEM.
  * @param [in] err_vecode Validity error code in case of LY_EVALID error code.
+ * @param [in] path   	Path to the node causing the error.
```
rkrejci: inconsistent indentation, keep there just a single space

rkrejci (changes requested):

src/plugins_types.c
rkrejci: I see that it was changed in a later commit, but I don't see any reason to do not to rebase the fix back into the commit it introduced the spaces. And even after the later change, there are still more spaces.

rkrejci (approved):

PR #1458, opened 2021-03-11
yang extensions API cleanup - part 2

### LINKED ISSUE #1336: Cleanup YANG extension plugins API

rkrejci:
the extensions API should be ready now, but the headers are not installed yet until the mechanism of loading external plugins will be prepared. Also the DOCs will be added together with this mechanism. This should close #1336


src/plugins_exts.c
```diff
@@ -13,19 +13,26 @@
...
+#include "plugins_exts_compile.h"
+#include "plugins_exts_print.h"
 
 #include <stdint.h>
 #include <string.h>
 
-#include "common.h"
 #include "plugins_exts_metadata.c"
 #include "plugins_exts_nacm.c"
 #include "plugins_exts_yangdata.c"
```
michalvasko, line 24: Do not include C files.

michalvasko (changes requested):

src/plugins_exts.c
rkrejci: right, changed

src/tree_data.c
```diff
@@ -1624,9 +1617,10 @@ lyd_new_path_check_find_lypath(struct ly_path *path, const char *str_path, const
     return ret;
 }
 
-API LY_ERR
-lyd_new_path2(struct lyd_node *parent, const struct ly_ctx *ctx, const char *path, const void *value,
-        LYD_ANYDATA_VALUETYPE value_type, uint32_t options, struct lyd_node **new_parent, struct lyd_node **new_node)
+static LY_ERR
```
michalvasko, line 1620: I suppose there should be some doxygen even though it will just be a copy of the API functions.

src/tree_data.c
```diff
@@ -945,6 +945,55 @@ lyd_new_list(struct lyd_node *parent, const struct lys_module *module, const cha
...
+cleanup:
+    va_end(ap);
+    if (rc) {
+        lyd_free_tree(ret);
+        ret = NULL;
+    } else if (node) {
+        *node = ret;
+    }
+    return rc;
+
```
michalvasko, line 994: Extra line.

michalvasko (changes requested):
Generally, seems fine to me.

src/tree_data.c
rkrejci: agreed

src/tree_data.c
rkrejci: thanks, fixed

michalvasko (approved):

PR #1460, opened 2021-03-11
libyang2: json BUGFIX invalid BMP character oob LOGVAL read

jvijtiuk:
Hello,

This PR fixes an issue that was found by OSSFuzz in `lyd_parse_mem_json`.
It caused an oob read in `LOGVAL`, as invalid BMP values were logged using an offset from the `in` pointer, which is moved around while strings are parsed in `lyjson_string_`. Replacing it with the `start` pointer which doesn't move around fixes the issue.

I've also added the file that causes the issue to the fuzz regression tests.

### DIFF
--- a/src/json.c
+++ b/src/json.c
@@ -153,7 +153,7 @@ lyjson_string_(struct lyjson_ctx *jsonctx)
     while (in[offset]) {
         if (in[offset] == '\\') {
             /* escape sequence */
-            size_t slash = offset;
+            const char *slash = &in[offset];
             uint32_t value;
             uint8_t i = 1;
 
@@ -221,7 +221,7 @@ lyjson_string_(struct lyjson_ctx *jsonctx)
                 offset++;
                 for (value = i = 0; i < 4; i++) {
                     if (!in[offset + i]) {
-                        LOGVAL(jsonctx->ctx, LYVE_SYNTAX, "Invalid basic multilingual plane character \"%s\".", &in[slash]);
+                        LOGVAL(jsonctx->ctx, LYVE_SYNTAX, "Invalid basic multilingual plane character \"%s\".", slash);
                         goto error;
                     } else if (isdigit(in[offset + i])) {
                         u = (in[offset + i] - '0');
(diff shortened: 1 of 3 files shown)


michalvasko:
Looking at the code it seems to me it may fix the particular test-case but not many others. Could you please try the patch below? I did not manage to verify it fixes the problem for some reason, thanks.
```patch
diff --git a/src/json.c b/src/json.c
index 2092a095..e460d297 100644
--- a/src/json.c
+++ b/src/json.c
@@ -153,7 +153,7 @@ lyjson_string_(struct lyjson_ctx *jsonctx)
     while (in[offset]) {
         if (in[offset] == '\\') {
             /* escape sequence */
-            size_t slash = offset;
+            const char *slash = &in[offset];
             uint32_t value;
             uint8_t i = 1;
 
@@ -221,7 +221,7 @@ lyjson_string_(struct lyjson_ctx *jsonctx)
                 offset++;
                 for (value = i = 0; i < 4; i++) {
                     if (!in[offset + i]) {
-                        LOGVAL(jsonctx->ctx, LYVE_SYNTAX, "Invalid basic multilingual plane character \"%s\".", &in[slash]);
+                        LOGVAL(jsonctx->ctx, LYVE_SYNTAX, "Invalid basic multilingual plane character \"%s\".", slash);
                         goto error;
                     } else if (isdigit(in[offset + i])) {
                         u = (in[offset + i] - '0');
@@ -243,7 +243,7 @@ lyjson_string_(struct lyjson_ctx *jsonctx)
             offset += i;   /* add read escaped characters */
             LY_CHECK_ERR_GOTO(ly_pututf8(&buf[len], value, &u),
                     LOGVAL(jsonctx->ctx, LYVE_SYNTAX, "Invalid character reference \"%.*s\" (0x%08x).",
-                    (int)(offset - slash), &in[slash], value),
+                    (int)(&in[offset] - slash), slash, value),
                     error);
             len += u;      /* update number of bytes in buffer */
             in += offset;  /* move the input by the processed bytes stored in the buffer ... */
```

jvijtiuk:
> Looking at the code it seems to me it may fix the particular test-case but not many others. 

You are correct, I've managed to make an input that triggers a second version of the issue with my initial patch applied. The patch you provided fixes that issue as well. I've added that second file to the fuzz regression tests as well and force pushed that with the updated patch.

The second issue was triggered in the `LY_CHECK_ERR_GOTO` statement

> Could you please try the patch below? I did not manage to verify it fixes the problem for some reason, thanks.

Sorry, I forgot to mention that it doesn't crash libyang, but only causes invalid reads which are reported by valgrind or sanitizers.
I verified, and the new patch fixes both issues.

michalvasko:
> Sorry, I forgot to mention that it doesn't crash libyang, but only causes invalid reads which are reported by valgrind or sanitizers.

Right, I could have checked that. Fine then, merging.

PR #1463, opened 2021-03-16
libyang2 type plugins api

### LINKED ISSUE #1332: Cleanup type plugins API
Review `plugins_types.h` and clarify its usage. The API of the internal types must be the same as for external plugins (what we need in internal types might need also developers of the external plugins).

rkrejci:
Split `plugins_types.c` into separated source files of each internal plugin and make all the plugins use only the functions from libyang public (the `plugins_types.h` is not yet installed, but it will be soon) API.

Fixes #1332


rkrejci:
The types plugins API is not finished, but the only thing to do is the list of available plugins and it will probably change with the mechanism of loading external plugins, so I will do them together.

src/plugins_types.h
```diff
@@ -101,6 +101,24 @@ struct ly_err_item *ly_err_new(LY_LOG_LEVEL level, LY_ERR code, LY_VECODE vecode
  */
 void ly_err_free(void *ptr);
 
+/**
+ * @brief Check that the type is suitable for the parser's hit (if any) in the specified format
```
michalvasko, line 110: s/hit/hint

src/log.c
```diff
@@ -119,23 +119,36 @@ ly_errapptag(const struct ly_ctx *ctx)
 }
 
 API struct ly_err_item *
-ly_err_new(LY_LOG_LEVEL level, LY_ERR no, LY_VECODE vecode, char *msg, char *path, char *apptag)
+ly_err_new(LY_ERR *ecode, LY_VECODE vecode, char *path, char *apptag, const char *err_msg, ...)
```
michalvasko, line 122: Interesting solution for the memory allocation problem but is it really worth it since it will, well, never happen? Meaning remove this parameters and if it does happen, the return `LY_ERR` value will not match the error in this structure?

src/plugins_types.h
```diff
@@ -77,19 +77,21 @@ struct lysp_module;
...
+ * @param [in] vecode Validity error code in case of LY_EVALID error code.
+ * @param [in] path Path to the node causing the error.
+ * @param [in] apptag Error-app-tag value.
+ * @param [in] err_msg error formating message.
  * @return NULL in case of memory allocation failure.
- * @return Created error information structure that can be freed using ::ly_err_free().
+ * @return Created error information structure that can be freed using ::ly_err_free() or passed back from callback into libyang.
  */
-struct ly_err_item *ly_err_new(LY_LOG_LEVEL level, LY_ERR code, LY_VECODE vecode, char *msg, char *path, char *apptag);
+struct ly_err_item *ly_err_new(LY_ERR *ecode, LY_VECODE vecode, char *path, char *apptag, const char *err_msg, ...);
```
michalvasko, line 97: I suggest the same simplification as for the API function.

src/plugins_types.c
```diff
@@ -1755,6 +1702,11 @@ extern LY_ERR ly_type_store_empty(const struct ly_ctx *ctx, const struct lysc_ty
         const struct lysc_node *ctx_node, struct lyd_value *storage, struct lys_glob_unres *unres, struct ly_err_item **err);
 extern LY_ERR ly_type_compare_empty(const struct lyd_value *val1, const struct lyd_value *val2);
 
+/* plugins_types_enumeration.c */
+extern LY_ERR ly_type_store_enum(const struct ly_ctx *ctx, const struct lysc_type *type,
```
michalvasko, line 913: I do not like all these externs much but I suppose it saves a bunch of headers (there could be one for all the internal types, actually) so whatever. But what would make more sense to me is define the whole plugin structure (an item in `ly_builtin_type_plugins`) in the plugin C file and make only that extern.

src/plugins_types.h
```diff
@@ -129,15 +129,16 @@ LY_ERR ly_type_check_hints(uint32_t hints, const char *value, size_t value_len,
  * Use only in implementations of ::ly_type_store_clb which provide all the necessary parameters for this function.
  *
  * @param[in] ctx libyang context.
+ * @param[in] ctx_node Context node to determine the module if needed (unprefixed data in specific @p format)
```
michalvasko, line 133: I see, this may just be a working solution even for non-prefixed node names. But it needs much more format-specific details because `ctx_node` is confusing and even incorrect, I think.

For schema formats, I would refer to it as *current node* because that would correspond to how it is called in YANG RFC. It is the schema node closest to the statement, whatever it is. And it must always be defined in the same module that the statement is in (so it does not exist for a leafref path in a typedef, for example).

For JSON format, it is simply the resolved parent node in the path. Is not used if there is no parent since then the node must be prefixed. The function could maybe check this, that it is not accessing NULL `ctx_node` and write the correct format-specific error.

And it is not used for XML, but that should probably be mentioned, too.

michalvasko (changes requested):

src/plugins_types.c
rkrejci: yes, I agree, it will be necessary for external plugins, so I expect it will be unified then.

src/plugins_types.h
```diff
@@ -77,19 +80,21 @@ struct lysp_module;
...
+ *
+ * Helper function for various plugin functions to generate error information structure.
  *
- * @param[in] level Error level of the error.
- * @param[in] code Code of the error.
- * @param[in] vecode Validity error code in case of LY_EVALID error code.
- * @param[in] msg Error message.
- * @param[in] path Path to the node causing the error.
- * @param[in] apptag Error-app-tag value.
+ * @param [in] ecode Code of the error to fill. If there is not enough memory for allocating error record then this code is
```
radek-isa, line 88: This parameter is now [in,out]

radek-isa (changes requested):

radek-isa (commented):
I would prefer to create new directory _**plugins**_ and put all plugins implementation into the directory. Also would be nice to create a function **_ struct lysc_type_plugin lysc_type_plugin_get(enum LY_DATA_TYPE data_type)_** which return struct lysc_type with required functions.

src/log.c
rkrejci: It was actually just taken from ly_err_msg_create(), but yeah, your suggestion makes sense and improves/simplifies the usage.

src/plugins_types.h
rkrejci: which API function? I'm confused, this is the header making ly_err_new() public. The function is implemented in log.c just because it extends the functionality of some other functions present there, but the only expected use is for type plugins, so the declaration is in this header.

michalvasko: Right, sorry, I somehow missed the fact that this is just a prototype.

rkrejci:
> Also would be nice to create a function **_ struct lysc_type_plugin lysc_type_plugin_get(enum LY_DATA_TYPE data_type)_** which return struct lysc_type with required functions.

What would be a use case for such a function?

src/plugins_types.h
rkrejci: fixed

src/log.c
```diff
@@ -119,23 +119,36 @@ ly_errapptag(const struct ly_ctx *ctx)
...
+    eitem->no = *ecode;
     eitem->vecode = vecode;
-    eitem->msg = msg;
     eitem->path = path;
     eitem->apptag = apptag;
 
+    if (err_msg) {
+        va_start(print_args, err_msg);
+
+        vasprintf(&msg, err_msg, print_args);
```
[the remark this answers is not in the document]
rkrejci: ok, fixed, but the only handling of it can be just setting err_msg to NULL.

src/plugins_types.h
rkrejci: the function was changed according to @michalvasko comment

src/plugins_types.h
rkrejci: I've tried to find similar functionality as provide ly_type_resolve_prefix(), but despite there are several similar use cases, there are always some important differences. So I've just renamed the function to `ly_type_identity_module()` to show its connection to identity types.

src/log.c
```diff
@@ -118,25 +118,52 @@ ly_errapptag(const struct ly_ctx *ctx)
...
+    if (!err || ecode == LY_SUCCESS) {
+        /* nothing to do */
+        return ecode;
+    }
 
-    eitem = malloc(sizeof *eitem);
-    LY_CHECK_ERR_RET(!eitem, LOGMEM(NULL), NULL);
-    eitem->prev = eitem;
-    eitem->next = NULL;
+    va_list(print_args);
```
michalvasko, line 132: It took me a few seconds to understand, why is it even possible to write a variable declaration this way? Please write it normally and put at the function start.

michalvasko (changes requested):

src/plugins_types.h
```diff
@@ -77,19 +80,23 @@ struct lysp_module;
...
-/**
- * @brief Helper function for various plugin functions to generate error information structure.
+/*
+ * @brief Create and fill error structure.
+ *
+ * Helper function for various plugin functions to generate error information structure.
  *
- * @param[in] level Error level of the error.
- * @param[in] code Code of the error.
+ * @param[out] err Pointer to store a new error structure filled according to the input parameters.
```
radek-isa, line 88: This parameter is in,out. It can be seen in file src/log.c on line 136 and 138-139

radek-isa (changes requested):

radek-isa:
> What would be a use case for such a function?

I was meaning instead of pointing directly to the plugin array _ly_builtin_type_plugins_ use function which returns the required structure. I thought it would be a little bit clearer to have one function returning plugin structure. But I have forgotten there is always a pointer to plugin structure.

src/plugins_types.h
rkrejci: right, fixed and description extended

src/log.c
```diff
@@ -118,25 +118,52 @@ ly_errapptag(const struct ly_ctx *ctx)
...
+    e->path = path;
+    e->apptag = apptag;
+
+    if (err_msg) {
+        va_start(print_args, err_msg);
+
+        if (vasprintf(&msg, err_msg, print_args) == -1) {
+            /* we don't have anything more to do, just set err_msg to NULL to avoid undefined content,
+             * still keep the information about the original error instead of LY_EMEM or other printf's error */
+            err_msg = NULL;
```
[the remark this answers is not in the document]
rkrejci: fixed, good catch

src/log.c
rkrejci: it wasn't intended, it was actually just copy-pasted from the previous form of the function, fixed (not moved to the function start, but just into the block which needs it)

rkrejci:
> > What would be a use case for such a function?
> 
> I was meaning instead of pointing directly to the plugin array _ly_builtin_type_plugins_ use function which returns the required structure. I thought it would be a little bit clearer to have one function returning plugin structure. But I have forgotten there is always a pointer to plugin structure.

There will be some changes regarding the list when the mechanism to load external plugins will be added, so I will let this change to a separate PR coming with this feature.

michalvasko (approved):
That `va_list()` declaration should ideally be fixed in the existing code, but that can be separate, not in this PR.

rkrejci:
> That `va_list()` declaration should ideally be fixed in the existing code, but that can be separate, not in this PR.

I'm confused - what's wrong? I thought I've fixed it.

michalvasko:
> I'm confused - what's wrong? I thought I've fixed it.

Sorry, my bad, ignore that.

radek-isa (approved):

PR #1466, opened 2021-03-19
libyang2: reimplementation bits plugin

radek-isa:


tests/utests/utests.h
```diff
@@ -844,16 +864,17 @@ struct utest_context {
...
  */
-#define CHECK_LYD_VALUE_BITS(NODE, CANNONICAL_VAL, VALUE) \
+#define CHECK_LYD_VALUE_BITS(NODE, CANNONICAL_VAL, ...) \
     assert_non_null((NODE).canonical); \
     assert_string_equal((NODE).canonical, CANNONICAL_VAL); \
     assert_non_null((NODE).realtype); \
     assert_int_equal(LY_TYPE_BITS, (NODE).realtype->basetype); \
     { \
-        LY_ARRAY_COUNT_TYPE arr_size = sizeof(VALUE) / sizeof(VALUE[0]); \
+        const char *arr[]={ __VA_ARGS__ };\
```
michalvasko, line 872: Strictly adhering to the coding style, missing spaces around `=` and before the last `\`.

src/plugins_types_bits.c
```diff
@@ -40,19 +40,19 @@ ly_type_store_bits(const struct ly_ctx *ctx, const struct lysc_type *type, const
...
-    size_t buf_size = 0;
-    char *buf = NULL;
-    size_t index;
-    LY_ARRAY_COUNT_TYPE u;
     struct lysc_type_bits *type_bits = (struct lysc_type_bits *)type;
-    ly_bool iscanonical = 1;
-    size_t ws_count;
-    size_t lws_count; /* leading whitespace count */
-    const char *can = NULL;
+    unsigned int index;
```
michalvasko, line 47: Please do not use `unsigned int`, `uint32_t` is the better alternative. Also, if there is no special reason, put declarations of same-type variables together.

src/plugins_types_bits.c
```diff
@@ -40,19 +40,19 @@ ly_type_store_bits(const struct ly_ctx *ctx, const struct lysc_type *type, const
...
+    unsigned int index_end = 0;
+    unsigned int item_len;
+    const char *item = NULL;
+    struct lysc_type_bitenum_item **bits_items = NULL;
+    // unsigned int buf_index = 0;
```
michalvasko, line 52: Should be removed, I suppose.

michalvasko (changes requested):

src/plugins_types_bits.c
```diff
@@ -62,113 +62,120 @@ ly_type_store_bits(const struct ly_ctx *ctx, const struct lysc_type *type, const
...
-                ret = ly_set_add(items, &type_bits->bits[u], 0, &inserted);
-                LY_CHECK_GOTO(ret, cleanup);
-                if (inserted != items->count - 1) {
-                    ret = ly_err_new(err, LY_EVALID, LYVE_DATA, NULL, NULL,
-                            "Bit \"%s\" used multiple times.", type_bits->bits[u].name);
-                    goto cleanup;
-                }
-                goto next;
+        item_len = index_end - index;
+        if (item_len != 0) {
```
rkrejci, line 82: your code doesn't cover `item_len == 0`

src/plugins_types_bits.c
```diff
@@ -62,113 +62,120 @@ ly_type_store_bits(const struct ly_ctx *ctx, const struct lysc_type *type, const
...
+            uint32_t item_pos;
+            uint32_t item_pos_expected;
+            LY_ARRAY_COUNT_TYPE array_it = 0;
+            LY_ARRAY_COUNT_TYPE array_size = LY_ARRAY_COUNT(type_bits->bits);
+
+            /* looking for correct name */
+            while ((array_it < array_size) &&
```
rkrejci, line 89: I believe that using `LY_ARRAY_FOR` is a more common and therefore also a more clear way to go through the sized array. The good point is to avoid the `next` label, but I would stay more with `LY_ARRAY_FOR` instead of rewriting this part to use `while` loops.

src/plugins_types_bits.c
rkrejci: ok, ignore the previous note, but maybe it would be clearer to detect earlier that the while loop was terminated because of `index < value_len` condition  (so `index_len` and `index` are the same) and the whole loop can be terminated (no item to store). Then this condition here will not be needed.

rkrejci (changes requested):

src/parser_yin.c
```diff
@@ -438,6 +438,13 @@ yin_validate_value(struct lys_yin_parser_ctx *ctx, enum yang_arg val_type)
         LY_CHECK_ERR_RET(already_read > ctx->xmlctx->value_len, LOGINT(ctx->xmlctx->ctx), LY_EINT);
     }
 
+    if ((already_read == 0) &&
+            ((val_type == Y_PREF_IDENTIF_ARG) || (val_type == Y_IDENTIF_ARG))) {
+        /* Empty identifier or prefix*/
+        LOGVAL_PARSER(ctx, LYVE_SYNTAX_YIN, "empty identifier");
```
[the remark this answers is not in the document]
radek-isa: Yes There is test for this case.
https://github.com/CESNET/libyang/blob/40eda514656b9d8e2f1551c81a3f5288130b607b/tests/utests/types/bits.c#L458-L465

src/plugins_types_bits.c
radek-isa: I don't think so. The string can contain only white characters then we need to skip this character.  Probably I should change variable name _index_ to _index_start_.

src/plugins_types_bits.c
radek-isa: For me is more readable when code contains gotos as less as possible
 I'm convinced that command goto should be used only for error handling.

michalvasko: I would not be so strict here and let everyone use whatever they prefer. As for `goto`, I agree it should preferably be avoided but what should strictly be followed is `goto` referencing only a label **below** itself. This is even what some tools checks and consider an error.

src/plugins_types_bits.c
```diff
@@ -40,19 +40,19 @@ ly_type_store_bits(const struct ly_ctx *ctx, const struct lysc_type *type, const
...
-    LY_ARRAY_COUNT_TYPE u;
     struct lysc_type_bits *type_bits = (struct lysc_type_bits *)type;
-    ly_bool iscanonical = 1;
-    size_t ws_count;
-    size_t lws_count; /* leading whitespace count */
-    const char *can = NULL;
+    struct lysc_type_bitenum_item **bits_items = NULL;
+    struct ly_set *items = NULL;
+
+    int index;
```
michalvasko, line 50: Well, even `int` usage should be avoided :) Except for cases when it should not, such as storing `int` return value of a standard function. Do you have something against using `uint32_t` and other types of this kind? It was an intentional effort to get rid of these types with variable length so please stick to it.

michalvasko (changes requested):

src/parser_yin.c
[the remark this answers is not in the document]
rkrejci: so why isn't this "bugfix" in that commit changing it? You are just fixing the issue you recently introduced in the same PR - rebase and join the commits.

src/plugins_types_bits.c
radek-isa: Ok. I'm going to change it to `uint32_t`. I have nothing against it.

src/plugins_types_bits.c
rkrejci: As I wrote, I'm appreciating removing the goto here, and I don't want to be strict regarding the while/for, but since most of the libyang code use `LY_ARRAY_FOR` for this purpose, it seems more natural to stick with it also here instead of using while in this part. Or am I'm missing some important point?

src/parser_yin.c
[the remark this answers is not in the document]
radek-isa: This test was added with a bunch of different tests for type bits. This test covers another bug that I founded. Then almost all fixes would be in one commit.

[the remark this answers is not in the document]
rkrejci: no, what I'm saying is that you can move that particular test from that bunch of tests and add it there together with this fix.

src/plugins_types_bits.c
rkrejci: yes, but I'm talking about the clarity of the code. What you describe is actually detection that there is no (more) bit name in the string (so not just when the string contains only spaces, but you will end up here also with the trailing spaces). Try just to invert the condition: if `item_len == 0` then terminate the loop with the appropriate comment. The rest of the bit name processing inside this `if` block can be then just put after that loop termination condition even without indentation.

src/plugins_types_bits.c
radek-isa: OK. Probably I can write it with FOR without goto.

src/plugins_types_bits.c
radek-isa: Then I would have to use command goto on non-error handling. OK, I'm going to do it but I really don't like it. As I write before. I'm convinced that command goto should be used only for error handling.

src/plugins_types_bits.c
```diff
@@ -22,6 +22,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdbool.h>
```
michalvasko, line 25: We are not using this header, there is `ly_bool` type defined for this purpose (with values `0` and `1`). There are some compatibility problems with `stdbool`.

src/plugins_types_bits.c
```diff
@@ -62,113 +69,123 @@ ly_type_store_bits(const struct ly_ctx *ctx, const struct lysc_type *type, const
...
-                            "Bit \"%s\" used multiple times.", type_bits->bits[u].name);
-                    goto cleanup;
-                }
-                goto next;
-            }
+
+        /* check if name of bit is valid */
+        item = &value[index_start];
+        item_len = index_end - index_start;
+        if (item_len == 0){
```
michalvasko, line 89: Missing space.

src/plugins_types_bits.c
```diff
@@ -62,113 +69,123 @@ ly_type_store_bits(const struct ly_ctx *ctx, const struct lysc_type *type, const
...
-            options &= ~LY_TYPE_STORE_DYNAMIC;
-            LY_CHECK_GOTO(ret, cleanup);
-        } else {
-            ret = lydict_insert(ctx, value_len ? &value[lws_count] : "", value_len - ws_count - lws_count, &can);
-            LY_CHECK_GOTO(ret, cleanup);
+
+        /* looking for correct name */
+        item_present = false;
+        LY_ARRAY_FOR(type_bits->bits, it){
+            if(!ly_strncmp(type_bits->bits[it].name, item, item_len)){
```
michalvasko, line 97: Missing spaces.

michalvasko (changes requested):

michalvasko (approved):

tests/utests/schema/test_parser_yin.c
```diff
@@ -657,7 +657,7 @@ test_validate_value(void **state)
     YCTX->xmlctx->value = "#invalid";
     YCTX->xmlctx->value_len = 8;
     assert_int_equal(yin_validate_value(YCTX, Y_IDENTIF_ARG), LY_EVALID);
-    CHECK_LOG_CTX("Invalid identifier character '#' (0x0023).", "Line number 1.");
+    CHECK_LOG_CTX("Invalid identifier first character '#' (0x0023).", "Line number 1.");
```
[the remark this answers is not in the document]
rkrejci: I still don't understand it - if you are changing the expected message (its wording) in tests, it means that the message changed somewhere in the code. But I don't see such a change here. Where is it?

src/plugins_types_bits.c
```diff
@@ -62,113 +68,123 @@ ly_type_store_bits(const struct ly_ctx *ctx, const struct lysc_type *type, const
...
-                }
-                goto next;
-            }
+
+        /* check if name of bit is valid */
+        item = &value[index_start];
+        item_len = index_end - index_start;
+        if (item_len == 0) {
+            /* loop read all bits names*/
+            goto bits_reading_end;
```
rkrejci, line 90: doesn't the `break` works the same way here?

src/plugins_types_bits.c
rkrejci: I believe it can be done without the label and goto - see the comment in the updated code.

src/plugins_types_bits.c
```diff
@@ -62,113 +68,123 @@ ly_type_store_bits(const struct ly_ctx *ctx, const struct lysc_type *type, const
     ret = ly_set_new(&items);
     LY_CHECK_GOTO(ret != LY_SUCCESS, cleanup_value);
 
-    ws_count = lws_count = 0;
-    for (index = 0; index < value_len; ++index) {
-        if (isspace(value[index])) {
-            ++ws_count;
-            continue;
+    /* get all values */
+    while (1) {
```
rkrejci, line 72: the condition can be probably `index_end < value_len` to cover the most common case when the string ends right after a bit name. It would also avoid illusory infinite loops.

src/plugins_types_bits.c
rkrejci: Thanks

rkrejci (changes requested):
Please remove the unnecessary goto.

tests/utests/schema/test_parser_yin.c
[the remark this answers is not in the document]
rkrejci: Or is it caused by changed order so a different error is now in charge?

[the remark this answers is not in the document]
radek-isa: Yes. The message is changed because of changing the order of code.

rkrejci (approved):

PR #1471, opened 2021-03-24
BUGFIX and improvements in tree_data and tree_data_helpers

lePici:


src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
  */
 int ly_strncmp(const char *refstr, const char *str, size_t str_len);
 
+/**
+ * @brief Get xml namespace prefix or just some substring from @p str.
```
michalvasko, line 373: This is not specific to XML, you can call it `YANG identifier prefix`, I guess.

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+ *
+ * The function looking for possible xml namespace prefix at the beginning of the @p str_begin and return it.
+ * If there is some string at the @p str_begin and then follows the prefix, then the string is returned.
+ * If @p str_begin does not contain any prefix, then the entire @p str_begin is returned.
+ * The @p rest can be used as the @p str_begin in the next iteration.
+ *
+ * @param[in] str_begin Begin of the input string.
+ * @param[in] str_end End of the input string. If set to NULL then the @p str_begin must be NULL-terminated string.
+ * @param[out] substr_type Type of substring found. Set to 1 for the xml namespace prefix, otherwise 0.
+ * @param[out] rest String the remainder part located behind the found substring.
```
michalvasko, line 383: `behind`, are you sure? I would guess it is after.

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+ * @brief Get xml namespace prefix or just some substring from @p str.
+ *
+ * The function looking for possible xml namespace prefix at the beginning of the @p str_begin and return it.
+ * If there is some string at the @p str_begin and then follows the prefix, then the string is returned.
+ * If @p str_begin does not contain any prefix, then the entire @p str_begin is returned.
+ * The @p rest can be used as the @p str_begin in the next iteration.
+ *
+ * @param[in] str_begin Begin of the input string.
+ * @param[in] str_end End of the input string. If set to NULL then the @p str_begin must be NULL-terminated string.
+ * @param[out] substr_type Type of substring found. Set to 1 for the xml namespace prefix, otherwise 0.
```
michalvasko, line 382: Either call it `is_prefix`, for example, or change the type to an enum. Preferably the former.

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+/**
+ * @brief Get xml namespace prefix or just some substring from @p str.
+ *
+ * The function looking for possible xml namespace prefix at the beginning of the @p str_begin and return it.
+ * If there is some string at the @p str_begin and then follows the prefix, then the string is returned.
+ * If @p str_begin does not contain any prefix, then the entire @p str_begin is returned.
+ * The @p rest can be used as the @p str_begin in the next iteration.
+ *
+ * @param[in] str_begin Begin of the input string.
+ * @param[in] str_end End of the input string. If set to NULL then the @p str_begin must be NULL-terminated string.
```
michalvasko, line 381: This information would be more convenient as the string length instead, don't you agree?

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+ * @param[out] substr_type Type of substring found. Set to 1 for the xml namespace prefix, otherwise 0.
+ * @param[out] rest String the remainder part located behind the found substring.
+ * If the @p substr_type is set to 1 then the colon character is skipped.
+ * If no string remains, it is set to NULL.
+ * @return A number of bytes of substring starting at position @p str_begin.
```
michalvasko, line 386: I would reword this to `Length of found substring starting at @p str_begin.`, for example.

michalvasko (commented):
Some tweaks needed but the full functionality is there.

src/common.h
lePici: I think the mention of bytes is more accurate, because the `ly_getutf8` function can have up to 4 bytes for one character.

michalvasko: Fair enough, then make it `Number of bytes of found substring starting at @p str_begin.`. Generally, when documenting this way with doxygen, skip articles at the beginning, that is the custom.

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+ *
+ * The function looking for possible YANG identifier prefix at the beginning of the @p str_begin and return it.
+ * If there is some string at the @p str_begin and then follows the prefix, then the string is returned.
+ * If @p str_begin does not contain any prefix, then the entire @p str_begin is returned.
+ * The @p rest can be used as the @p str_begin in the next iteration.
+ *
+ * @param[in] str_begin Begin of the input string.
+ * @param[in] str_end Length of the @p str_begin. If set to NULL then the @p str_begin must be NULL-terminated string.
+ * @param[out] is_prefix Type of substring found. Set to 1 for prefix, otherwise 0.
+ * @param[out] rest String the remainder part located after the found substring.
```
michalvasko, line 383: I would rename to `next` to be consistent with all similar iterator functions, even with the function name. The comment could then be improved to something like `Rest of the string after the found substring.`.

src/tree_data_helpers.c
```diff
@@ -568,38 +570,26 @@ ly_store_prefix_data(const struct ly_ctx *ctx, const char *value, size_t value_l
             prefixes = *prefix_data_p;
         }
 
+        value_end = value + value_len;
```
michalvasko, line 573: I would switch this line and the next one, to signify that this line is directly connected to the loop and functionality in it, it is not separate.

src/tree_data_helpers.c
```diff
@@ -616,39 +606,27 @@ ly_store_prefix_data(const struct ly_ctx *ctx, const char *value, size_t value_l
             ns_list = *prefix_data_p;
         }
 
+        value_end = value + value_len;
```
michalvasko, line 609: Same as the previous comment.

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
  */
 int ly_strncmp(const char *refstr, const char *str, size_t str_len);
 
+/**
+ * @brief Get YANG identifier prefix or just some substring from @p str.
```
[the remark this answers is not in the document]
lePici: ok

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+ *
+ * The function looking for possible YANG identifier prefix at the beginning of the @p str_begin and return it.
```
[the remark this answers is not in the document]
lePici: ok

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+ * If there is some string at the @p str_begin and then follows the prefix, then the string is returned.
+ * If @p str_begin does not contain any prefix, then the entire @p str_begin is returned.
+ * The @p rest can be used as the @p str_begin in the next iteration.
+ *
+ * @param[in] str_begin Begin of the input string.
+ * @param[in] str_end Length of the @p str_begin. If set to NULL then the @p str_begin must be NULL-terminated string.
+ * @param[out] is_prefix Type of substring found. Set to 1 for prefix, otherwise 0.
+ * @param[out] rest String the remainder part located after the found substring.
+ * If the @p substr_type is set to 1 then the colon character is skipped.
```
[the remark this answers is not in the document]
lePici: ok

src/common.h
lePici: ok

src/tree_data_helpers.c
lePici: ok

src/tree_data_helpers.c
lePici: ok

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+ *
+ * The function looks for possible prefix at the beginning of the @p str_begin and returns it.
+ * If there is some string at the @p str_begin and then follows the prefix, then the string is returned.
+ * If @p str_begin does not contain any prefix, then the entire @p str_begin is returned.
+ * The @p next can be used as the @p str_begin in the next iteration.
+ *
+ * @param[in] str_begin Begin of the input string.
+ * @param[in] str_end Length of the @p str_begin. If set to NULL then the @p str_begin must be NULL-terminated string.
+ * @param[out] is_prefix Type of substring found. Set to 1 for prefix, otherwise 0.
+ * @param[out] next Rest of the string after the found substring.
```
michalvasko, line 383: Just to be consistent with your other parameter names, maybe call this `str_next`.

lePici: ok

src/tree_data.c
```diff
@@ -2796,25 +2879,33 @@ lyd_compare_single(const struct lyd_node *node1, const struct lyd_node *node2, u
         }
     }
 
-    if ((LYD_CTX(node1) != LYD_CTX(node2)) || (node1->schema != node2->schema)) {
+    if (node1->hash != node2->hash) {
         return LY_ENOT;
     }
+    /* equal hashes do not mean equal nodes, they can be just in collision (or both be 0) so the nodes must be checked explicitly */
```
michalvasko, line 2885: I know you did not change this, but I do not think the hashes can be `0` so I would remove that.

src/tree_data.c
```diff
@@ -2847,7 +2938,16 @@ lyd_compare_single(const struct lyd_node *node1, const struct lyd_node *node2, u
...
+
+            if (LYD_CTX(node1) == LYD_CTX(node2)) {
+                return term1->value.realtype->plugin->compare(&term1->value, &term2->value);
+            } else {
+                if (strcmp(LYD_CANON_VALUE(node1), LYD_CANON_VALUE(node2))) {
+                    return LY_ENOT;
+                } else {
+                    return LY_SUCCESS;
+                }
+            }
```
michalvasko, line 2950: It took me a few seconds to notice that `break` is not missing because a `return` must be called but you can save these seconds for any next person by refactoring this code a bit. Something like this would be sufficient but it is up to you to improve it.
```C
            if (LYD_CTX(node1) == LYD_CTX(node2)) {
                return term1->value.realtype->plugin->compare(&term1->value, &term2->value);
            }

            /* different context */
            if (strcmp(LYD_CANON_VALUE(node1), LYD_CANON_VALUE(node2))) {
                return LY_ENOT;
            } else {
                return LY_SUCCESS;
            }
```
One `if-else` is fine, but not nested.

src/tree_data_helpers.c
```diff
@@ -567,39 +569,28 @@ ly_store_prefix_data(const struct ly_ctx *ctx, const char *value, size_t value_l
             assert(*format_p == LY_PREF_SCHEMA_RESOLVED);
             prefixes = *prefix_data_p;
         }
+        //TODO remove
```
michalvasko, line 572: I suppose all of these should be removed...

src/xml.c
```diff
@@ -1100,38 +1104,75 @@ lyxml_value_compare(const struct ly_ctx *ctx, const char *value1, void *val_pref
...
+            mod2 = ly_resolve_prefix(ctx2, value2_iter, value2_len, LY_PREF_XML, val_prefix_data2);
+        }
+        if (!mod1 || !mod2) {
+            /* not a prefix or maps to different namespaces */
+            ret = LY_ENOT;
+            break;
+        }
+
+        if (mod1->ctx == mod2->ctx) {
+            if ((mod1->name != mod2->name) || (mod1->revision != mod2->revision)) {
```
michalvasko, line 1150: I would like a comment `/* same contexts */` above this line.

src/xml.c
```diff
@@ -1100,38 +1104,75 @@ lyxml_value_compare(const struct ly_ctx *ctx, const char *value1, void *val_pref
...
+                ret = LY_ENOT;
+                break;
+            }
+        } else {
+            if (strcmp(mod1->name, mod2->name)) {
```
michalvasko, line 1155: And `/* different contexts */` above this.

src/tree_data.c
```diff
@@ -2779,6 +2779,82 @@ lyd_target(const struct ly_path *path, const struct lyd_node *tree)
...
+    }
+
+    if (parent1 || parent2) {
+        return 1;
+    }
+
+    return 0;
+}
+
+
```
michalvasko, line 2857: Extra line.

src/tree_data.c
```diff
@@ -2788,6 +2864,13 @@ lyd_compare_single(const struct lyd_node *node1, const struct lyd_node *node2, u
     struct lyd_node_opaq *opaq1, *opaq2;
     size_t len1, len2;
 
+/* Internal flag for lyd_compare_single function.
+ * This optimization flag will be set in the options variable
+ * if the lyd_compare_parental_schemas function is called.
+ * As a result, the lyd_compare_parental_schemas is performed only once.
+ */
+#define LYD_PARENTAL_SCHEMAS_COMPARED 0x80000000
```
michalvasko, line 2872: I do not like mixing API and internal flags in one variable but I suppose I could let this one slide.

src/tree_data.c
```diff
@@ -2779,6 +2779,82 @@ lyd_target(const struct ly_path *path, const struct lyd_node *tree)
...
+ */
+static ly_bool
+lyd_compare_schemas(const struct lysc_node *schema1, const struct lysc_node *schema2)
+{
+    if (schema1 == schema2) {
+        return 0;
+    } else if (!schema1 || !schema2) {
+        return 1;
+    }
+
```
michalvasko, line 2797: Here I would expect this code, with some comment even to explain it.
```C
if (schema1->module->ctx == schema2->module->ctx) {
    return 1;
}
```
Keep the same-context efficiency.

src/tree_data.c
```diff
@@ -2779,6 +2779,82 @@ lyd_target(const struct ly_path *path, const struct lyd_node *tree)
...
+ * @brief Compare all parental schemas of two data nodes.
+ *
+ * @param schema1 of first node.
+ * @param schema2 of second node.
+ * @return 0 if the all parental schemas are equivalent otherwise 1.
+ */
+static ly_bool
+lyd_compare_parental_schemas(const struct lyd_node *node1, const struct lyd_node *node2)
+{
+    const struct lyd_node_inner *parent1, *parent2;
```
michalvasko, line 2836: Just a hint, to avoid having to use `struct lyd_node_inner *` you can use the function `lyd_parent()`, it returns `struct lyd_node *`.

src/tree_data.c
```diff
@@ -2779,6 +2779,82 @@ lyd_target(const struct ly_path *path, const struct lyd_node *tree)
...
+
+    if (node1->schema == node2->schema) {
```
michalvasko, line 2838: You already did this comparison, is there any purpose to doing it again? You can put an assert, if you to make sure it works the way it should.

src/tree_data.c
```diff
@@ -2779,6 +2779,82 @@ lyd_target(const struct ly_path *path, const struct lyd_node *tree)
...
+}
+
+/**
+ * @brief Compare all parental schemas of two data nodes.
+ *
+ * @param schema1 of first node.
+ * @param schema2 of second node.
+ * @return 0 if the all parental schemas are equivalent otherwise 1.
+ */
+static ly_bool
```
michalvasko, line 2833: You are returning a boolean so I would expect `1` to be equal, `0` to be different but it inverted. Also, the name `compare` would imply return values just like `strcmp` but they are not, so this prototype is confusing. My idea would be to keep the boolean with inverted values and rename the function to `lyd_compare_schema_parents_equal`. No need for `schemas` (you cannot compare a single schema), `lyd_compare` prefix means it is specific to `lyd_compare` functions (not sure it is, once it is used somewhere else, it should be renamed), and `equal` means it is a predicate (boolean return).

src/tree_data.c
```diff
@@ -2779,6 +2779,82 @@ lyd_target(const struct ly_path *path, const struct lyd_node *tree)
...
 
+/**
+ * @brief Compare schemas of two Compiled YANG data nodes.
+ *
+ * @param schema1 of first node.
+ * @param schema2 of second node.
+ * @return 0 if the schemas are equivalent otherwise 1.
+ */
+static ly_bool
+lyd_compare_schemas(const struct lysc_node *schema1, const struct lysc_node *schema2)
```
michalvasko, line 2790: Rename to `lyd_compare_schema_equal`, same rationale as for the parent variant.

src/tree_data.c
lePici: So no changes are required?

src/tree_data.c
lePici: why is this necessary? I thought that when the pointers `schema1 == schema2` are equal, they are in the same contexts. Isn't that so?

src/tree_data.c
lePici: assert won't help here. The purpose of this condition is optimization. Thanks to it, it does not have to iterate through the `for `cycle below. I assume that if the schemes are equal, all parent nodes must be equal.

src/tree_data.c
lePici: ok

src/tree_data.c
lePici: ok

src/tree_data.c
lePici: ok

src/tree_data.c
lePici: ok

src/xml.c
lePici: ok

src/xml.c
lePici: ok

src/tree_data_helpers.c
lePici: yes

src/tree_data.c
michalvasko: But what about the situation when they are in the same context but `schema1 != schema2`? Right now the whole comparison for different context is performed, which is completely useless. What you should generally do is strictly distinguish in every comparison whether the contexts are the same or not, it save you these kinds of problems.

src/tree_data.c
lePici: ok

src/tree_data.c
lePici: ok

src/tree_data.c
michalvasko: Right, I have forgotten a comment I wanted to put here, again distinguishing between the same context and a different one. You do not do that at all and only comparing the schemas for equality is not enough, as I have mentioned.

src/tree_data.c
michalvasko: I guess, I will let @rkrejci decide.

src/tree_data.c
lePici: ok, I get it

src/tree_data.c
```diff
@@ -2779,6 +2779,83 @@ lyd_target(const struct ly_path *path, const struct lyd_node *tree)
...
+        return 0;
+    }
+
+    assert(schema1->module->ctx != schema2->module->ctx);
+
+    if (schema1->nodetype != schema2->nodetype) {
+        return 0;
+    }
+
+    if (schema1->module->latest_revision != schema2->module->latest_revision) {
```
michalvasko, line 2804: Ah, another thing I forgot about before, but no harm done in this case. This is redundant and actually wrong, ignore those `latest_revision` flags, they have no relevance for this purpose.

src/common.h
```diff
@@ -369,6 +369,24 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+ *
+ * @param[in] str_begin Begin of the input string.
+ * @param[in] str_end Length of the @p str_begin. If set to NULL then the @p str_begin must be NULL-terminated string.
+ * @param[out] is_prefix Type of substring found. Set to 1 for prefix, otherwise 0.
+ * @param[out] str_next Rest of the string after the found substring.
+ * If the @p is_prefix is set to 1 then the colon character is skipped.
+ * If no string remains, it is set to NULL.
+ * @return Number of bytes (length) of the found substring starting at @p str_begin.
+ */
+size_t ly_string_next(const char *str_begin, const char *str_end, int *is_prefix, const char **str_next);
```
michalvasko, line 388: I have not noticed before, why is `is_prefix` `int`? It should definitely be a `ly_bool`.

src/xml.c
```diff
@@ -1100,38 +1104,77 @@ lyxml_value_compare(const struct ly_ctx *ctx, const char *value1, void *val_pref
...
+            value1_iter = value1_next, value2_iter = value2_next) {
+        value1_len = ly_string_next(value1_iter, NULL, &is_prefix1, &value1_next);
+        value2_len = ly_string_next(value2_iter, NULL, &is_prefix2, &value2_next);
+
+        if (is_prefix1 != is_prefix2) {
+            ret = LY_ENOT;
+            break;
+        }
+
+        if (is_prefix1 != 1) {
```
michalvasko, line 1123: Connected to a previous comment, it is a boolean value (even based on the name), so treat it that way.

michalvasko (commented):
Just a side note, be careful when using `size_t` type because it is not possible (or insanely complex) to print such a value in a portable way. It does not matter in this case but the general rule is sticking to basic `stdint` types.

src/tree_data.c
lePici: It is interesting that it has to be deleted. If I delete it, it does not pass tests in sysrepo. Are you sure about that?

michalvasko: Yes, pretty sure. Please, never adjust the code for the tests without understanding the specific test-case. So, when you understand what exactly is going on (what I can help with), let me know and we can discuss what to do.

src/xml.h
```diff
@@ -164,16 +164,17 @@ void lyxml_ctx_free(struct lyxml_ctx *xmlctx);
 /**
  * @brief Compare values and their prefix mappings.
  *
- * @param[in] ctx libyang context.
+ * @param[in] ctx1 libyang context of First value.
```
rkrejci, line 167: refer to the parameter: `@p value1`

rkrejci (changes requested):

src/xml.h
michalvasko: It was my initial idea as well but I let it be since `value1` does not have a context, strictly speaking. So the description will have to be a bit more verbose, something like `ctx1 libyang context for resolving prefixes in @p value1`.

src/common.h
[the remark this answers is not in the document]
michalvasko: Not sure if @lePici is able to come up with something accurate enough, I think we should help. The idea is that this function can take any string in a YANG module (hence called YANG string) such as an extension name but even an instance-identifier default value and iteratively return for the caller all possible prefixes (substrings that could be interpreted as a prefix if the whole string conforms to some subset of XPath). The use-case is quite simple, I think, but accurately describing this function is not. Can you @rkrejci suggest some specific improvements based on what I wrote?

Now that I think about, maybe I was forcing this function to be more generic than is needed because for cases such as the extension instance name, this function is not (and will likely never be) used because of strict syntax restrictions. Meaning it will be used only for YANG data values (and YANG schema default values). So maybe `YANG string` can be switched for `YANG value` and the function renamed to `ly_value_next`.

lePici: ok

src/xml.c
lePici: ok

src/common.h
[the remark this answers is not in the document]
lePici: I will use the names `ly_value_prefix_next` and `YANG value` for now.

src/xml.h
lePici: ok

src/common.c
```diff
@@ -66,6 +66,73 @@ ly_strncmp(const char *refstr, const char *str, size_t str_len)
     }
 }
 
+uint64_t
```
michalvasko, line 69: I mean, it probably does not matter but this is redundant, `uint32_t` is more than enough.

src/tree_data.c
michalvasko: Does not seem fixed.

src/common.h
```diff
@@ -369,6 +369,36 @@ char *ly_strnchr(const char *s, int c, size_t len);
...
+ * Examples of inputs and outputs are shown in the table below.
+ * Output string @p str_next is used in the next iteration as input parameter @p str_begin.
+   @verbatim
+   |            INPUT              |                         OUTPUT                        |
+   |                               |   iteration 1    |   iteration 2    |   iteration 3   |
+   |------------------------------ |------------------|------------------|-----------------|
+   | /namespace_prefix:some_string | /                | namespace_prefix | some_string     |
+   | namespace_prefix:some_string  | namespace_prefix | some_string      | NULL            |
+   | /some_string                  | /some_string     | NULL             | NULL            |
+   @endverbatim
```
rkrejci, line 389: I miss the example regarding the case 2 described above, probably this (not sure if it is correct, especially regarding the semicolon character):
```
| string1;string2     | string1;  | string2 | NULL
```

rkrejci (changes requested):

src/common.h
michalvasko: Semicolon is not treated specially in any way, there is nothing in the description hinting this, right? It should simply be
```
| string1;string2     | string1;string2 | NULL
```
Right @lePici ?

lePici: Yes only the colon is taken into account. I will add information about the colon to the documentation. But if there is a use-case with a semicolon somewhere, I can extend the function specification.

michalvasko: No, to my knowledge it would be redundant to do anything with a semicolon so I do not know why @rkrejci asked about it.

rkrejci: ok, what about `string1;prefix:something`? That's the case that matters, right? My point is that point 2 in the description talks about some substring before the prefix, but it is not specified how the substring and prefix are delimited.

rkrejci: and it's not only semicolon, it is recognized using `is_xmlqnamechar()`

michalvasko: Oh, okay, this information can be added, I suppose. Can you suggest something concise? The description was a bit general because I do not think you need to know the exact delimiting characters to be able to use it, you just know that you have a YANG value and want to get the prefixes from it, so you will use this function. Also, for this to work perfectly the input format should probably matter as in JSON the prefixes can only be YANG module names meaning a subset of the characters allowed for XML. But even if something else is returned for JSON, it will never match a module name, so no harm done.

lePici: ok, I'll add this comment to the description.

michalvasko (approved):

rkrejci (approved):

PR #1479, opened 2021-03-30
context FEATURE use module hash as the context ID

michalvasko:
So that it is always different for a different
set of modules, not just relatively for single
context changes.
Refs sysrepo/sysrepo#2378


rkrejci:
After thinking about this more, I'm afraid that hash is not a good approach here because of possible collisions. The use case for the context ID is to allow simply get know that the context changed and an application needs to refresh its understanding of the context. But it is not doable with hash, because the same hash does not mean the same context.

Probably a global counter (maybe stored externally to survive restarts) will be better because it has the opposite logic in contrast to the hash. It is actually fine for the application to get 2 different values, think that the context changed and after processing the context get to know that they are actually the same.

PR #1481, opened 2021-03-31
context FEATURE allow custom content-id in yang-library data

michalvasko:


src/context.c
```diff
@@ -322,10 +322,11 @@ ly_ctx_unset_options(struct ly_ctx *ctx, uint16_t option)
 }
 
 API uint16_t
-ly_ctx_get_module_set_id(const struct ly_ctx *ctx)
+ly_ctx_get_change_count(const struct ly_ctx *ctx)
```
rkrejci, line 325: please add the information about the change into the transition manual

src/context.h
```diff
@@ -510,24 +508,27 @@ uint32_t ly_ctx_internal_modules_count(const struct ly_ctx *ctx);
...
+ * with the key value "complete".
+ *
+ * If the data identifier can be limited to the existence and changes of this context, the following
+ * last 2 parameters can be used:
+ *
+ * "%u" as @p content_id_format and ::ly_ctx_get_change_count() as its parameter.
+ *
  * @param[in] ctx Context with the modules.
  * @param[out] root Generated yang-library data.
+ * @param[in] content_id_format Format string for the yang-library data identifier, which is the "content_id" node
```
rkrejci, line 526: please mention explicitely that it is printf's format string

tests/utests/basic/test_yanglib.c
```diff
@@ -114,7 +114,7 @@ test_yanglib(void **state)
     UTEST_ADD_MODULE(schema_a, LYS_IN_YANG, feats, NULL);
     UTEST_ADD_MODULE(schema_b, LYS_IN_YANG, NULL, NULL);
 
-    assert_int_equal(LY_SUCCESS, ly_ctx_get_yanglib_data(UTEST_LYCTX, &tree));
+    assert_int_equal(LY_SUCCESS, ly_ctx_get_yanglib_data(UTEST_LYCTX, &tree, "%u", ly_ctx_get_change_count(UTEST_LYCTX)));
```
rkrejci, line 117: please add also some weird format strings, like `"<<%u>>"`, `"%u" , -10`, to make sure that these situations are handled correctly (from the code I believe they are).

rkrejci: or just `""`.

rkrejci (changes requested):

src/context.h
michalvasko: Fixed.

tests/utests/basic/test_yanglib.c
michalvasko: The implementation should be completely dependent on `vasprintf` functionality so I considered more tests redundant. In any case, the ones you mentioned were added.

PR #1485, opened 2021-04-02
printers MAINTENANCE restore json printing to previous style

choppsv1:
Makes 2 changes, to return formatting to what was in libyang1

- Change:
```
    {
      list-node-content
    },    {
      list-node-content
    }
```
 Back to the definitely more common:
```
    {
      list-node-content
    },
    {
      list-node-content
    }
```
- Change leaf lists back to be one item per line when formatting. While
one line might look nice for leaf lists with only a couple items, it
can't be considered "formated" when that list grows to many items
and everything is on a single line wrapping around the screen.

### DIFF
--- a/src/printer_json.c
+++ b/src/printer_json.c
@@ -97,8 +97,7 @@ matching_node(const struct lyd_node *node1, const struct lyd_node *node2)
 static LY_ERR
 json_print_array_open(struct jsonpr_ctx *ctx, const struct lyd_node *node)
 {
-    /* leaf-list's content is always printed on a single line */
-    ly_print_(ctx->out, "[%s", (!node->schema || node->schema->nodetype != LYS_LEAFLIST) && DO_FORMAT ? "\n" : "");
+    ly_print_(ctx->out, "[%s", DO_FORMAT ? "\n" : "");
     LY_CHECK_RET(ly_set_add(&ctx->open, (void *)node, 0, NULL));
     LEVEL_INC;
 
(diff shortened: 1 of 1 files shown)


src/printer_json.c
```diff
@@ -135,12 +135,7 @@ json_print_array_close(struct jsonpr_ctx *ctx)
 
     LEVEL_DEC;
     ly_set_rm_index(&ctx->open, ctx->open.count - 1, NULL);
-    if (schema && (schema->nodetype == LYS_LEAFLIST)) {
```
rkrejci, line 138: The `schema` variable is unused now.

src/printer_json.c
```diff
@@ -98,7 +98,7 @@ static LY_ERR
 json_print_array_open(struct jsonpr_ctx *ctx, const struct lyd_node *node)
 {
     /* leaf-list's content is always printed on a single line */
```
rkrejci, line 100: This comment is no more right.

rkrejci (changes requested):
I see that when doing formatting (indentation, etc.) of the output, it should be done completely including the leaf-list's values. But, besides shrinking everything on a single line with `LYD_PRINT_SHRINK`, does it make sense to keep the current way of formating via some printer option? Let's say `LYD_PRINT_JSON_LEAFLIST_ONELINE`?

choppsv1:
> I see that when doing formatting (indentation, etc.) of the output, it should be done completely including the leaf-list's values. But, besides shrinking everything on a single line with `LYD_PRINT_SHRINK`, does it make sense to keep the current way of formating via some printer option? Let's say `LYD_PRINT_JSON_LEAFLIST_ONELINE`?

I don't think it's worth it. These types of choices typically are made based on tty columns etc, seems not worth the complexity in the API. Perhaps you could open an issue tracking that enhancement idea since adding flags would be backward compatible at some future date.

src/printer_json.c
choppsv1: fixed

src/printer_json.c
choppsv1: fixed

choppsv1:
@rkrejci the CI failure seems to be related to the network, not the change. Can you please rerun the workflow?

rkrejci:
> Perhaps you could open an issue tracking that enhancement idea since adding flags would be backward compatible at some future date.

That's fair, I'm going to do that.

rkrejci (approved):

Issue #1483, opened 2021-04-02
ly2: keyless lists and paths broken


choppsv1:
There appears to be a problem with keyless lists. In particular `ly_path_eval_partial` when finding the element in the path using the position predicate `LYD_LIST_FOR_INST` is used to locate the node, this macro calls `lyd_find_sibling_val` which then calls `lyd_find_sibling_schema` (no key_or_value given) when the list is inside a container there is a children_ht so a hash lookup is done with

```
        ht_cb = lyht_set_cb(parent->children_ht, lyd_hash_table_schema_val_equal);

        /* find by hash */
        if (!lyht_find(parent->children_ht, &schema, hash, (void **)&match_p)) {
```

This lyht_find is not returning the first child of the parent with the matching schema which leads to all sorts of trouble later. When I disable the hash lookup so that the else condition is run (the slower linear search), everything works fine.

choppsv1:
Here's repro code, first the bad result:

```
{
  "main:c": {
    "l": [
      {
        "a1": 0,
        "a2": 1
      },
      {
        "a1": 2,
        "a2": 3
      },
      {
        "a1": 4,
        "a2": 5
      },
      {
        "a1": 6,
        "a2": 7
      },
      {
        "a1": 8
      },
      {
        "a2": 9
      },
      {
        "a1": 10
      },
      {
        "a2": 11
      }
    ]
  }
}
```

main.yang:

```
module main {
  yang-version 1.1;
  namespace "http://example.com/main";
  prefix m;

  container c {
    config false;
    list l {
      leaf a1 { type uint32; }
      leaf a2 { type uint32; }
    }
  }
}
```

main.c:

```
#include <assert.h>
#include <err.h>
#include <stddef.h>
#include <stdio.h>

#include <libyang/libyang.h>

int main(int argc, char **argv)
{
    struct ly_ctx *ctx = NULL;
    struct lyd_node *cnode, *node = NULL;
    const struct lys_module *main = NULL;
    char xpath[32], value[32];
    int i, j, v = 0;

    ly_log_options(LY_LOLOG | LY_LOSTORE);
    assert(!ly_ctx_new(".", 0, &ctx));
    assert(main = ly_ctx_load_module(ctx, "main", NULL, NULL));
    assert(!lyd_new_inner(NULL, main, "c", 0, &cnode));

    for (int i = 1; i < 7; i++) {
        for (int j = 1; j <= 2; j++) {
[...]

PR #1487, opened 2021-04-06
data tree BUGFIX temp workaround for keyless list insertions

choppsv1:
This is a temporary fix for Issue #1483 see that comment thread in
https://github.com/CESNET/libyang/issues/1483 for discussion on
eventually permanent solution.

Signed-off-by: Christian Hopps <chopps@labn.net>


michalvasko:
Okay, fine, but looking more at the code, this really is a "workaround" at best and should not remain in the code longer than necessary.

PR #1488, opened 2021-04-06
Schema compilation fixes

michalvasko:
Disabled schema node-id targets and some possibly false-positive warnings.

### DIFF
--- a/src/schema_compile.c
+++ b/src/schema_compile.c
@@ -1397,6 +1397,7 @@ lys_compile_unres_mod(struct lysc_ctx *ctx)
     struct lysc_node *node;
     struct lysc_augment *aug;
     struct lysc_deviation *dev;
+    const struct lys_module *mod;
     uint32_t i;
 
     /* remove all disabled nodes */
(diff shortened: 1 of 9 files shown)


rkrejci (commented):
I don't know - I understand the motivation (https://mailarchive.ietf.org/arch/msg/netconf/3mu3VULbrwseOBjgG0Md_EqEtEc/), but this approach based on guessing that the module has something disabled, doesn't seem very trustworthy. There are so many cases when the checking features of the module itself doesn't work ('not' in YANG 1.1, features imported from other module, really invalid path), so there will be many false positives as well as false negatives case. At least mark it as some TODO or TEMPORARY solution, because this cannot be the final solution.
I know that not an optimal, but at least much more accurate, approach would be to remember the list of paths which were disabled when compiling the schema. Ideally the lists should be hidden in context to allow later optimization/change of the solution

michalvasko:
Alright, a proper solution will be implemented that should always be able to distinguish between an invalid target, disabled target, and existing target.

PR #1491, opened 2021-04-07
Libyang2 context

Brwo:
Added two functions ly_ctx_new_ylmem and ly_ctx_new_ylpath from old libyang.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -42,6 +42,7 @@
 #include "set.h"
 #include "tree.h"
 #include "tree_data.h"
+#include "tree_data_internal.h"
 #include "tree_schema.h"
 #include "tree_schema_internal.h"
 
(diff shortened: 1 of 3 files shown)


src/context.c
```diff
@@ -290,6 +291,171 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+    LY_CHECK_RET(ret = lyd_find_xpath(yltree, "/ietf-yang-library:yang-library/modules-state/module", &set), ret);
+
+    /* process the data tree */
+    for (uint32_t i = 0; i < set->count; ++i) {
+        module = set->dnodes[i];
+
+        /* initiate */
+        const char *name = NULL, *revision = NULL;
+        struct ly_set features = {0};
+        ly_bool imported = 0;
```
michalvasko, line 311: I am not sure about this. On one hand, the way you wrote it it is concise and clearly separate from the rest of the code. But at the same time variable declarations should always be at the beginning of a code block. I would prefer you always stick to this rule.

src/context.c
```diff
@@ -290,6 +291,171 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+
+        LY_LIST_FOR(*lyd_node_child_p(module), node) {
```
michalvasko, line 313: The same functionality is achieved by `lyd_child(module)`.

src/context.c
```diff
@@ -290,6 +291,171 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+            if (!strcmp(node->schema->name, "name")) {
+                name = ((struct lyd_node_term *)node)->value.canonical;
```
michalvasko, line 315: You can use a macro `name = LYD_CANON_VALUE(node);`, many such improvements possible.

src/context.c
```diff
@@ -290,6 +291,171 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+            } else if (!strcmp(node->schema->name, "revision")) {
+                revision = ((struct lyd_node_term *)node)->value.canonical;
+            } else if (!strcmp(node->schema->name, "feature")) {
+                LY_CHECK_ERR_RET(ret = ly_set_add(&features, node, 0, NULL),
+                        ly_set_free(&features, free); ly_set_free(set, free), ret);
+            } else if (!strcmp(node->schema->name, "conformance-type") &&
+                    ((struct lyd_node_any *)node)->meta->value.enum_item->value) {
```
michalvasko, line 322: Um, what is the meaning of this code? I am clueless.

src/context.c
```diff
@@ -290,6 +291,171 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+                break;
+            }
+        }
+
+        if (imported) {
+            continue;
+        }
+
+        feature_arr = malloc(sizeof(char *) * (features.count + 1));
+        LY_CHECK_ERR_RET(feature_arr == NULL, ly_set_free(&features, free), LY_EMEM);
```
michalvasko, line 335: Use just `!` for `NULL` equivalence checking.

src/context.c
```diff
@@ -290,6 +291,171 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
     return rc;
 }
 
+static LY_ERR
+ly_ctx_new_yl_legacy(struct ly_ctx **ctx, struct lyd_node *yltree)
```
michalvasko, line 295: Did I miss it or the `ctx` param can be simply `struct ly_ctx *ctx`?

src/context.c
```diff
@@ -290,6 +291,171 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+        /* perhaps a legacy data tree? */
+        LY_CHECK_GOTO(ret = ly_ctx_new_yl_legacy(ctx, yltree), error);
+    } else {
+        /* process the data tree */
+        for (uint32_t i = 0; i < set->count; ++i) {
+            module = set->dnodes[i];
+
+            /* initiate */
+            const char *name = NULL, *revision = NULL;
+
```
michalvasko, line 389: Extra line.

src/context.c
```diff
@@ -290,6 +291,171 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+                imported = 1;
+                break;
+            }
+        }
+
+        if (imported) {
+            continue;
+        }
+
+        feature_arr = malloc(sizeof(char *) * (features.count + 1));
```
michalvasko, line 334: Avoid `sizeof(type)` whenever possible, always prefer using a variable directly (if you change its type later, the size will change automatically). So write `sizeof *feature_arr` instead.

src/context.c
```diff
@@ -290,6 +291,171 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+            feature_arr = NULL;
+            if (!mod) {
+                LOGERR(NULL, LY_EINVAL, "Unable to load module specified by yang library data.");
+                ret = LY_EINVAL;
+                goto error;
+            }
+        }
+    }
+
+    if (0) {
```
michalvasko, line 424: Nope, never write something like this again. I believe the whole `err` variable is useless because you can simply use `ret` - `!ret` - success, `ret` - error.

michalvasko (changes requested):

src/context.h
```diff
@@ -212,6 +213,53 @@ struct ly_ctx;
...
+ *
+ * This function loads the yang-library data from the given path. If you need to pass the data as
+ * string, use ::ly_ctx_new_ylpath(). Both functions extend functionality of ::ly_ctx_new() by loading
+ * modules specified in the ietf-yang-library form into the context being created.
+ * The preferred tree model revision is 2019-01-04. However, only the first module-set is processed and loaded
+ * into the context. If there are no matching nodes from this tree, the legacy tree (originally from model revision 2016-04-09)
+ * is processed. Note, that the modules are loaded the same way as in case of ::ly_ctx_load_module(), so the schema paths in the
+ * yang-library data are ignored and the modules are loaded from the context's search locations. On the other hand, YANG features
+ * of the modules are set as specified in the yang-library data.
+ * To get yang library data from a libyang context, use ::ly_ctx_info().
```
rkrejci, line 228: replace `::ly_ctx_info()` by `::ly_ctx_get_yanglib_data()`

src/context.h
```diff
@@ -212,6 +213,53 @@ struct ly_ctx;
...
+ *
+ * This function loads the yang-library data from the given string. If you need to pass the data as
+ * path to a file holding the data, use ::ly_ctx_new_ylpath(). Both functions extend functionality of
+ * ::ly_ctx_new() by loading modules specified in the ietf-yang-library form into the context being created.
+ * The preferred tree model revision is 2019-01-04. However, only the first module-set is processed and loaded
+ * into the context. If there are no matching nodes from this tree, the legacy tree (originally from model revision 2016-04-09)
+ * is processed. Note, that the modules are loaded the same way as in case of ::ly_ctx_load_module(), so the schema paths in the
+ * yang-library data are ignored and the modules are loaded from the context's search locations. On the other hand, YANG features
+ * of the modules are set as specified in the yang-library data.
+ * To get yang library data from a libyang context, use ::ly_ctx_info().
```
rkrejci, line 251: replace ::ly_ctx_info() by ::ly_ctx_get_yanglib_data()

src/context.h
```diff
@@ -212,6 +213,53 @@ struct ly_ctx;
...
+ *
+ * @param[in] search_dir Directory where libyang will search for the imported or included modules and submodules.
+ * If no such directory is available, NULL is accepted.
+ * @param[in] data Path to the file containing yang-library-data in the specified format
```
rkrejci, line 232: replace `data` by `path`

rkrejci (changes requested):
The commit message for the tests' commit should be:
```
tests FEATURE ...
```

src/context.c
```diff
@@ -290,6 +291,184 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+                break;
+            }
+        }
+
+        if (imported) {
+            continue;
+        }
+
+        feature_arr = malloc(sizeof(*feature_arr) * (features.count + 1));
+        LY_CHECK_ERR_RET(!feature_arr, ly_set_free(&features, free), LY_EMEM);
```
michalvasko, line 339: Same cleanup problem as before.

michalvasko (changes requested):

rkrejci (changes requested):
Please fix the compiler's warnings for `context.c` (see the Actions' results)

src/context.c
```diff
@@ -290,6 +291,189 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+                imported = 1;
+                break;
+            }
+        }
+
+        if (imported) {
+            continue;
+        }
+
+        feature_arr = malloc(sizeof *feature_arr * (features.count + 1));
```
michalvasko, line 337: Just a note, you can keep it this way, but the custom is to first write number of items and then the size of each item, you wrote it the other way around. If you stick to this rule, it will follow the way `calloc` has its arguments and it will avoid possible confusion with `*` since the `sizeof` is always at the end.

src/context.c
```diff
@@ -290,6 +291,189 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+            LOGERR(ctx, LY_EINVAL, "Unable to load module specified by yang library data.");
+            ly_set_free(set, free);
+            return LY_EINVAL;
+        }
+    }
+
+error:
+    ly_set_clean(&features, free);
+    ly_set_free(set, free);
+    if (feature_arr) {
```
michalvasko, line 361: Redundant `if`.

src/context.c
```diff
@@ -290,6 +291,189 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+    }
+
+error:
+
+    /* cleanup */
+    /* yang library data tree */
+    lyd_free_all(yltree);
+    ly_set_free(set, NULL);
+    ly_set_erase(&features, NULL);
+    if (feature_arr) {
```
michalvasko, line 448: Again, redundant.

src/context.c
```diff
@@ -290,6 +291,189 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+        free(feature_arr);
+    }
+    if (ctx_yl != ctx_new) {
+        ly_ctx_destroy(ctx_yl, NULL);
+    }
+    *ctx = ctx_new;
+    if (ret) {
+        ly_ctx_destroy(*ctx, NULL);
+        *ctx = NULL;
+    }
```
michalvasko, line 458: Just a minor improvement, I would write it like
```C
    if (ret) {
        ly_ctx_destroy(ctx_new, NULL);
        *ctx = NULL;
    } else {
        *ctx = ctx_new;
    }
```
which clearly shows the difference between success and error. But you can keep it this way.

michalvasko (changes requested):

src/context.c
```diff
@@ -290,6 +291,181 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
...
+                ret = LY_EINVAL;
+                goto error;
+            }
+        }
+    }
+
+error:
+
+    /* cleanup */
+    /* yang library data tree */
```
michalvasko, line 439: Seems like an extra line and both comments are redundant since the label clearly states what the following code is. Actually, have noticed just now, the label should say `cleanup` instead of `error`.

michalvasko (changes requested):

michalvasko (approved):

rkrejci (approved):

PR #1498, opened 2021-04-08
data tree FEATURE support for lists with duplicate instances

michalvasko:
Changes mainly in diff processing.
Refs #1483


rkrejci:
Please note that the changed `yang` module affects also #1491, so the later PR being merged will have to reflect changes of the other.

rkrejci (approved):

PR #1499, opened 2021-04-08
schema compile CHANGE remove possibly false-positive warnings

michalvasko:
Changed to verbose messages.

### DIFF
--- a/src/schema_compile_node.c
+++ b/src/schema_compile_node.c
@@ -2420,7 +2420,7 @@ lys_compile_node_(struct lysc_ctx *ctx, struct lysp_node *pnode, struct lysc_nod
     /* list ordering */
     if (node->nodetype & (LYS_LIST | LYS_LEAFLIST)) {
         if ((node->flags & (LYS_CONFIG_R | LYS_IS_OUTPUT | LYS_IS_NOTIF)) && (node->flags & LYS_ORDBY_MASK)) {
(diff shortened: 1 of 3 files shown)


src/schema_compile_node.c
```diff
@@ -2420,7 +2420,7 @@ lys_compile_node_(struct lysc_ctx *ctx, struct lysp_node *pnode, struct lysc_nod
     /* list ordering */
     if (node->nodetype & (LYS_LIST | LYS_LEAFLIST)) {
         if ((node->flags & (LYS_CONFIG_R | LYS_IS_OUTPUT | LYS_IS_NOTIF)) && (node->flags & LYS_ORDBY_MASK)) {
-            LOGWRN(ctx->ctx, "The ordered-by statement is ignored in lists representing %s (%s).",
+            LOGVRB("The ordered-by statement is ignored in lists representing %s (%s).",
```
rkrejci, line 2423: Why is this just a verbose? The implicit `LYS_ORDBY_USER` is added into the list after this check and explicitly adding ordered-by into the state lists isn't correct, is it?

michalvasko: The reason for this change is a standardized module using a `grouping` with a `list` with `ordered-by` in a notification, so that no warnings are printed in this case.

rkrejci: ok

rkrejci (approved):

PR #1500, opened 2021-04-12
context FEATURE added LY_CTX_SET_PRIV_PARSED

lePici:
With this option, libyang sets ::lysc_node.priv pointers as a reference to the corresponding ::lysp_node.
This was introduced due to the printer tree module (RFC 8340), which needs both a compiled and a parsed tree to print.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -473,12 +473,30 @@ ly_ctx_get_options(const struct ly_ctx *ctx)
 API LY_ERR
 ly_ctx_set_options(struct ly_ctx *ctx, uint16_t option)
 {
+    LY_ERR lyrc;
     LY_CHECK_ARG_RET(ctx, ctx, LY_EINVAL);
     LY_CHECK_ERR_RET(option & LY_CTX_NO_YANGLIBRARY, LOGARG(ctx, option), LY_EINVAL);
+    lyrc = LY_SUCCESS;
+
+    if (!(ctx->flags & LY_CTX_SET_PRIV_PARSED) && (option & LY_CTX_SET_PRIV_PARSED)) {
+        ctx->flags |= LY_CTX_SET_PRIV_PARSED;
+        if ((lyrc = lys_recompile(ctx, 0))) {
+            ly_ctx_unset_options(ctx, LY_CTX_SET_PRIV_PARSED);
+        }
+    }
 
     /* set the option(s) */
-    ctx->flags |= option;
+    if (!lyrc) {
+        ctx->flags |= option;
+    }
 
+    return lyrc;
+}
+
+static LY_ERR
+lysc_node_clear_priv_dfs_cb(struct lysc_node *node, void *UNUSED(data), ly_bool *UNUSED(dfs_continue))
+{
+    node->priv = NULL;
     return LY_SUCCESS;
 }
 
@@ -487,6 +505,22 @@ ly_ctx_unset_options(struct ly_ctx *ctx, uint16_t option)
 {
     LY_CHECK_ARG_RET(ctx, ctx, LY_EINVAL);
     LY_CHECK_ERR_RET(option & LY_CTX_NO_YANGLIBRARY, LOGARG(ctx, option), LY_EINVAL);
+
+    if ((ctx->flags & LY_CTX_SET_PRIV_PARSED) && (option & LY_CTX_SET_PRIV_PARSED)) {
+        struct lys_module *mod;
+        uint32_t index;
+
+        index = 0;
+        while ((mod = (struct lys_module *)ly_ctx_get_module_iter(ctx, &index))) {
+            /* ignore modules that didn't compile */
+            if (!mod->compiled) {
+                continue;
+            }
+
+            /* set NULL for all ::lysc_node.priv pointers in module */
+            lysc_module_dfs_full(mod, lysc_node_clear_priv_dfs_cb, NULL);
+        }
+    }
 
     /* unset the option(s) */
     ctx->flags &= ~option;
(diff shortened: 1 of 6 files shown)


src/context.c
```diff
@@ -473,25 +473,63 @@ ly_ctx_get_options(const struct ly_ctx *ctx)
...
-    ctx->flags |= option;
+    if (!lyrc) {
+        ctx->flags |= option;
+    }
+
+    return lyrc;
+}
 
+static LY_ERR
+lysc_module_priv2null(struct lysc_node *node, void *UNUSED(data), ly_bool *UNUSED(dfs_continue))
```
michalvasko, line 497: The custom is to use the `x2y` notation for conversion functions, that is not the case here, so please keep that in mind. Also, I prefer every callback to be clearly marked that way so I prefer a suffix `_cb`. So you could rename this to `lysc_node_clear_priv_dfs_cb` or something similar.

src/context.c
```diff
@@ -473,25 +473,63 @@ ly_ctx_get_options(const struct ly_ctx *ctx)
...
+
+        index = 0;
+        while ((mod = (struct lys_module *)ly_ctx_get_module_iter(ctx, &index))) {
+            /* ignore modules that didn't compile */
+            if (!mod->compiled) {
+                continue;
+            }
+
+            /* set NULL for all ::lysc_node.priv pointers in module */
+            if ((lyrc = lysc_module_dfs_full(mod, lysc_module_priv2null, NULL))) {
```
michalvasko, line 523: Since your callback never returns any error, you do not need to check the return value here at all.

src/context.h
```diff
@@ -161,8 +161,6 @@ struct ly_ctx;
  *
  * Options to change context behavior.
  *
- * Note that the flags 0xFF00 are reserved for internal use.
- *
```
michalvasko, line 165: Could have been in a separate commit but I take it that you checked this and found that there are no internal flags anymore.

src/tree_schema.h
```diff
@@ -72,7 +72,7 @@ struct lyxp_expr;
  * the default nodes are not added into any data tree and mandatory nodes are not checked in the data trees.
  *
  * The compiled schema tree nodes are able to hold private objects (::lysc_node.priv as a pointer to a structure,
- * function, variable, ...) used by a caller application.
+ * function, variable, ...) used by a caller application unless ::LY_CTX_SET_PRIV_PARSED is not set.
```
michalvasko, line 75: Should be `unless ::LY_CTX_SET_PRIV_PARSED is set`. But you can be more verbose here and at least briefly explain that if the option is set, the `priv` objects are actually filled by *libyang*.

tests/utests/basic/test_context.c
```diff
@@ -690,6 +690,154 @@ test_ylmem(void **state)
...
+            "  }\n"
+            "}\n";
+
+    /* use own context with extra flags */
+    ly_ctx_destroy(UTEST_LYCTX, NULL);
+    const char *feats[] = {"f1", NULL};
+    assert_int_equal(LY_SUCCESS, ly_ctx_new(NULL, LY_CTX_SET_PRIV_PARSED, &UTEST_LYCTX));
+    UTEST_ADD_MODULE(schema_a, LYS_IN_YANG, feats, NULL);
+
+    print_message("[          ] create context\n");
```
michalvasko, line 823: Um, not sure what I missed but I could not find this function declaration anywhere. In any case, I think you should remove it because it only clutters the test output.

michalvasko (changes requested):

tests/utests/basic/test_context.c
lePici: It should be cmock function. I added it there because, cmock asserts are done in function and not in sequence, as usual, so the failure information on line xxx is useless. The test consists of three phases. If I remove this marking, it will not be obvious at first glance at which phase it failed. However, if the test fails, gdb will probably be used anyway. I don't know which would be better, so take it as a suggestion.

src/context.h
lePici: Yes, I did not find any internal flags. I will make it a special commit, which will be the first in this PR.

tests/utests/basic/test_context.c
michalvasko: Okay, I suppose (although they could definitely have chosen a better name for the function), so it can stay there, if you want.

src/context.c
lePici: ok

src/context.c
lePici: ok

src/context.h
```diff
@@ -187,6 +185,12 @@ struct ly_ctx;
                                         directory, which is by default searched automatically (despite not
                                         recursively). */
 #define LY_CTX_PREFER_SEARCHDIRS 0x20 /**< When searching for schema, prefer searchdirs instead of user callback. */
+#define LY_CTX_SET_PRIV_PARSED 0x40 /**< For all compiled nodes, their private objects (::lysc_node.priv) are used
```
rkrejci, line 188: `ly_ctx_destroy()` should also reflect the new option. I would not blindly expect that it will be called without the destructor, let's just ignore the destructor callback in case this option is in use.

rkrejci (changes requested):

src/context.h
lePici: I also took it into consideration, but currently the destructor for private pointers is not used anyway.
[see here](https://github.com/CESNET/libyang/blob/619dc2388d3f76355842a21c1e56c2c098339ef2/src/tree_schema_free.c#L954)
But I should have added it there at least some if statement. So I'll add it.

src/tree_schema.h
lePici: ok

michalvasko (approved):

rkrejci (approved):

src/context.h
rkrejci: ok, after reviewing the mechanism of freeing `priv` pointers, it seems better to remove it from the context destructor, but let's make it in a separate PR #1501.

PR #1501, opened 2021-04-12
remove private_destructor parameter from ly_ctx_destroy()

rkrejci:
The mechanism was not reliable and not used often. libyang provides a function to traverse the schema and free the set data manually or the caller can maintain some information to free the set data directly.

The difference of the API in contrast to 1.x (and so far to 2.0) shouldn't be a big issue since the parameter was used rarely.

### DIFF
--- a/doc/transition.dox
+++ b/doc/transition.dox
@@ -103,6 +103,15 @@
  * introduce some kind of context lock to completely abandon any change of the YANG modules after starting work with the
  * instantiated data.
  *
+ * Another change to note is the removed destructor of private data (::lysc_node.priv) in ::ly_ctx_destroy(). The mechanism
+ * was not reliable with the context recompilation and the splitted parsed and compiled schema trees or the complexity of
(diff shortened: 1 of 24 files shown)


doc/transition.dox
```diff
@@ -126,6 +131,7 @@
  * ly_ctx_unset_searchdirs() | ::ly_ctx_unset_searchdir()           | Simplify API and instead of index numbers, work with the values themselves.
  * ly_ctx_set*()             | ::ly_ctx_set_options()               | API simplification.
  * ly_ctx_unset*()           | ::ly_ctx_unset_options()             | ^
+ * ly_ctx_destroy()          | ::ly_ctx_destroy()                   | The destructor callback parameter was removed, see .
```
michalvasko, line 134: Missing link, I guess?

michalvasko (changes requested):

doc/transition.dox
rkrejci: right, fixed

michalvasko (approved):

PR #1503, opened 2021-04-12
build BUGFIX fix for OpenBSD and exposing realpath()

choppsv1:
The openbsd sys/cdefs.h is doing arithmetic with the defines, so set to 1 rather than just defined.


michalvasko:
Okay, should not cause any problems.

PR #1504, opened 2021-04-13
tree data BUGFIX move defn of named union out of struct defn

choppsv1:
Defining a union inside a struct does not define that union globally in
C++. Thus one can't use "union lyd_any_value *foo" in C++.

### DIFF
--- a/src/path.h
+++ b/src/path.h
@@ -37,22 +37,29 @@ enum ly_path_pred_type {
 };
 
 /**
- * @brief Structure for holding one segment of resolved path on schema including simple predicates.
- * Is used as a [sized array](@ref sizedarrays).
+ * @brief Structure for simple path predicate.
+ */
+struct ly_path_predicate {
+    union {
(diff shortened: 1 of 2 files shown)


michalvasko:
Okay, I suppose, even though `path.h` is not a public header so it should not matter there for C++.

choppsv1:
> Okay, I suppose, even though `path.h` is not a public header so it should not matter there for C++.

Originally I actually only fixed the public one. Then went back to look for others for completeness sake, and I didn't notice path.h was't private. Consistency is good though. :)

choppsv1:
FWIW I think that we should probably enable -Wc++-compat at least on the public API. There are a bunch of cast from void * warnings which I guess would be fine for internal code, but for external APIs (e.g., inline macros I think are getting hit) it will break things.

rkrejci:
Do you see any other way to fix it besides `__typeof__`? There are things like this:
```
ARRAY = malloc(...);
```
and the macro so far does not need to know the type of ARRAY.

PR #1506, opened 2021-04-13
format test

rkrejci:
Make the format checking part of the test suit.

GitHub Action runners do not contain an environment with a sufficient `uncrustify` version (ubuntu 20.10 should work) and I'm not sure that compiling it from source is worth it, so for now the test is not running in Actions.


michalvasko (approved):
Not merging yet, installing suitable Uncrustify for github actions should be added.

PR #1507, opened 2021-04-13
c++ compatibility of public headers

rkrejci:
Rewrite macros to not use implicit casts - mentioned in #1504.

@choppsv1 - could you please check, that the changes help? I don't want to add `-Wc++-compat` directly to our build since the issues reported are fine from the C point of view and this is C code. I agree that the public headers are promising usability even for C++, so I tried to solve the issues there. Unfortunately, I'm not sure if there is a way to simply check only the (macros in) public headers.

### DIFF
--- a/src/plugins_types.c
+++ b/src/plugins_types.c
@@ -489,7 +489,7 @@ lyplg_type_validate_patterns(struct lysc_pattern **patterns, const char *str, si
                 return ly_err_new(err, LY_EVALID, LYVE_DATA, NULL, eapptag, patterns[u]->emsg);
             } else {
                 const char *inverted = patterns[u]->inverted ? "inverted " : "";
-                return ly_err_new(err, LY_EVALID, 0, NULL, eapptag,
+                return ly_err_new(err, LY_EVALID, LYVE_DATA, NULL, eapptag,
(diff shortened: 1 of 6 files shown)


choppsv1:
> Unfortunately, I'm not sure if there is a way to simply check only the (macros in) public headers.

How about compile unit tests with -Werror=c++-compat?

michalvasko (approved):

rkrejci:
> > Unfortunately, I'm not sure if there is a way to simply check only the (macros in) public headers.
> 
> How about compile unit tests with -Werror=c++-compat?

Then it covers the tests' code, so again false positives regarding the implicit cast of `void *` or conversion of `0` to enum values.

choppsv1:
> > > Unfortunately, I'm not sure if there is a way to simply check only the (macros in) public headers.
> > 
> > 
> > How about compile unit tests with -Werror=c++-compat?
> 
> Then it covers the tests' code, so again false positives regarding the implicit cast of `void *` or conversion of `0` to enum values in *.c.

I ran into the issue b/c FRR has a c++-compat unit test, so perhaps a dedicated unit test that includes the public API headers and uses the macros? Not perfect but perhaps better than nothing.

michalvasko:
Okay, such a "test" was added.

PR #1508, opened 2021-04-14
User types

michalvasko:
Implement basic ietf-inet-types and ietf-yang-types.

### DIFF
--- a/src/plugins_types/date_and_time.c
+++ b/src/plugins_types/date_and_time.c
@@ -1,0 +1,275 @@
+/**
+ * @file date_and_time.c
+ * @author Michal Vasko <mvasko@cesnet.cz>
+ * @brief ietf-yang-types date-and-time type plugin.
+ *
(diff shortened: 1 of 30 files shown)


src/plugins_types/date_and_time.c
```diff
@@ -0,0 +1,274 @@
...
+    uint32_t i;
+    char *exp_str;
+
+    for (i = 0; c[i]; ++i) {
+        if ((*str)[0] == c[i]) {
+            break;
+        }
+    }
+    if (!c[i]) {
+        exp_str = malloc(3 + (i - 1) * 7 + 1);
```
rkrejci, line 78: some comment explaining these magic numbers would be useful here

rkrejci (changes requested):

src/plugins_types/date_and_time.c
michalvasko: Okay, I have tried.

rkrejci (approved):

PR #1510, opened 2021-04-15
libyang DOC add brief thread information

michalvasko:


src/libyang.h
```diff
@@ -131,6 +131,22 @@ extern "C" {
...
+ * It is safe to read from ::ly_ctx structure concurrently and use its dictionary, which is protected by a lock.
+ * Thread-safe functions include any ones working with data trees and all the `ly_ctx_get_*()` functions. Generally,
+ * they are the functions with `const` context parameter.
+ *
+ * @section data Data Trees
+ *
+ * The general rule is always the same. Only the functions with non-const ::lyd_node parameters modify the node(s).
+ * No concurrent execution of such functions should be allowed on a single data tree or subtrees of one. Conforming
+ * to this, it should be safe to do everything else.
+ */
```
jktjkt, line 148: Disclaimer: I've never used libyang2 before, but I have some experience with the `1.x` branch of libyang. I think the intention is that there can only ever be one writer or N readers per each tree, but the way it's written may allow interpreting that as "no concurrent writers" while allowing a mixture of R+W workloads.

jktjkt (changes requested):

src/libyang.h
```diff
@@ -131,6 +131,22 @@ extern "C" {
...
+ *
+ * @section context Context
+ *
+ * It is safe to read from ::ly_ctx structure concurrently and use its dictionary, which is protected by a lock.
+ * Thread-safe functions include any ones working with data trees and all the `ly_ctx_get_*()` functions. Generally,
+ * they are the functions with `const` context parameter.
+ *
+ * @section data Data Trees
+ *
+ * The general rule is always the same. Only the functions with non-const ::lyd_node parameters modify the node(s).
```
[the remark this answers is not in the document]
michalvasko: Fair enough, removed :)

src/libyang.h
michalvasko: There are no theoretical thread-safety differences between *libyang* 1 and 2. However, there are several use-cases of *libyang* 1 not being thread-safe despite a reasonable expectation for it to be. That should be fixed in *libyang* 2.

src/libyang.h
```diff
@@ -131,6 +131,26 @@ extern "C" {
...
+/**
+ * @page howtoThreads Threading Limitations
+ *
+ * @section context Context
+ *
+ * It is safe to read from ::ly_ctx structure concurrently and use its dictionary, which is protected by a lock.
+ * Thread-safe functions include any ones working with data trees (only context dictionary is accessed) and all
+ * the `ly_ctx_get_*()` functions. Generally, they are the functions with `const` context parameter.
+ *
+ * @section schema Schema Trees
```
rkrejci, line 143: Not sure if there should be a separate section for schema trees. They are actually tightly connected with the context since schema modification can be actually done via context.

michalvasko: They were not there originally, I myself was not sure whether it makes sense or not. So I have removed it.

rkrejci (approved):

michalvasko:
@jktjkt Any more comments or is now clear enough?

jktjkt (approved):

PR #1511, opened 2021-04-15
lyb FEATURE thread-safe cached hash generation

michalvasko:


src/tree_data_internal.h
```diff
@@ -34,23 +34,29 @@ struct lysc_module;
 #define LY_LYB_SUFFIX_LEN 4
 
 /**
- * @brief Hash schema sibling to be used for LYB data.
+ * @brief Get single hash for a schema node to be used for LYB data. Read from cache, if possible.
  *
- * @param[in] sibling Sibling to hash.
+ * @param[in] node Node to hash.
  * @param[in] collision_id Collision ID of the hash to generate.
```
rkrejci, line 40: The `collision_id` seems like a kind of magic and I'm not able to find any explanation of its meaning.

rkrejci (commented):
It's kind of cryptic, but probably because I'm not familiar with the format.

src/tree_data_internal.h
michalvasko: Fair point, I have reorganized it.

rkrejci (approved):

PR #1512, opened 2021-04-15
TSAN: ignore races around lys_node.hash

jktjkt:
As per Michal's advice, this is tricky to fix without breaking the ABI. Given that the hashing is only an opportunistic speedup anyway, and that the race involves both threads writing the same deterministic value, let's just flag the cache manipulation as "ignored" by TSAN.

This first helper function relies upon the fact that a `LYB_HASH` of `0` means "invalid value". That's how the old code was effectively behaving, so I hope it's reasonable.

Fixes: https://github.com/sysrepo/sysrepo/issues/2323


michalvasko:
Okay, I suppose, although it is quite a messy solution.

PR #1514, opened 2021-04-19
Explicit compilation and recompilation refactoring

michalvasko:


src/context.c
```diff
@@ -485,7 +523,14 @@ ly_ctx_set_options(struct ly_ctx *ctx, uint16_t option)
 
     if (!(ctx->flags & LY_CTX_SET_PRIV_PARSED) && (option & LY_CTX_SET_PRIV_PARSED)) {
         ctx->flags |= LY_CTX_SET_PRIV_PARSED;
-        if ((lyrc = lys_recompile(ctx, 0))) {
+        if (ctx->flags & LY_CTX_EXPLICIT_COMPILE) {
+            /* compile it now so that the priv pointers are set */
+            lyrc = ly_ctx_compile(ctx);
```
rkrejci, line 539: does this work? Isn't the logic reverse here? Let's say that the context was already compiled, so there is no `to_compile` set, so this call will do nothing but we would like to recompile. If the context wasn't compiled, we probably don't need to compile it now, just by setting the option.

src/log.h
```diff
@@ -249,8 +249,10 @@ typedef enum
     LY_EINT,        /**< Internal error */
     LY_EVALID,      /**< Validation failure */
     LY_EDENIED,     /**< Operation is not allowed */
-    LY_EINCOMPLETE, /**< The operation did not failed, but for some reason it was not possible to finish it completely.
+    LY_EINCOMPLETE, /**< The operation did not fail, but for some reason it was not possible to finish it completely.
                          According to the specific use case, the caller is usually supposed to perform the operation again. */
+    LY_ERECOMPILE,  /**< The operation did not fail, but requires context recompilation before it can be completed.
```
rkrejci, line 254: so is this value now expected even in API functions adding modules into the context? If so, it should be probably mentioned in the description of `LY_CTX_EXPLICIT_COMPILE`

rkrejci (changes requested):

src/context.c
michalvasko: You are right, that was wrong. I have kept it the way it was so that only previously compiled modules will be recompiled, which is probably the best thing to do, anyway.

src/log.h
michalvasko: The only API function that can return `LY_ERECOMPILE` should be `lyplg_type_make_implemented()`. In that case it should be treated the same way as any other error value. In all other cases (internally) simply the `recompile` flag in global `unres` is set and it is checked at relevant places.

I wanted to unify the approach but checking some flag (of an internal structure currently) in a type plugin seemed to me like a bad idea, so I have added the error value. So I have tried using the error value through all of *libyang*, which resulted only in me wasting a few hours and it still not working. The problem mostly was that this is not a standard error value and should "bubble up" the stack only up to a certain point, where the context should be recompiled and compilation retried. Also, in the functions that are returning the error, they also needed to often behave specially. And trying to implement all this, there were just more and more functions with messy cleanups and finding new problems was taking too long. So I ended up with the current solution as the best, under the circumstances.

I want to refactor some functions later and ideally use the return value instead of the unres flag but I am not so sure anymore, whether I can come up with anything better.

rkrejci (approved):

PR #1522, opened 2021-04-23
printer tree FEATURE added various features

lePici:
Features:
1. printing subtree or node
2. printing submodule (yanglint also changed)
3. line length option added to yanglint
4. printing yang-data (extra changes in _context_, _schema_compile_, _tree_schema_tree_, _tree_schema_internal_)

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -570,6 +570,38 @@ lysc_node_clear_priv_dfs_cb(struct lysc_node *node, void *UNUSED(data), ly_bool 
     return LY_SUCCESS;
 }
 
+static void
+lysc_node_clear_all_priv(struct lys_module *mod)
+{
+    LY_ARRAY_COUNT_TYPE u, v;
+
+    if (mod->compiled) {
+        /* set NULL for all ::lysc_node.priv pointers in module */
+        lysc_module_dfs_full(mod, lysc_node_clear_priv_dfs_cb, NULL);
+    }
+
+    /* only lys_compile_extension_instance()
+     * can set ::lysp_ext_instance.parsed
+     */
+    if (mod->parsed) {
+        struct lysp_ext_instance *exts_p;
+        exts_p = mod->parsed->exts;
+        LY_ARRAY_FOR(exts_p, u) {
+            if (exts_p[u].parsed) {
+                /* lys_compile_extension_instance() was called */
+                struct lysc_ext_substmt *substmts;
+                struct lysc_node *root;
+                /* set NULL for all ::lysc_node.priv pointers in extensions */
+                substmts = mod->compiled->exts[u].substmts;
+                LY_ARRAY_FOR(substmts, v) {
+                    root = *(struct lysc_node **)substmts[v].storage;
+                    lysc_tree_dfs_full(root, lysc_node_clear_priv_dfs_cb, NULL);
+                }
+            }
+        }
+    }
+}
+
 API LY_ERR
 ly_ctx_unset_options(struct ly_ctx *ctx, uint16_t option)
 {
@@ -582,13 +614,7 @@ ly_ctx_unset_options(struct ly_ctx *ctx, uint16_t option)
 
         index = 0;
         while ((mod = (struct lys_module *)ly_ctx_get_module_iter(ctx, &index))) {
-            /* ignore modules that didn't compile */
-            if (!mod->compiled) {
-                continue;
-            }
-
-            /* set NULL for all ::lysc_node.priv pointers in module */
-            lysc_module_dfs_full(mod, lysc_node_clear_priv_dfs_cb, NULL);
+            lysc_node_clear_all_priv(mod);
         }
     }
 
--- a/src/printer_tree.c
+++ b/src/printer_tree.c
@@ -37,28 +37,28 @@
  * Manager functions are at the peak of abstraction. They are
  * able to print individual sections of the YANG tree diagram
  * (eg module, notifications, rpcs ...) and they call
- * Browse functions (\ref TRP_trb).
+ * Browse functions (@ref TRP_trb).
  *
  * @subsubsection TRP_trb trb
  * Browse functions contain a general algorithm (Preorder DFS)
  * for traversing the tree. It does not matter what data type
- * the tree contains (\ref lysc_node or \ref lysp_node), because it
+ * the tree contains (@ref lysc_node or @ref lysp_node), because it
  * requires a ready-made getter functions for traversing the tree
- * (\ref trt_fp_all) and transformation function to its own node
- * data type (\ref trt_node). These getter functions are generally
- * referred to as \ref TRP_tro. Browse functions can repeatedly
+ * (@ref trt_fp_all) and transformation function to its own node
+ * data type (@ref trt_node). These getter functions are generally
+ * referred to as @ref TRP_tro. Browse functions can repeatedly
  * traverse nodes in the tree, for example, to calculate the alignment
  * gap before the nodes \<type\> in the YANG Tree Diagram.
- * The obtained \ref trt_node is passed to the \ref TRP_trp functions
+ * The obtained @ref trt_node is passed to the @ref TRP_trp functions
  * to print the Tree diagram.
  *
  * @subsubsection TRP_tro tro
  * Functions that provide an extra wrapper for the libyang library.
  * The Obtain functions are further specialized according to whether
- * they operate on lysp_tree (\ref TRP_trop) or lysc_tree
- * (\ref TRP_troc). If they are general algorithms, then they have the
+ * they operate on lysp_tree (@ref TRP_trop) or lysc_tree
+ * (@ref TRP_troc). If they are general algorithms, then they have the
  * prefix \b tro_. The Obtain functions provide information to
- * \ref TRP_trb functions for printing the Tree diagram.
+ * @ref TRP_trb functions for printing the Tree diagram.
  *
  * @subsubsection TRP_trop trop
  * Functions for Obtaining information from Parsed schema tree.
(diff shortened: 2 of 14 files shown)


src/printer_tree.c
```diff
@@ -3402,6 +3402,60 @@ trb_print_nodes(struct trt_wrapper wr, struct trt_parent_cache ca, struct trt_pr
     } while (sibling_flag);
 }
 
+/**
+ * @brief Calculate the wrapper about how deep in the tree the node is.
+ * @param[in] node from which to count.
+ * @return wrapper for @p node.
+ */
+static struct trt_wrapper
+trb_count_immersion(const struct lysc_node *node)
```
michalvasko, line 3436: Took me a few second to understand, use `depth` instead of `immersion`.

tools/lint/cmd_print.c
```diff
@@ -185,6 +190,12 @@ cmd_print(struct ly_ctx **ctx, const char *cmdline)
             }
             break;
 
+        case 'L': /* --tree-line-length */
+            if (sscanf(optarg, "%zu", &line_length) != 1) {
```
michalvasko, line 194: Avoid `scanf` and any associated functions, they are often doing what they want and not what you expect. Instead, you can use `atoi()` when you do not care about an error that much or `strtoul()` when you do.

tools/lint/main_ni.c
```diff
@@ -468,6 +461,12 @@ fill_context(int argc, char *argv[], struct context *c)
             c->list = 1;
             break;
 
+        case 'L': /* --tree-line-length */
+            if (sscanf(optarg, "%zu", &c->line_length) != 1) {
```
michalvasko, line 465: Same, avoid `scanf`.

src/tree_schema_free.c
```diff
@@ -513,6 +513,10 @@ lysc_extension_free(struct ly_ctx *ctx, struct lysc_ext **ext)
 void
 lysc_ext_instance_free(struct ly_ctx *ctx, struct lysc_ext_instance *ext)
 {
+    if (ly_ctx_get_options(ctx) & LY_CTX_SET_PRIV_PARSED) {
```
michalvasko, line 516: Useless to call an API function internally and it is better to access the flags directly. In this case it does not matter because the function does not do anything but in general there may be some additional code that you usually do not want to execute internally (such as in `ly_ctx_set_options()`).

michalvasko (changes requested):

src/plugins_exts.h
```diff
@@ -130,14 +130,14 @@ void lyplg_ext_instance_substatements_free(struct ly_ctx *ctx, struct lysc_ext_s
  * function can be used to let the compilation to libyang following the standard rules for processing the YANG statements.
  *
  * @param[in] cctx Current compile context.
- * @param[in] p_ext Parsed extension instance data.
+ * @param[in,out] p_ext Parsed extension instance data. Item ::lysc_ext_instance.parsed can be modified.
```
michalvasko, line 133: Yes, it can, but I would pretend it cannot :) By passing a non-`const` parameter in a function you are saying it can be modified, but the user should definitely not modify this parsed structure and treat it as `const` instead. Yes, strictly speaking, it may be modified but only by our function and only by caching the parsed tree in it. So, in order for the `const` keyword to be useful, it may not always be treated that strictly. Please leave the `const` here and everywhere else.

src/plugins_exts_compile.h
```diff
@@ -99,14 +99,15 @@ const char *lysc_ctx_get_path(const struct lysc_ctx *ctx);
...
+ * @param[in,out] ext_p Parsed representation of the extension instance being processed.
+ * Item ::lysc_ext_instance.parsed will be set.
  * @param[in,out] ext Compiled extension instance with the prepared ::lysc_ext_instance.substmts array, which will be updated
  * by storing the compiled data.
  * @return LY_SUCCESS on success.
  * @return LY_EVALID if compilation of the substatements fails.
  * @return LY_ENOT if the extension is disabled (by if-feature) and should be ignored.
  */
-LY_ERR lys_compile_extension_instance(struct lysc_ctx *ctx, const struct lysp_ext_instance *ext_p, struct lysc_ext_instance *ext);
+LY_ERR lys_compile_extension_instance(struct lysc_ctx *ctx, struct lysp_ext_instance *ext_p, struct lysc_ext_instance *ext);
```
michalvasko, line 110: And leave `const` here as well, just use a cast inside the function to get rid of it.

src/context.c
```diff
@@ -583,13 +583,30 @@ ly_ctx_unset_options(struct ly_ctx *ctx, uint16_t option)
...
             /* set NULL for all ::lysc_node.priv pointers in module */
             lysc_module_dfs_full(mod, lysc_node_clear_priv_dfs_cb, NULL);
+
+            /* set NULL for all ::lysc_node.priv pointers in extensions */
+            exts = mod->compiled->exts;
+            LY_ARRAY_FOR(exts, u) {
+                substmts = exts[u].substmts;
+                LY_ARRAY_FOR(substmts, v) {
+                    if (LY_STMT_IS_DATA_NODE(substmts[v].stmt)) {
+                        root = *(struct lysc_node **)substmts[v].storage;
```
michalvasko, line 605: I do not think this is safe, in general. You can do it only if `lys_compile_extension_instance()` was called so it should probably set some flag that will be read here. Based on your other changes and my suggested ones, you can use the presence of `parsed` instead of a flag, since it can be set only by `lys_compile_extension_instance()`, which is the only proper way to get the compiled node structures.

src/tree_schema.h
```diff
@@ -523,7 +523,9 @@ struct lysp_stmt {
 struct lysp_ext_instance {
     const char *name;                       /**< extension identifier, including possible prefix */
     const char *argument;                   /**< optional value of the extension's argument */
-    LY_VALUE_FORMAT format;                /**< prefix format of the extension name/argument (::LY_VALUE_XML is YIN format) */
+    LY_VALUE_FORMAT format;                 /**< prefix format of the extension name/argument (::LY_VALUE_XML is YIN format) */
+    struct lysp_node *parsed;               /**< Simply parsed (unresolved) YANG schema tree for ::lyplg_ext.compile callback,
```
michalvasko, line 527: The comment should probably say that it serves as a cache and is filled only if `lys_compile_extension_instance()` is called. It should not be needed to mention that it must not be modified because that is true for all of these members and understood implicitly.

michalvasko (changes requested):

src/context.c
```diff
@@ -571,6 +571,51 @@ lysc_node_clear_priv_dfs_cb(struct lysc_node *node, void *UNUSED(data), ly_bool
     return LY_SUCCESS;
 }
 
+static void
+lysc_node_clear_all_priv(struct lys_module *mod)
+{
+    LY_ARRAY_COUNT_TYPE u, v;
+    /* lys_compile_extension_instance() was called */
+    ly_bool lcei_called = 0;
```
michalvasko, line 579: The comment is fine but the flag name does not need to be that specific, I would call it simply `ext_has_compiled_nodes` or something similar.

src/context.c
```diff
@@ -571,6 +571,51 @@ lysc_node_clear_priv_dfs_cb(struct lysc_node *node, void *UNUSED(data), ly_bool
...
+        LY_ARRAY_FOR(exts, u) {
+            substmts = exts[u].substmts;
+            LY_ARRAY_FOR(substmts, v) {
+                if (LY_STMT_IS_DATA_NODE(substmts[v].stmt)) {
+                    root = *(struct lysc_node **)substmts[v].storage;
+                    lysc_tree_dfs_full(root, lysc_node_clear_priv_dfs_cb, NULL);
+                }
+            }
+        }
+    }
```
michalvasko, line 616: This still does not seem safe to me. Just because you have found a single parsed extension that has the compiled nodes, you treat all of them that way? It must be specific for each parsed extension instance.

src/schema_compile.c
```diff
@@ -573,12 +573,27 @@ lys_compile_extension_instance(struct lysc_ctx *ctx, const struct lysp_ext_insta
                 case LY_STMT_NOTIFICATION:
                 case LY_STMT_RPC:
                 case LY_STMT_USES:
-                    r = lysp_stmt_parse(ctx, stmt, &parsed, NULL);
-                    LY_CHECK_ERR_GOTO(r, ret = r, cleanup);
+
```
michalvasko, line 576: Extra line.

src/schema_compile.c
```diff
@@ -573,12 +573,27 @@ lys_compile_extension_instance(struct lysc_ctx *ctx, const struct lysp_ext_insta
...
+                    if (!ext_p->parsed) {
+                        struct lysp_ext_instance *unconst_ext_p;
+                        r = lysp_stmt_parse(ctx, stmt, &parsed, NULL);
+                        LY_CHECK_ERR_GOTO(r, ret = r, cleanup);
+                        unconst_ext_p = (struct lysp_ext_instance *)ext_p;
+                        unconst_ext_p->parsed = parsed;
+                    } else {
+                        struct lysp_node *node, *last_node;
+                        /* get last parsed node */
+                        for (node = ext_p->parsed; node; node = node->next) {
```
michalvasko, line 586: You should be able to replace this with `LY_LIST_FOR`.

michalvasko (changes requested):

michalvasko (approved):

src/printer_tree.c
```diff
@@ -2418,20 +2417,53 @@ tro_modi_get_notifications(struct trt_tree_ctx *tc)
...
-    return TRP_EMPTY_KEYWORD_STMT;
+    assert(tc);
+    const void *node;
+    const char *yang_data_name;
+
+    if (tc->lysc_tree) {
+        struct lysc_ext_instance *exts;
+        struct lysc_ext_substmt *substmts;
+
+        assert(LY_ARRAY_COUNT(tc->pmod->exts) == LY_ARRAY_COUNT(tc->cmod->exts));
```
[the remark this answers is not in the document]
lePici: I just wanted to make sure that, for example, two parsed extensions do not result in some way to one compiled extension. Now that I've looked in more detail at the code, I don't believe it can happen. So I'll delete it.

rkrejci (approved):

PR #1525, opened 2021-04-26
data tree REFACTOR add canonical getter for lyd_value

choppsv1:
- inline canonical getter functions (helps reduce indirect linking
  requirements)
- fix return type of the plugin printer to be const char * (this
  avoids c++ compat warning regarding const void * -> const char *,
  but just seems right anyway)

### DIFF
--- a/src/tree_data.h
+++ b/src/tree_data.h
@@ -978,12 +978,37 @@ struct lyd_node *lyd_first_sibling(const struct lyd_node *node);
 int lyd_lyb_data_length(const char *data);
 
 /**
+ * @brief Get the (canonical) value of a lyd_value.
+ *
+ * @param[in] ctx Context for the value
+ * @param[in] value value node to use.
+ * @return Canonical value.
+ */
+const char *lyd_value_get_canonical(const struct ly_ctx *ctx, const struct lyd_value *value);
+
+/**
  * @brief Get the (canonical) value of a data node.
  *
  * @param[in] node Data node to use.
  * @return Canonical value.
  */
-const char *lyd_get_value(const struct lyd_node *node);
+static inline const char *
+lyd_get_value(const struct lyd_node *node)
+{
+    if (!node) {
+        return NULL;
+    }
+
+    if (!node->schema) {
+        return ((struct lyd_node_opaq *)node)->value;
+    } else if (node->schema->nodetype & LYD_NODE_TERM) {
+        const struct lyd_value *value = &((struct lyd_node_term *)node)->value;
+        return value->_canonical
+            ? value->_canonical
+            : lyd_value_get_canonical(LYD_CTX(node), value);
+    }
+    return NULL;
+}
 
 /**
  * @brief Get the (canonical) value of a metadata node.
@@ -991,7 +1016,17 @@ const char *lyd_get_value(const struct lyd_node *node);
  * @param[in] meta Metadata node to use.
  * @return Canonical value.
  */
-const char *lyd_get_meta_value(const struct lyd_meta *meta);
+static inline const char *
+lyd_get_meta_value(const struct lyd_meta *meta)
+{
+    if (meta) {
+        const struct lyd_value *value = &meta->value;
+        return value->_canonical
+            ? value->_canonical
+            : lyd_value_get_canonical(meta->annotation->module->ctx, value);
+    }
+    return NULL;
+}
 
 /**
  * @brief Get anydata string value.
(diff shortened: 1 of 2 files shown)


michalvasko:
Okay, the function `lyd_value_get_canonical()` probably can be added mostly to be used for the schema default values. But, the functions were not inlined on purpose to avoid having to include `plugins_types.h` in `tree_schema.h`, it is meant to be a separate header that needs to be included explicitly. Also, what exactly is the problem of the current code and C++ compatibility? Everything is placed into functions and the canonical value is always returned as `char *`. Plugins use `void *` on purpose, again, to warn users that it really can be a binary value, `char *` in C sort-of implies it is a string.

choppsv1:
The print routine can return a non char * value? Ok if so.

One issue with inlining, other than performance, had to do with things that link a library that links libyang. It's really not a big deal for me to add libyang again to the gcc ld lines I guess

However, I do think a lot of these getters being function calls instead of inline is not great for performance.

What is the gain to offset for the performance loss for not including "plugins_types.h".

FWIW I suspect libyang could do with a lot more inlining; function calls in dynamic linked code are not free, and cannot be optimized like inline code can be (e.g., often getters become no-ops b/c the surrounding code has already satisfied conditionals etc).

michalvasko:
> The print routine can return a non char * value? Ok if so.

Yes, if you choose the binary LYB format.

As for inlining, you seems to be a fan but unless you show me actual benchmarks showing a substantial performance gain or a  trustworthy source stating this, I would avoid using it. Generally, I prefer no code in headers and I do not consider myself smarter than the current compilers so I would leave any inlining up to them.

choppsv1:
> > The print routine can return a non char * value? Ok if so.
> 
> Yes, if you choose the binary LYB format.
> 
> As for inlining, you seems to be a fan but unless you show me actual benchmarks showing a substantial performance gain or a trustworthy source stating this, I would avoid using it. Generally, I prefer no code in headers and I do not consider myself smarter than the current compilers so I would leave any inlining up to them.

The compiler cannot optimize a non-inlined function, even if inlined in the linker. Substantial gain? It's not going to change the O() characteristics of your algorithms if that's what you mean. Millions of calls to indirect functions vs inline code though? Sure it's going to gain you something. Not sure I have the will to try and prove this to you though

michalvasko:
What could also be done is to keep only `lyd_value_get_canonical()` (as a standard function, only a wrapper for the print callback) and put back the macros for getting the canonical value of a term node or metadata, which would only decide between  returning the cached `_canonical` and calling `lyd_value_get_canonical`. That should solve all the problems, what do you think?

choppsv1:
> What could also be done is to keep only `lyd_value_get_canonical()` (as a standard function, only a wrapper for the print callback) and put back the macros for getting the canonical value of a term node or metadata, which would only decide between returning the cached `_canonical` and calling `lyd_value_get_canonical`. That should solve all the problems, what do you think?

Inlines represent almost a 20% improvement in performance over calling a function in a .so library.

Actual benchmark results: https://github.com/choppsv1/benchcall

michalvasko:
Thanks for the benchmark. So what about the solution I proposed, using macro instead of an inline function? In that case the function would always be called once at most and in that case the plugin callback needs to be called anyway.

Meaning replace `lyd_get_value()` by something like `LYD_GET_VALUE(node) (((struct lyd_node_term *)node)->value._canonical ? ((struct lyd_node_term *)node)->value._canonical : lyd_value_get_canonical())` with `NULL` checking. The same for metadata. And I am not sure about the name, whether `LYD_GET_VALUE` or as it was before `LYD_CANON_VALUE`, I am fine either way.

choppsv1:
> Thanks for the benchmark. So what about the solution I proposed, using macro instead of an inline function? In that case the function would always be called once at most and in that case the plugin callback needs to be called anyway.
> 
> Meaning replace `lyd_get_value()` by something like `LYD_GET_VALUE(node) (((struct lyd_node_term *)node)->value._canonical ? ((struct lyd_node_term *)node)->value._canonical : lyd_value_get_canonical())` with `NULL` checking. The same for metadata. And I am not sure about the name, whether `LYD_GET_VALUE` or as it was before `LYD_CANON_VALUE`, I am fine either way.

That is what I have in this PR except it uses inline functions (which are better by almost any measure) vs preprocessor macros. Why do you want to use #defines?

michalvasko:
Okay, I suppose, there can be inline functions instead of macros but the function `lyd_value_get_canonical()` should not be inline to avoid requiring the header `plugins_types.h`, like I said before.

choppsv1:
> Okay, I suppose, there can be inline functions instead of macros but the function `lyd_value_get_canonical()` should not be inline to avoid requiring the header `plugins_types.h`, like I said before.

Right, will push a new version

michalvasko:
Thanks, merging, I will fix the formatting and documentation myself.

PR #1526, opened 2021-04-27
LYB data DOC specific description of data stored in LYB format

rkrejci:
Change the doxygen of LYB to directly include type-specific details of storing values.


michalvasko:
Looks good, thanks.

PR #1531, opened 2021-04-29
distro FEATURE apkg-powered upstream rpm and deb packaging

jruzicka-nic:
Using this `.spec` file kindly provided by Martin Winter with some modifications and [apkg](https://pkg.labs.nic.cz/pages/apkg/) magic, I'm able to build both source package
    
    libyang2-1.999.8e9f8794-1.fc33.src.rpm

and native RPM packages

    libyang2-1.999.8e9f8794-1.fc33.x86_64.rpm
    libyang2-debuginfo-1.999.8e9f8794-1.fc33.x86_64.rpm
    libyang2-debugsource-1.999.8e9f8794-1.fc33.x86_64.rpm
    libyang2-devel-1.999.8e9f8794-1.fc33.x86_64.rpm
    libyang2-tools-1.999.8e9f8794-1.fc33.x86_64.rpm
    libyang2-tools-debuginfo-1.999.8e9f8794-1.fc33.x86_64.rpm

by running

    apkg build -i

on this commit. It should work on any RPM-based distro including SUSE.

I was able to install resulting packages on each testing system as well.

I've tested this on:
 
* Fedora 33 VM (generic/fedora33)
* Fedora 34 docker container (fedora:34)
* openSUSE 15 docker container (opensuse/leap)

distro/README.md contains (3 lines of) instructions on howto test this using apkg in a VM/container/machine of your choice.


michalvasko (changes requested):
Also, the idea is that the package name will be `libyang` but it will follow the convention that packages of libraries are suffixed with their SO version. That means that there will be `libyang2` package with the library itself, but `libyang-devel` of the development package. Please reflect this in the scripts, best if the SO version was read from the main CMakeList.

oerdnj:
> That means that there will be libyang2 package with the library itself, but libyang-devel of the development package.

This has one major problem - the API has changes substantially, so libyang2-dev isn't really a drop-in replacement for libyang-dev.  This is more like gtk2 -> gtk3 change, so we need to reflect this in the package names.

michalvasko:
> This has one major problem - the API has changes substantially, so libyang2-dev isn't really a drop-in replacement for libyang-dev. This is more like gtk2 -> gtk3 change, so we need to reflect this in the package names.

Well, if there is a rule that it cannot be called the same then there is nothing we can do. But then there must be a conflict with `libyang-dev` package because they share almost all the files.

oerdnj:
The deb package already has a proper conflicts field in place

jruzicka-nic:
Thanks for review! I've removed unnecessary code you've pointed out from `libyang2.spec`, fixed some nits, and also added `Conflicts:` to enable installation on a system with `libyang`/ `libyang-devel` installed.

However this requires `--allowerasing` argument to `dnf install` and it's not a proper solution as [Fedora  guidelines](https://fedoraproject.org/wiki/Packaging:Conflicts#Acceptable_Uses_of_Conflicts:) state that `as a general rule, Fedora packages must NOT contain any usage of the Conflicts: field`.

I'll look at proper solutions next week and also at integrating Ondřej's `libyang2-debian-packaging` into `libyang2-apkg`.

jruzicka-nic:
> Also, the idea is that the package name will be `libyang` but it will follow the convention that packages of libraries are suffixed with their SO version. That means that there will be `libyang2` package with the library itself, but `libyang-devel` of the development package. Please reflect this in the scripts, best if the SO version was read from the main CMakeList.

As Ondra pointed out, such approach isn't viable for ABI-breaking 2.0 as it would break everything using old `libyang` pkgs on upgrade to new ones.

I think it's necessary to prefix new packages `libyang2` including `libyang2-devel` and `libyang2-tools` (which collide with tools in `libyang` btw)

Depending packages should require major libyang version explicitly - either `libyang` (1.0) or `libyang2` (2.0) to maintain compatibility.

michalvasko (changes requested):

mwinter-osr:
Started playing with this and in general it seems to work well.
One small change is for the version as there is a version tag now. The `tools/make-dev-archive.sh` would need to be 
adjusted, but the current suggest way is wrong (version tag has a prepended `v` and the `-` in the git describe are not
allowed in version numbers)

I suggest the following change:
```
diff --git a/tools/make-dev-archive.sh b/tools/make-dev-archive.sh
index afd7d7b..2edfd69 100755
--- a/tools/make-dev-archive.sh
+++ b/tools/make-dev-archive.sh
@@ -2,10 +2,10 @@
 # create archive from current source using git
 
 # once v2.0.0 tag is released consider using
-#VERSION=$(git describe --tags --always)
-
-# temporary hack before 2.0.0 release
-VERSION=1.999.$(git rev-parse --short HEAD)
+VERSION=$(git describe --tags --always)
+# skip "v" from start of version number (if it exists) and replace - with .
+VERSION=${VERSION#v}
+VERSION=${VERSION//[-]/.}
 
 NAMEVER=libyang2-$VERSION
 ARCHIVE=$NAMEVER.tar.gz
```

One more issue is that the Debian symbols (`libyang2.symbols`) in the debian directory need to be updated to match the released version.

As a sidenote, our CI infrastructure has the git pulling and the package building on different machines. Building directly from git is therefor painful. (I do this as github is frequently unreliable and has some rate-limits. This limits the required git access to a single machine per CI run and not one per distro)

with `apkg` I do on the first box (which does the download):
`apkg make-archive`
then grab the `libyang2-2.0.0.4.g2b7ef0d.tar.gz` from the `pkg/archives/dev/` directory to build on all the distros.

On each distro box, I just unpack that tar.gz and then run
`apkg build -i --archive ../libyang2-2.0.0.4.g2b7ef0d.tar.gz`

This seems to work reasonable well so far. Now testing on more platforms...

jruzicka-nic:
@mwinter-osr awesome, you even figured out the trick with `--archive` 🚀

I'll integrate your nice `make-dev-archive.sh` changes which result in proper version string and archive name. I'll also update symbols file and other nits Michal found soon-ish.

Seeing your archive-based workflow, I think you might be interested in `-u`/`--upstream` mode which uses `distro/` files from archive as opposed to those in current project dir's `distro/` so you don't need to unpack the archive before build:

```
apkg build --upstream -i -a path/to/archive.tgz
```
This isn't really used anywhere yet because it requires `distro` files in project archives so it might not work perfectly but this workflow is definitely supported so I'll give it a test spin and fix any issues I hit - feel free to test that too if you want.

Furthermore, if you see `apkg` behaving in sub-optimal way please do let me know I'm eager to make it state-of-the-art upstream packaging tool.

Finally, I plan to eventually add libyang2 to [apkg integration tests](https://gitlab.nic.cz/packaging/apkg/-/merge_requests/47) for extra CI coverage and some safety apkg keeps working for libyang2 during its active development.

mwinter-osr:
Before merging this, please check PR #1539 with my comment on the debhelper.
I hope we can lower it as commented there, but Ondrej may have some reason on why he moved it to 10.

Still testing more platforms, but with the debhelper change (and forcing Jinja2<3 for apkg PIP dependency it seems to work on my oldest Ubuntu 16.04 32bit platform.

I'll expect to finish testing it on all the platforms I care about FRR within the next 12hrs.

distro/pkg/deb/tests/control
```diff
@@ -0,0 +1,7 @@
+Test-Command: set -e ; for py in $(py3versions -r 2>/dev/null) ; do cd "$AUTOPKGTEST_TMP" ; echo "Testing with $py:" ; $py -c "import yang; print(yang)" ; done
+Depends: python3-all,
+         python3-yang
```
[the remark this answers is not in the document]
jruzicka-nic: Removed, thanks.

distro/pkg/rpm/libyang2.spec
```diff
@@ -0,0 +1,87 @@
...
+%package -n libyang2-devel
+Summary:    Headers of libyang library
+Conflicts:  libyang-devel
+Requires:   %{name} = %{version}-%{release}
+Requires:   pcre2-devel
+
+%package -n libyang2-tools
+Summary:    Helper Tools and examples for libyang library
+Conflicts:  libyang
+Provides:   linyang-tools
```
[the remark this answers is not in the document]
jruzicka-nic: 👍🏽

jruzicka-nic:
This revision is rebased on latest `libyang2` and:

* `deb/tests`: remove python
* `deb/control`: fix typo linyang -> libyang
* `deb/control`: depend on any debhelper version
* `deb/compat`: use 9 instead of 10 to support older debian distros
* `tools/make-dev-archive.sh`: generate proper version from version tag (still commented, ready for first version tag)

I've also tested `-u`/`--upstream` option and it seems to work well - you can do `apkg build -uia libyang2-1.999.78beb147.tar.gz ` in any directory and results will be in `pkg` as usual.

jruzicka-nic:
This revision is rebased on latest `libyang2` and:

* `tools/make-dev-archive.sh`: generate proper version from version tag based on `git describe`
* `deb/control`: fix typo linyang -> libyang
* `deb/control`: depend on any debhelper version
* `deb/compat`: use 9 instead of 10 to support older debian distros
* `deb/tests`: remove python
* `deb/libyang2.symbols`: update symbols

I've also tested `-u`/`--upstream` option and it seems to work well - you can do `apkg build -uia libyang2-1.999.78beb147.tar.gz ` in any directory and results will be in `pkg` as usual.

This worked for me on Ubuntu 16.04 and 20.04 with 18.04 still producing man-related error as described above.

jruzicka-nic:
Regarding `debhelper` compatibility level (`debian/compat`) - I'm currently using 9 for all upstream packaging because Debian 9 and Ubuntu 16.04 are still in LTS. `libyang` seems to work fine on older distros and I found no particular benefit for using newer debhelper in this case.

If we find a reason to bump in the future we can, especially after Debian 9 LTS end on 2022-01-01.

jruzicka-nic:
I've changed `make-dev-archive` and others to mimic [upstream release](https://github.com/CESNET/libyang/releases/tag/v2.0.0) so that `apkg` works with both dev and upstream archives.

* upstream archive: `libyang-2.0.0.tar.gz` (root dir: `libyang-2.0.0/`)
* dev archive: `libyang-2.0.0.10.ge0e8e204.tar.gz` (root dir: `libyang-2.0.0.10.ge0e8e204/`)

I've also added [apkg get-archive support for Content-Disposition headers](https://gitlab.nic.cz/packaging/apkg/-/merge_requests/51) in order to download GitHub archives with proper names and released `apkg-0.0.5`.

With `apkg-0.0.5` (upgrade with `pip3 install -U apkg`) and these changes:

* apkg is able to detect latest upstream version using new tiny `scripts/upstream-version.sh` script
* `apkg get-archive` downloads latest upstream archive: `pkg/archives/upstream/libyang-2.0.0.tar.gz`
* `apkg get-archive -v 2.0.0` downloads upstream archive of version 2.0.0
* `apkg build --archive $ARCHIVE` works on both dev and upstream archives
* `apkg build --upstream` will download upstream archive and build package exclusively from the archive (including `distro/` files) as soon as upstream release contains `distro/` files (IOW first release after this is merged)

To test build from upstream tarball:

```
apkg get-archive | apkg build --archive -F -
```

jruzicka-nic:
@michalvasko I've adapted this for the actual release `libyang-2.0.0` archive as provided by GitHub and added apkg upstream functionality and I think it's ready to merge unless someone finds issues with this ᕕ( ᐛ )ᕗ

jruzicka-nic:
Oh, and...  congratz on the `2.0.0` 🎉 🥳 🎉

oerdnj:
I really disagree with the debcompat change to level 9

jruzicka-nic:
> I really disagree with the debcompat change to level 9

I'm fairly neutral on the issue. From my POV level 9 works on all distros I'd bother testing on today including old ones, level 10 may require specific steps on older distros (like enabling backports) but can be made to work.

This is up to upstream and the Debian maintainer so let me know if I should update the PR back to 10.

Once this is merged we can fine-tune details in individual PRs.

jruzicka-nic:
FYI I'm thinking about a mechanism in `apkg` to assign proper debhelper compat level through template macro like `{{ debhelper_level }}` based on a distro it's run on but I'm not sure such approach would be without negative consequences related to things behaving differently on different compat levels 🤔

@oerdnj do you think that's legit for upstream packaging with single source for many distro releases or it's better to simply use a set compat level like we do now?

Also what are your reasons for strong preference for compat level 10?

oerdnj:
Even Ubuntu 16.04 (which is recently EOL have debhelper 10 in backports), there's no reason to not have at least debhelper 10 everywhere now.

It's not a runtime requirement, just build time, so I really don't know what's the problem here. We are pushing the bleeding edge to old distro, so if people want to build their own packages on EOL system they should be made jump through some hoops.

jruzicka-nic:
@oerdnj That makes sense, I've reverted to original `debhelper >= 10`.

michalvasko:
> I think it's ready to merge unless someone finds issues with this ᕕ( ᐛ )ᕗ

Great, that is what I was waiting for and if @mwinter-osr has no issue with debhelper 10, it seems you are all in agreement it is ready. I am just wondering why there are some things hard-coded (such as the exported symbols) and not generated instead. I do not really care but it seems to just add redundant work for the maintainers.

jruzicka-nic:
> > I think it's ready to merge unless someone finds issues with this ᕕ( ᐛ )ᕗ
> 
> Great, that is what I was waiting for and if @mwinter-osr has no issue with debhelper 10, it seems you are all in agreement it is ready. I am just wondering why there are some things hard-coded (such as the exported symbols) and not generated instead. I do not really care but it seems to just add redundant work for the maintainers.

Oh yes the symbols file is my least favorite aspect of Debian packaging. It's not strictly necessary ("Maintainers should not drop symbols file directly into Debian without understanding the mechanism for packaging these symbols files.") but projects tend to bother with it so I assume compat advantages overweight the effort required to keep them up-to-date.

Other such seemingly redundant things are often a legacy of the packaging format - RPM a deb formats are over 20 years old. Some things were improved and some simply remained the same for the better or worse.

Regarding debhelper compat - I think @mwinter-osr simply wanted packages to build. With a known way to install debhelper 10 on older distros and a fair argument against 9 from the Debian maintainer I think we are good to go with 10.

This is just the initial packaging, it can be adjusted as needed in the future as any other part of libyang code.

michalvasko:
Alright then, merging. So it seems #1539 is included and can be closed but I will leave that up to @oerdnj.

distro/pkg/deb/copyright
```diff
@@ -0,0 +1,178 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: libyang
+License: BSD-3-clause
+
+Files: *
+Copyright: 2015-2018 by CESNET, z.s.p.o.
```
michalvasko, line 6: Fix year.

distro/pkg/deb/copyright
```diff
@@ -0,0 +1,178 @@
...
+License: BSD-3-clause
+
+Files: swig/*
```
michalvasko, line 9: Non-existing.

distro/pkg/deb/copyright
```diff
@@ -0,0 +1,178 @@
...
+Copyright: 2017-2018 by Deutsche Telekom AG
+License: BSD-3-clause
+
+Files: swig/java/*
```
michalvasko, line 13: Non-existing.

distro/pkg/deb/copyright
```diff
@@ -0,0 +1,178 @@
...
+Copyright: 2018 by Enbo Zhang <zhangenbo@huawei.com>
+License: BSD-3-clause
+
+Files: swig/python/config.py.in swig/python/tests/test_libyang.py swig/python/tests/test_tree_data.py swig/python/tests/test_tree_schema.py
```
michalvasko, line 17: Non-existing.

distro/pkg/deb/copyright
```diff
@@ -0,0 +1,178 @@
...
+ WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+Files: swig/cpp/tests/microtest.h
```
michalvasko, line 65: Non-existing.

distro/pkg/deb/copyright
```diff
@@ -0,0 +1,178 @@
...
+ .
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+
+Files: swig/java/tests/hamcrest-core-1.3.jar
```
michalvasko, line 87: Non-existing.

distro/pkg/deb/copyright
```diff
@@ -0,0 +1,178 @@
...
+Copyright: Copyright 2000-2018 hamcrest.org
+License: BSD-3-clause
+
+Files: swig/java/tests/junit-4.12.jar
```
michalvasko, line 91: Non-existing.

distro/pkg/deb/copyright
```diff
@@ -0,0 +1,178 @@
...
+ This Agreement is governed by the laws of the State of New York and the intellectual property laws of the United States of America. No party to this Agreement will bring a legal action under this Agreement more than one year after the cause of action arose. Each party waives its rights to a jury trial in any resulting litigation.
+
+Files: swig/*.deb
```
michalvasko, line 138: Non-existing.

distro/pkg/deb/copyright
```diff
@@ -0,0 +1,178 @@
...
+Copyright:
+ Copyright (c) 1995-2011 The SWIG Developers
+ Copyright (c) 2005-2006 Arizona Board of Regents (University of Arizona).
+ Copyright (c) 1998-2005 University of Chicago.
+ Copyright (c) 1995-1998 The University of Utah and the Regents of the University of California
+License: GPL-3
+ Please refer to /usr/share/common-licenses/GPL-3 on Debian systems.
+
+Files: debian/*
```
michalvasko, line 147: Non-existing, maybe update path? But can probably be removed.

michalvasko (changes requested):

distro/pkg/deb/copyright
jruzicka-nic: Yeah, I think this is out of date - removed.

jruzicka-nic:
@michalvasko eagle eye 👁 I've removed all swig/java refs from `copyright` and also fixed path to linenoise.

I've compared `cme update dpkg-copyright` output with updated state and found no discrepancies.

tools/upstream-version.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# get latest upstream libyang version from github
+
+RLS_URL=https://api.github.com/repos/CESNET/libyang/releases/latest
```
michalvasko, line 4: I am not sure this will work. There will definitely be at least one more *libyang* 1 release, which will be marked as latest instead, no?

michalvasko:
Sorry for not noticing this all before, but this should be the very last issue.

tools/upstream-version.sh
jruzicka-nic: Good point. In that case I recommend updating the script to whatever you think will work... But I don't think it's really a blocker, it's more of a bonus functionality. You can always override autoversion using `-v`/`--version` options.

You can update the script first thing after merging this ;) Or if you have a solution in mind let me know and I'll update the MR once more...

Either way I'd really like to start next week with this already merged if at all possible so that individual issues may be addressed individually.

jruzicka-nic:
> Sorry for not noticing this all before, but this should be the very last issue.

No problem, reviews are iterative by nature.

tools/upstream-version.sh
michalvasko: > so that individual issues may be addressed individually.

Right, I just wanted not to merge something we know has an issue already. And the correct fix should probably be to get all the tags and find the highest, no? Should not be too difficult.

michalvasko (approved):

jruzicka-nic:
`upstream-version.sh` was updated by Michal to work with upstream release scheme and I've also added `apkg.compat = 1` to `apkg.toml` to allow smooth apkg config changes without disruption in the future.

Issue #1528, opened 2021-04-28
libyang2 fails to load modules


choppsv1:
With the following commit 1ccbf54

my FRR unit tests are failing in the following way:

2021/04/22 10:07:35 libyang: Invalid argument parent || module || ext (lys_getnext_()).
2021/04/22 10:07:35 libyang: Not found node "lib" in path. (Schema location /frr-routing:routing/control-plane-protocols/control-plane-protocol/frr-bgp:bgp/global/afi-safis/afi-safi/ipv4-flowspec/flow-spec-config/interface.)
2021/04/22 10:07:35 [Y3P23-A91R3][EC 100663318] yang_module_load: failed to load data model: frr-bgp

In order to reproduce one can run FRR unit test:

```
git clone https://github.com/LabNConsulting/frr.git
cd frr
git checkout chopps/ly2-triage
autoreconf -i
mkdir build
cd build
CPPFLAGS="$(pkg-config --cflags-only-I libyang)" ../configure --prefix=/usr --enable-dev-build --localstatedir=/var/run/frr --sbindir=/usr/lib/frr --sysconfdir=/etc/frr --enable-vtysh --enable-pimd --enable-sharpd --enable-snmp=agentx --enable-multipath=64 --enable-user=frr --enable-group=frr --enable-vty-group=frrvty --with-pkg-extra-version=-my-manual-build
make -j16
make check #<-- it will run and things will fail, then the specific test to run is:
tests/bgpd/.libs/test_peer_attr
```

I added an abort to the CHECK_RET macro so it would stop at the failure point and then I turned all debugging on in libyang and reran the test. I've attached that large debug output.

[ly2-debug.txt](https://github.com/CESNET/libyang/files/6391865/ly2-debug.txt)

choppsv1:
If you would also like to enable debug in the test program here is the diff to do that:
```
modified   tests/bgpd/test_peer_attr.c
@@ -18,6 +18,7 @@
  */
 #include <zebra.h>
 
+#include "debug.h"
 #include "memory.h"
 #include "plist.h"
 #include "bgpd/bgpd.h"
@@ -1459,6 +1460,8 @@ int main(void)
 	struct test_peer_attr *pa, *pac;
 	struct listnode *node, *nnode;
 
+	DEBUG_ON(&nb_dbg_events);
+	yang_debugging_set(1);
 	bgp_startup();
 
 	pa_list = list_new();
```

michalvasko:
[...]

PR #1533, opened 2021-04-30
context FEATURE callback for every compiled module

michalvasko:
Note that even when a single context change is being performed (new module compiled, feature changed), the callback may be called several times for the same module.

@choppsv1
Refs #1528

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -300,8 +300,12 @@ struct ly_ctx {
     struct dict_table dict;           /**< dictionary to effectively store strings used in the context related structures */
     struct ly_set search_paths;       /**< set of directories where to search for schema's imports/includes */
     struct ly_set list;               /**< set of loaded YANG schemas */
+
     ly_module_imp_clb imp_clb;        /**< Optional callback for retrieving missing included or imported models in a custom way. */
     void *imp_clb_data;               /**< Optional private data for ::ly_ctx.imp_clb */
+    ly_module_compiled_clb compiled_clb; /**< Optional callback called for every compiled module. */
+    void *compiled_clb_data;          /**< Optional private data for ::ly_ctx.compiled_clb */
+
     uint16_t change_count;            /**< Count of changes of the context, on some changes it could be incremented more times */
     uint16_t flags;                   /**< context settings, see @ref contextoptions. */
     pthread_key_t errlist_key;        /**< key for the thread-specific list of errors related to the context */
--- a/src/context.c
+++ b/src/context.c
@@ -627,6 +627,26 @@ ly_ctx_get_module_imp_clb(const struct ly_ctx *ctx, void **user_data)
     return ctx->imp_clb;
 }
 
+API ly_module_compiled_clb
+ly_ctx_get_module_compiled_clb(const struct ly_ctx *ctx, void **user_data)
+{
+    LY_CHECK_ARG_RET(ctx, ctx, NULL);
+
+    if (user_data) {
+        *user_data = ctx->compiled_clb_data;
+    }
+    return ctx->compiled_clb;
+}
+
+API void
+ly_ctx_set_module_compiled_clb(struct ly_ctx *ctx, ly_module_compiled_clb clb, void *user_data)
+{
+    LY_CHECK_ARG_RET(ctx, ctx, );
+
+    ctx->compiled_clb = clb;
+    ctx->compiled_clb_data = user_data;
+}
+
 API const struct lys_module *
 ly_ctx_get_module_iter(const struct ly_ctx *ctx, uint32_t *index)
 {
--- a/src/context.h
+++ b/src/context.h
@@ -408,6 +408,33 @@ ly_module_imp_clb ly_ctx_get_module_imp_clb(const struct ly_ctx *ctx, void **use
 void ly_ctx_set_module_imp_clb(struct ly_ctx *ctx, ly_module_imp_clb clb, void *user_data);
 
 /**
+ * @brief Callback called for each (re)compiled module in a context.
+ *
+ * @param[in] mod Implemented and compiled module.
+ * @param[in] user_data Arbitrary user data.
+ */
+typedef void (*ly_module_compiled_clb)(const struct lys_module *mod, void *user_data);
+
+/**
+ * @brief Get callback for compiled or recompiled modules in a context.
+ *
+ * @param[in] ctx Context to read from.
+ * @param[out] user_data Optional pointer for the set user data.
+ * @return Callback or NULL if not set.
+ */
+ly_module_compiled_clb ly_ctx_get_module_compiled_clb(const struct ly_ctx *ctx, void **user_data);
+
+/**
+ * @brief Set callback called for each compiled or recompiled module in a context. It can be used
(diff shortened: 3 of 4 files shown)


choppsv1:
Testing now.

michalvasko:
So is it okay?

choppsv1:
> So is it okay?

Sorry it didn't work drop in replacement for me, and neither did the compile at the very end trick. I haven't had time to debug this though. Could be something on my end, but probably not for the latter issue (the explicit compile is pretty straight-forward and shouldn't have led to failures in testing I think).

michalvasko:
I see. Could be more specific about what exactly happened? When using the callback, at the end you still had some module nodes without their `priv` pointer set? And the same when using the explicit compilation, after calling `ly_ctx_compile()` and setting all the pointers, some were still not set?

choppsv1:
> I see. Could be more specific about what exactly happened? When using the callback, at the end you still had some module nodes without their `priv` pointer set? And the same when using the explicit compilation, after calling `ly_ctx_compile()` and setting all the pointers, some were still not set?

I've rechecked things and I am successfully using the explicit compile now. Will try and figure out what's different with the callback.

choppsv1:
> > I see. Could be more specific about what exactly happened? When using the callback, at the end you still had some module nodes without their `priv` pointer set? And the same when using the explicit compilation, after calling `ly_ctx_compile()` and setting all the pointers, some were still not set?
> 
> I've rechecked things and I am successfully using the explicit compile now. Will try and figure out what's different with the callback.

Sorry I was wrong.. I had only enabled some of the explicit code, but no the actual passing of the flag when creating the context..  So here is an example of a failure when I enable explicit compile:
```
2021/05/04 18:27:55 PBR: [X1FAK-PXRDY] XXXX nmodules: 2
2021/05/04 18:27:55 PBR: [TXF4T-RPRMM] loading frr-filter
2021/05/04 18:27:55 PBR: [TXF4T-RPRMM] loading frr-interface
2021/05/04 18:27:55 PBR: [ZYP1G-DR2ZA] XXXX before explicit compile
2021/05/04 18:27:55 PBR: libyang: Invalid argument parent || module || ext (lys_getnext_()).
2021/05/04 18:27:55 PBR: libyang: Not found node "lib" in path. (Schema location /frr-interface:lib/interface/vrf.)
```

That error message looks familiar to something you recently fixed; however, I rebased the callback PR on the most recent libyang2 branch HEAD.

```
39b8580a * pr/1533 context FEATURE callback for every compiled module
b287b21e * upstream/libyang2 xml BUGFIX regard to utf8 in lyxml_parse_value()
32874e17 * build BUGFIX fix BSD builds, again, and cleanup
...
```

choppsv1:
So I actually can reproduce the problem with the simpler "make check" test. Basically I see the same `Invalid Argument` error. I noticed that the test only explicitly loaded it's own module; however, this test module (`frr-test-module.yang`) imports `frr-interface` which the path is trying to access. So I added explicitly loading frr-interface.yang prior to loading frr-test-module.yang, I then got an error about a path in frr-vrf.yang; frr-vrf.yang is imported from frr-interface.yang. So I add explicitly loading frr-vrf.yang. At this point the test passes.

choppsv1:
> So I actually can reproduce the problem with the simpler "make check" test. Basically I see the same `Invalid Argument` error. I noticed that the test only explicitly loaded it's own module; however, this test module (`frr-test-module.yang`) imports `frr-interface` which the path is trying to access. So I added explicitly loading `frr-interface.yang` prior to loading `frr-test-module.yang`, I then got an error about a path in `frr-vrf.yang`; `frr-vrf.yang` is imported from `frr-interface.yang`. So I add explicitly loading `frr-vrf.yang`. At this point the test passes.

That was for explicitly compile. For the callback the make check test `bgpd/test_peer_attr` is failing. In this case I'm first called back for `frr-bgp`. `frr-bgp` is an interesting module since it basically augments other modules (e.g., `frr-routing`), so in my compiled callback I'm trying to access paths that through these other modules to get to the augment (e.g., `/frr-routing:routing/control-plane-protocols/control-plane-protocol/frr-bgp:bgp`); however, I have not been called back yet for `frr-routing` which is probably why it's failing to do something with that path.

In this case I was able to fix the problem by changing the order of loading modules so that all the augmented modules are loaded prior to loading bgp. I'm thinking that libyang should have compiled those imported modules prior to compiling bgp rather than saying it has finished compilng bgp w/o compilng the modules that bgp imports.

michalvasko:
Thanks for looking at this further and what you are saying makes sense. Realizing this only recently, the way it is implemented now a module can be compiled even before its own imports are, which has many advantages (and likely allows for further simplification of compilation but I do not want to change that before the release to avoid breaking things). But for this reason the callback allows only accessing the compiled nodes of the specific module without relying on any other modules being compiled. There is also the possible problem of disabled nodes (false `if-feature`) being still present in the module. To improve the callback (so that it is called only after all the modules were compiled), there would some additional mechanisms be needed, which I would strongly prefer to avoid.

So I would suggest you using the explicit compilation and I will try to fix any problems with that as long as you can provide some way to reproduce them, even a failing FRR test should be fine.

choppsv1:
@michalvasko did you see that I identified the issue with explicit compile as well?

> Basically I see the same Invalid Argument error. I noticed that the test only explicitly loaded it's own module; however, this test module (frr-test-module.yang) imports frr-interface which the path is trying to access. So I added explicitly loading frr-interface.yang prior to loading frr-test-module.yang, I then got an error about a path in frr-vrf.yang; frr-vrf.yang is imported from frr-interface.yang. So I add explicitly loading frr-vrf.yang. At this point the test passes.

michalvasko:
I did but it did not make sense to me so I was hoping you would describe it in more detail. So you have module `frr-test-module.yang` that has some path (assuming `leafref` or `augment`) to `frr-interface.yang`. If you use (simplified) `ly_ctx_load_module(frr-test-module)` and then `ly_ctx_compile()` (having specified `LY_CTX_EXPLICIT_COMPILE` when creating the context), you should finally get both the modules fully compiled in the context and be free to fill all the `priv` pointers because the path will cause the other module to become implemented and to compile as well. Based on what you said it seems this is not happening and if you can confirm this, I will create a test case exactly as I described and investigate more.

michalvasko:
Seems it is not needed after all.

choppsv1:
> Seems it is not needed after all.

Didn't say that. But it's broken so it's not useful.

michalvasko:
When writing it, I did not know about your extra requirement to be able to access other compiled modules, it would have worked otherwise.

choppsv1:
To be clear I don't know if it's all imports, but certainly the ones that the module in question augments with its own schema. The point of the callback was to know when it was safe to add priv data to the schema defined in the module (including it's augments).

michalvasko:
One of the differences of compilation in *libyang* version 2 is handling any augments or deviations as belonging to the target module, not the original one because it simplifies the compilation and actually makes more sense. So the callback does not guarantee that all the augments in your module are applied but it does guarantee that all the augments from all the other modules targeting your module are applied.

PR #1534, opened 2021-05-01
build BUGFIX fix BSD builds, again, and cleanup

choppsv1:
- just define _GNU_SOURCE and include sys/cdefs.h this gets everything
set the right way (no need to define _POSIX_C_SOURCE, etc).
- BSD requires more includes to get to sockaddr and AF_*

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -11,14 +11,13 @@
  *
  *     https://opensource.org/licenses/BSD-3-Clause
  */
-#define _GNU_SOURCE /* asprintf */
-#define _POSIX_C_SOURCE 200809L /* strdup */
-
+#define _GNU_SOURCE /* asprintf, strdup */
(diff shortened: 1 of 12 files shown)


michalvasko:
Generally, I have no problem with the changes but can you guarantee reasonable portability of the header `sys/cdefs.h`? What is there currently is taken directly from `man` pages so that should always work.

choppsv1:
> Generally, I have no problem with the changes but can you guarantee reasonable portability of the header `sys/cdefs.h`? What is there currently is taken directly from `man` pages so that should always work.

Using cdefs.h actually improves things, the explicit setting of the POSIX_C_SOURCE value was invalid for some architectures (having it set disabled other things that needed enabling), whereas using GNU_SOURCE and cdefs.h leads to everything being turned on (or closer to it than cherry picking).

PR #1535, opened 2021-05-03
libyang BUGFIX change LOGVAL to LOGVAL_ERRITEM

lePici:
Segfault occurs if the LOGVAR macro is called with the ly_err_item.msg parameter, which contains a conversion specification (character %). Commit should solve issues with numbers 31536, 31778, 32183 on oss-fuzz.

The warnings thrown by doxygen have also been fixed in the common.h file.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -105,10 +105,10 @@ void ly_log_location(const struct lysc_node *scnode, const struct lyd_node *dnod
 /**
  * @brief Revert the specific logger's location data by number of changes made by ::ly_log_location().
  *
- * @param[in] scnode_steps Number of items in ::log_location.scnodes to forget.
- * @param[in] dnode_steps Number of items in ::log_location.dnodes to forget.
- * @param[in] path_steps Number of path strings in ::log_location.paths to forget.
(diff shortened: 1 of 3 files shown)


src/common.h
```diff
@@ -105,10 +105,10 @@ void ly_log_location(const struct lysc_node *scnode, const struct lyd_node *dnod
 /**
  * @brief Revert the specific logger's location data by number of changes made by ::ly_log_location().
  *
- * @param[in] scnode_steps Number of items in ::log_location.scnodes to forget.
- * @param[in] dnode_steps Number of items in ::log_location.dnodes to forget.
- * @param[in] path_steps Number of path strings in ::log_location.paths to forget.
- * @param[in] in_steps Number of input handlers ::log_location.inputs to forget.
+ * @param[in] scnode_steps Number of items in log_location_s.scnodes to forget.
```
michalvasko, line 108: Please keep `::`, it simply means you require a link to an existing symbol and if it is not found, generate a warning.

michalvasko (changes requested):
Just a minor details, the solution you have used for the core problem is clean and simple.

michalvasko (approved):

PR #1537, opened 2021-05-04
xml BUGFIX regard to utf8 in lyxml_parse_value()

lePici:


src/xml.c
```diff
@@ -478,7 +478,9 @@ lyxml_parse_value(struct lyxml_ctx *xmlctx, char endchar, char **value, size_t *
             }
 
             /* continue */
-            ++offset;
+            in_aux = in;
+            LY_CHECK_GOTO(ly_getutf8(&in_aux, &n, &u), error);
```
michalvasko, line 482: `ly_getutf8` does not log so an error message should be generated here. Some more information about the problem should be included, the index of the character and it value in hex, for example.

michalvasko (changes requested):

michalvasko (approved):

PR #1539, opened 2021-05-04
Initial packaging of libyang2

oerdnj:
This adds initial packaging to the libyang2 branch.  With the "SLOPPY" format, we won't be needed a separate branch for the packaging.  It's a bit ... sloppy, but works well enough for keeping the `debian/` directory directly in the upstream branch.

The initial version of the package is `2.0.0~-1`, the `~` is less than `` (empty string), so it will just need a bump in the changelog when the first releases is ready.

//cc @mwinter-osr


michalvasko:
Please look at the comment in #1531.

mwinter-osr:
Why was debhelper requirement upgraded?
`debhelper (>= 9.20150101)` as it was before seems to work fine.

Requiring debhelper >= 10 will kill Ubuntu 16.04 (which we still care about)

oerdnj:
> Why was debhelper requirement upgraded?
> `debhelper (>= 9.20150101)` as it was before seems to work fine.
> 
> Requiring debhelper >= 10 will kill Ubuntu 16.04 (which we still care about)

debhelper >= 10 is available from xenial-backports (and Xenial is officially dead now)

mwinter-osr:
> > Why was debhelper requirement upgraded?
> > `debhelper (>= 9.20150101)` as it was before seems to work fine.
> > Requiring debhelper >= 10 will kill Ubuntu 16.04 (which we still care about)
> 
> debhelper >= 10 is available from xenial-backports (and Xenial is officially dead now)

I have the backports enabled and not found in them.
Is there a requirement for debhelper 10 as a minimum for a specific reason?

Sorry, network devices seem to be frequently outdated and we try our best to support old platforms. Our policy is to
not drop old platforms unless they become too painful to support. 
We recently just dropped CentOS 6 and Debian 8 for FRR..

oerdnj:
Could you post `apt-cache policy debhelper` output?

Xenial has been recently removed from the archives and it's only available for subscribers.

mwinter-osr:
```
ci@ci2009:~$ apt-cache policy debhelper
debhelper:
  Installed: 9.20160115ubuntu3
  Candidate: 9.20160115ubuntu3
  Version table:
     10.2.2ubuntu1~ubuntu16.04.1 100
        100 http://archive.ubuntu.com/ubuntu xenial-backports/main i386 Packages
 *** 9.20160115ubuntu3 500
        500 http://archive.ubuntu.com/ubuntu xenial/main i386 Packages
        100 /var/lib/dpkg/status
ci@ci2009:~$ sudo apt install debhelper
Reading package lists... Done
Building dependency tree       
Reading state information... Done
debhelper is already the newest version (9.20160115ubuntu3).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
ci@ci2009:~$ 
```

mwinter-osr:
Nevermind... missed to specify the backports repo on the install

But still curious if the debhelper 10 is required (I guess multiple users will run into this... not just me)

oerdnj:
`apt-get install -t xenial-backports debhelper` is what you need then.

oerdnj:
Well, there's a choice to be made. Debhelper compatibility level 10 introduces couple of changes that makes packaging easier and simpler.  It's like picking the C or C++ level, you simply have to draw a line somewhere. And while I understand your position, the debhelper 10 is available for Ubuntu 16.04 LTS, so it's a safe choice.

(Also not required at the installation time, just the build time.)

jruzicka-nic:
I've included this Debian packaging in #1531 alongside RPM in [apkg](https://pkg.labs.nic.cz/pages/apkg/) format which allows nice things discussed in the PR. It also contains some extra changes as requested (debhelper 9, updated symbols, ...) so I suggest we continue there and close this unless @oerdnj hates the apkg approach.

`apkg` is basically a generic wrapper around native packaging tools to allow upstream devs maintain packaging (for many different distros) in upstream source tree without going mad.

## Debian apkg tl;dr

```
apt install -y python3-pip
pip3 install apkg
```

`debian` files live in `distro/pkg/deb`. They are `jinja2` templated but most of them are exactly the same as in downstream with the exception of `control` and `changelog` that have version/release info replaced with `{{ version }}` and `{{ release }}` macros.

If you want to get project source with `debian` dir where you can `debuild` and such for non-trivial changes, you can:

```
apkg srcpkg
cd pkg/build/srcpkgs/debian-10/libyang2-2.0.0-1/libyang2
```

After you're happy just sync the changes from `debian` to `distro/pkg/deb`.

PR #1540, opened 2021-05-05
json BUGFIX exp fraction number

lePici:
This PR solves issues 31700 and 31424 from oss-fuzz.


michalvasko (approved):
+1 for the new test cases.

PR #1541, opened 2021-05-05
tree data REFACTOR allow lyd_value to be variably sized

choppsv1:
To gain performance from plugin types its better for the plugin to be
able to store it's value directly in lyd_value rather than having to
allocate another structure with another pointer-deref and another
cache-line load for each access.

The simple changes included here allow for lyd_value to be variable
sized, which allows for the above.

### DIFF
--- a/src/plugins_types.h
+++ b/src/plugins_types.h
@@ -476,6 +476,27 @@ typedef LY_ERR (*lyplg_type_dup_clb)(const struct ly_ctx *ctx, const struct lyd_
 typedef void (*lyplg_type_free_clb)(const struct ly_ctx *ctx, struct lyd_value *value);
 
 /**
+ * @brief Callback for obtaining the extra space required in lyd_value for
+ * values stored by ::lyplg_type_store_clb.
+ *
+ * @param[in] ctx libyang ctx to enable correct manipulation with values that are in the dictionary.
+ * @param[in] type Type of the value.
+ * @return value to add to lyd_value allocation size.
+ */
+typedef uint16_t (*lyplg_type_extra_clb)(const struct lysc_type *type);
+
+/**
+ * @brief MACRO for calculating the extra value for lyplg_type_extra_clb to return
+ *
+ * @param[in] value_size the total size of the value you need to store.
+ * @return the value to return from your callback.
+ */
+#define LYPLG_TYPE_EXTRA_CALC_EXTRA(value_size)                         \
+    ((value_size) < (sizeof(struct lyd_value) - 2 * sizeof(void *))     \
+     ? (sizeof(struct lyd_value) - 2 * sizeof(void *))                  \
+     : ((value_size) - (sizeof(struct lyd_value) - 2 * sizeof(void *))))
+
+/**
  * @brief Hold type-specific functions for various operations with the data values.
  *
  * libyang includes set of plugins for all the built-in types. They are, by default, inherited to the derived types.
@@ -493,7 +514,10 @@ struct lyplg_type {
     lyplg_type_hash_clb hash;           /**< hash callback to get the hash key of the value */
     lyplg_type_dup_clb duplicate;       /**< data duplication callback */
     lyplg_type_free_clb free;           /**< optional function to free the type-spceific way stored value */
+    lyplg_type_extra_clb extra;         /**< optional function to get extra needed value space */
 };
+
+#define LYPLG_TYPE_VALUE_EXTRA(type) ((type)->plugin && (type)->plugin->extra ? (type)->plugin->extra(type) : 0)
 
 struct lyplg_type_record {
     /* plugin identification */
@@ -905,6 +929,12 @@ LY_ERR lyplg_type_validate_union(const struct ly_ctx *ctx, const struct lysc_typ
  * Implementation of the ::lyplg_type_free_clb.
  */
 void lyplg_type_free_union(const struct ly_ctx *ctx, struct lyd_value *value);
+
+/**
+ * @brief Extra callback for the union value.
+ * Implementation of the ::lyplg_type_extra_clb.
+ */
+uint16_t lyplg_type_extra_union(const struct lysc_type *type);
 
 /** @} pluginsTypesUnion */
 
--- a/src/plugins_types/union.c
+++ b/src/plugins_types/union.c
@@ -92,7 +92,7 @@ lyplg_type_store_union(const struct ly_ctx *ctx, const struct lysc_type *type, c
     *err = NULL;
 
     /* prepare subvalue storage */
-    subvalue = calloc(1, sizeof *subvalue);
+    subvalue = calloc(1, sizeof *subvalue + lyplg_type_extra_union(type));
     if (!subvalue) {
         ret = ly_err_new(err, LY_EMEM, 0, NULL, NULL, LY_EMEM_MSG);
         goto cleanup_value;
@@ -217,7 +217,7 @@ lyplg_type_dup_union(const struct ly_ctx *ctx, const struct lyd_value *original,
 {
     LY_CHECK_RET(lydict_insert(ctx, original->_canonical, ly_strlen(original->_canonical), &dup->_canonical));
 
-    dup->subvalue = calloc(1, sizeof *dup->subvalue);
+    dup->subvalue = calloc(1, sizeof *dup->subvalue + LYPLG_TYPE_VALUE_EXTRA(original->subvalue->value.realtype));
     LY_CHECK_ERR_RET(!dup->subvalue, LOGMEM(ctx), LY_EMEM);
     LY_CHECK_RET(original->subvalue->value.realtype->plugin->duplicate(ctx, &original->subvalue->value, &dup->subvalue->value));
 
@@ -246,6 +246,22 @@ lyplg_type_free_union(const struct ly_ctx *ctx, struct lyd_value *value)
     }
 }
 
+API uint16_t
+lyplg_type_extra_union(const struct lysc_type *type)
+{
+    struct lysc_type **types = ((const struct lysc_type_union *)type)->types;
+    LY_ARRAY_COUNT_TYPE u;
+    uint16_t e, extra = 0;
+
+    LY_CHECK_RET2(type->basetype != LY_TYPE_UNION, 0);
+
+    for (u = 0; u < LY_ARRAY_COUNT(types); ++u)
+        if ((e = LYPLG_TYPE_VALUE_EXTRA(types[u])) > extra)
+            extra = e;
+
+    return extra;
+}
+
 /**
  * @brief Plugin information for union type implementation.
  *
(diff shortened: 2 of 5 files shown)


michalvasko:
I dislike this idea. If nothing else, union term nodes can change their type by simply changing their value, in which case a reallocation may be required, which is not acceptable. Other than that, having variable-size `lyd_node_term` structure seems like asking for trouble to me although I am not able to come up with an exact problem this may cause, except for the one mentioned.

Major performance improvement was already implemented in the type plugins PR that should allow for many new possibilities of working with the values efficiently. So API should be in place and after the release you can focus on utilizing it for greater efficiency and we plan to write some common use-cases and measure the performance with them. Then there will be more information about what exactly does/does not work and then there could be more improvements implemented. But I would leave it the way it is, up to the release at least.

choppsv1:
This is the type of change that would have to be made prior to freezing the API. At least the movement of the fields in the structure.
> I dislike this idea. If nothing else, union term nodes can change their type by simply changing their value, in which case a reallocation may be required, which is not acceptable.
> 
I was 50/50% on the realloc in the subvalue. If you say this isn’t OK then great, we can use the other option which is just to use the largest value_extra from the union up front and there’s no realloc then.
> Other than that, having variable-size lyd_node_term structure seems like asking for trouble to me although I am not able to come up with an exact problem this may cause, except for the one mentioned.
> 
It's real performance we’re talking about here. So I think we need better reasons.

michalvasko:
> This is the type of change that would have to be made prior to freezing the API. At least the movement of the fields in the structure.

Or leave it be and wait for the next version. This seems to me is such a fundamental change that would be fit for a new version.

> I was 50/50% on the realloc in the subvalue. If you say this isn’t OK then great, we can use the other option which is just to use the largest value_extra from the union up front and there’s no realloc then.

That is no good either for 2 reasons. Firstly, you do not know the largest value because at any point a new plugin may be dynamically added that will need larger value. Okay, you can argue that it can always define a pointer as its member and allocate additional memory of arbitrary size. But more importantly, this would cause most of the `struct lyd_node_term` to be larger that they need to be and we are optimizing for memory as well, if possible and reasonable.

Did you look at the new API functions in the initial `ipv6-address` type plugin PR? There are much more options now of providing the value for a node. The most efficient probably is using the LYB binary format. The idea is that you can provide a dynamically-allocated value in the exact same format as the specific type plugin expects it, when it will simply be taken and used without an more processing. That way you can handle the allocation whatever way you want, which is basically what this PR proposes (to allocate a single bigger block of memory instead of 2 smaller ones). Or, if you have the value in string format and can guarantee it is in its canonical form, the type plugin can treat it that way and skip the canonization, which may be costly for some types. I suppose you are most interested in IP addresses (likely both IPv6 and IPv4) and these should benefit from the mentioned possibilities most. Looking at the code, `ipv4/v6-address` and `ipv4/v6-address-no-zone` were implemented together which causes the no-zone variant to be overly complex so I will improve it by separating it. But this can also wait until after the release because the API is in place and type plugins should be possible to update afterwards as well.

> It's real performance we’re talking about here. So I think we need better reasons.

It seems to me there is simply no time to properly test this change and sensibly guarantee it will not cause problems not worth the optimization.

choppsv1:
I'm working on a cleaned up version of this. Will address comments after posting.

choppsv1:
@michalvasko please at least merge this now so we have options later w/o requiring a major version bump.

michalvasko:
Is there really any purpose to this? Allowing `lyd_value` to have arbitrary size later will definitely be a non-backwards compatible change of ABI. Also, using these "static" values will not allow dynamically-allocated values to be used directly and some form of `memcpy()` will always be required, which may be worse performance-wise at the end. I really do not like hasty solutions like this and would strongly prefer to do proper optimizations based on some actual numbers and usage-patterns.

choppsv1:
> Is there really any purpose to this? Allowing `lyd_value` to have arbitrary size later will definitely be a non-backwards compatible change of ABI. Also, using these "static" values will not allow dynamically-allocated values to be used directly and some form of `memcpy()` will always be required, which may be worse performance-wise at the end. I really do not like hasty solutions like this and would strongly prefer to do proper optimizations based on some actual numbers and usage-patterns.

Your right, I've resubmitted the proper change, implemented slightly differently. I resubmit that we should include this. I will do performance tests if required.

choppsv1:
BTW I find the comment about hasty late changes humorous, give this is a simple and small code change which passes testing, vs #1536 which you are proposing which is *GIGANTIC* and breaks existing code.

michalvasko:
> Your right, I've resubmitted the proper change, implemented slightly differently. I resubmit that we should include this. I will do performance tests if required.

And I stand by what I said, I am strongly against this change. If for no other reason than because if you have a term node and change its value, it could be reallocated and your node pointer become invalid, that just does not seem like a safe API. Also, what exactly do you want to test the performance of? You would need to adjust the type plugins first to use this new API.

> BTW I find the comment about hasty late changes a bit humorous, given this is a simple and small code change which passes testing, vs #1536 which you are proposing which is GIGANTIC and breaks existing code.

That is all based on your comment about making the canonical value non-mandatory. I did not want to leave the solution half-baked for the release and implement it all properly for all the types. And it did not really make sense without adding the LYB binary value format. But since it was planned for later anyway, I have implemented it all. We were perfectly fine keeping the type plugins the way they were before and implementing these improvements later.

Nevertheless, if really needed it is possible to leave the PR out of the release and merge it later. Only the few minor API changes can be merged and since the plugins are done, no more NBC changes will be needed.

Finally, there is a lot of code changed but the changes do not really seem that large to me and the LYB values are not used at all in neither your nor our tests. But when trying to use the values more efficiently I have reproduced several errors with non-0 terminated string values that were hidden in the previous type plugin implementations. It is possible that is what is causing the test failures and the problem may actually be on your end.

choppsv1:
> > Your right, I've resubmitted the proper change, implemented slightly differently. I resubmit that we should include this. I will do performance tests if required.
> 
> And I stand by what I said, I am strongly against this change. If for no other reason than because if you have a term node and change its value, it could be reallocated and your node pointer become invalid, that just does not seem like a safe API. Also, what exactly do you want to test the performance of? You would need to adjust the type plugins first to use this new API.

Have you looked at the diff since I resubmitted? This is not true, there is no reallocation.

rkrejci:
I didn't see the original version, but I actually like the current one. Maybe the same approach could be used also for the built-in types (internal plugins) to unify the approach to the stored values. So the rest of the `lyd_value`'s union members (so maybe the whole union?) could be removed and instead there would be just macros casting the data to simplify the use/access.

To fix the failing test, please just fix formatting by running `make format` (and then commit the changes).

choppsv1:
> I didn't see the original version, but I actually like the current one. Maybe the same approach could be used also for the built-in types (internal plugins) to unify the approach to the stored values. So the rest of the `lyd_value`'s union members (so maybe the whole union?) could be removed and instead there would be just macros casting the data to simplify the use/access.
> 
> To fix the failing test, please just fix formatting by running `make format` (and then commit the changes).

I like this idea it also allows for smaller allocations for smaller types (although alignment requirements on 64 bit architectures may hurt in the end).

I missed one embedded use case in the predicates, I need to fix that still.

choppsv1:
I have a different idea that maybe @michalvasko would be happier with, and would require very little code change.

What if we add uint8_t data[24] to the union?

This would increase the overall size by 16 bytes (a couple pointers). It would cover ipv6-address (i.e., in6_addr with a zone pointer), it would also cover `ip-prefix` (the longer `ipv6-prefix`) as well as `ipv6-address-and-prefix` both of which do not include the zone prefix. These are the largest fixed-sized things I can think of where you might have millions of them and want to do fast comparisons. As a proportion of the size of the node it's only adding the equivalent of 2 more pointers (16 bytes). Anything larger than this probably does deserve it's own separate allocation.

michalvasko:
So we have talked it over and we will accept this change. Specifically, there should be a new callback that returns the required value size of the particular type. Usually, simply a number will be returned but `union` will need to return the largest size from all its subtypes and `leafref` the size of its target type.

And since the `lyd_value` will be of variable size (which I am still not sure is okay but hopefully is), making `struct lyd_node_term` packed could save some memory as well. However, it would likely make sense only if we get rid of all the members and actually replace the union in `struct lyd_value` by `uint8_t value[1]` (`0` is not allowed, strictly speaking). Then, to access the values, define macros that will cast it properly for each type. Not sure it is worth the trouble, though. But some casting (for values larger than 64b) will always be required (built-in union even) so it may not be a bad idea to unify it, not sure.

choppsv1:
> So we have talked it over and we will accept this change. Specifically, there should be a new callback that returns the required value size of the particular type. Usually, simply a number will be returned but `union` will need to return the largest size from all its subtypes and `leafref` the size of its target type.
> 
> And since the `lyd_value` will be of variable size (which I am still not sure is okay but hopefully is), making `struct lyd_node_term` packed could save some memory as well. However, it would likely make sense only if we get rid of all the members and actually replace the union in `struct lyd_value` by `uint8_t value[1]` (`0` is not allowed, strictly speaking). Then, to access the values, define macros that will cast it properly for each type. Not sure it is worth the trouble, though. But some casting (for values larger than 64b) will always be required (built-in union even) so it may not be a bad idea to unify it, not sure.

Well what do you think about the data[24]? I'm actually warming up to this idea.. The changes in the predicate code are a bit daunting this late in the game, and given how often the user is going to use paths with predicates to do things, I'm not sure we end up winning since these all need to alloc lyd_values now. :(

michalvasko:
I would not worry about the predicates that much, not sure they are used that often. In any case, from all our past experiences the most severe performance problems are caused by large data sets, parsing a few predicates is not the case.

choppsv1:
> I would not worry about the predicates that much, not sure they are used that often. In any case, from all our past experiences the most severe performance problems are caused by large data sets, parsing a few predicates is not the case.

FRR is using lyd_new_path all over to create things. :)

choppsv1:
> > I would not worry about the predicates that much, not sure they are used that often. In any case, from all our past experiences the most severe performance problems are caused by large data sets, parsing a few predicates is not the case.
> 
> FRR is using lyd_new_path all over to create things. :)

Also, I got a bit more scared when I found more embeded uses on the stack (e.g., lyd_value_compare). I know I'm reversing myself here, but I'd prefer to just add "uint8_t data[24]" to lyd_value and then use that in the various IP types.

michalvasko:
> FRR is using lyd_new_path all over to create things. :)

If you care about performance that much, you should not be using it at all because the other API (`lyd_new_inner()`, `lyd_new_term()`, ...) is always faster. But it is true that after all the optimizations the difference should only be almost a constant.

> Also, I got a bit more scared when I found more embeded uses on the stack (e.g., lyd_value_compare). I know I'm reversing myself here, but I'd prefer to just add "uint8_t data[24]" to lyd_value and then use that in the various IP types.

You think allocating the value dynamically there would make such a big difference?

In any case, I suppose you do not care that much since you are mostly using IP addresses but in general that would make a lot of `struct lyd_term` instances 2 bytes larger than they need to be. For large data (lists with keys), it can make a noticeable difference although after doing some math I am not sure how relevant it is. So I am torn which way to go, if you think having the value larger but constant size is faster, then I am okay with it.

choppsv1:
#1552 is a replacement for this, using a fixed space inside lyd_value.

PR #1547, opened 2021-05-07
data tree FEATURE add sorting compare callback

choppsv1:
Add a traditional cmp compare (i.e., sorting) to plugin interface. Eventually libyang may wish to sort (e.g., leaf-lists or keyless lists); however, more importantly, users may wish to know *how* a value is not equal (is it less or more)..

### DIFF
--- a/src/plugins_types.c
+++ b/src/plugins_types.c
@@ -284,6 +284,21 @@ lyplg_type_compare_simple(const struct lyd_value *val1, const struct lyd_value *
     return LY_ENOT;
 }
 
+API int
+lyplg_type_sort_simple(const struct lyd_value *val1, const struct lyd_value *val2)
+{
+    int result;
+
+    if (lyplg_type_initial_sort(&val1, &val2, &result) == LY_SUCCESS) {
+        return result;
+    }
+
+    if (val1->_canonical == val2->_canonical) {
+        return 0;
+    }
+    return strcmp(val1->_canonical, val2->_canonical);
+}
+
 API const void *
 lyplg_type_print_simple(const struct ly_ctx *UNUSED(ctx), const struct lyd_value *value, LY_VALUE_FORMAT UNUSED(format),
         void *UNUSED(prefix_data), ly_bool *dynamic, size_t *value_len)
@@ -747,6 +762,54 @@ API void
 lyplg_type_lypath_free(const struct ly_ctx *ctx, struct ly_path *path)
 {
     ly_path_free(ctx, path);
+}
+
+static inline int
+is_int_type(const struct lysc_type *realtype)
+{
+    switch (realtype->basetype) {
+    case LY_TYPE_UINT8:
+    case LY_TYPE_UINT16:
+    case LY_TYPE_UINT32:
+    case LY_TYPE_UINT64:
+    case LY_TYPE_INT8:
+    case LY_TYPE_INT16:
+    case LY_TYPE_INT32:
+    case LY_TYPE_INT64:
+        return 1;
+    default:
+        return 0;
+    }
+}
+
+API LY_ERR
+lyplg_type_initial_sort(const struct lyd_value **val1, const struct lyd_value **val2, int *result)
+{
+    const struct lysc_type *rt1 = (*val1)->realtype, *rt2 = (*val2)->realtype;
+
+    if (rt1->basetype == LY_TYPE_UNION) {
+        *val1 = &(*val1)->subvalue->value;
+        rt1 = (*val1)->realtype;
+    }
+    if (rt2->basetype == LY_TYPE_UNION) {
+        *val2 = &(*val2)->subvalue->value;
+        rt2 = (*val2)->realtype;
+    }
+
+    if ((rt1 != rt2) && !(is_int_type(rt1) && is_int_type(rt2))) {
+        /* sort by type */
+        if (rt1->basetype == rt2->basetype) {
+            *result = strcmp(rt1->plugin->id, rt2->plugin->id);
+        } else if (rt1->basetype < rt2->basetype) {
+            *result = -1;
+        } else {
+            *result = 1;
+        }
+        return LY_SUCCESS;
+    }
+
+    /* caller should finish the compare */
+    return LY_EINCOMPLETE;
 }
 
 API LY_ERR
--- a/src/plugins_types.h
+++ b/src/plugins_types.h
@@ -391,6 +391,21 @@ LY_ERR lyplg_type_lypath_new(const struct ly_ctx *ctx, const char *value, size_t
 void lyplg_type_lypath_free(const struct ly_ctx *ctx, struct ly_path *path);
 
 /**
+ * @brief Check the inputs to see if the sort compare can be handled up-front
+ *
+ * This should be the first thing called in a callback's sort compare function. If it returns LY_SUCCESS then
+ * the callee should return result to the caller. Typically this function is dealing with dissimilar types
+ * between 'val1' and 'val2' values.
+ *
+ * @param[in/out] val1 First value to compare (may update).
+ * @param[in/out] val2 Second value to compare (may update)
+ * @param[out] result Value to return from callback if LY_SUCCESS is returned.
+ * @return LY_SUCCESS - comparison is done, return result from the callback.
+ * @return LY_EINCOMPLETE - continue doing comparison as types are directly comparable.
+ */
+LY_ERR lyplg_type_initial_sort(const struct lyd_value **val1, const struct lyd_value **val2, int *result);
+
+/**
  * @defgroup plugintypestoreopts Plugins: Type store callback options.
  *
  * Options applicable to ::lyplg_type_store_clb().
@@ -473,6 +488,20 @@ typedef LY_ERR (*lyplg_type_validate_clb)(const struct ly_ctx *ctx, const struct
 typedef LY_ERR (*lyplg_type_compare_clb)(const struct lyd_value *val1, const struct lyd_value *val2);
 
 /**
+ * @brief Callback for comparing 2 values for sorting
+ *
+ * In case the value types (::lyd_value.realtype) are different, ::LY_ENOT must always be returned.
+ * It can be assumed that the same context (dictionary) was used for storing both values.
+ *
+ * Input values are expected to be the same or comparable types (e.g., both integers) and valid.
+ *
+ * @param[in] val1 First value to compare.
+ * @param[in] val2 Second value to compare.
+ * @return -1 if val1 < val2, 0 if val1 == val2, 1 if val1 > val2.
+ */
+typedef int (*lyplg_type_sort_clb)(const struct lyd_value *val1, const struct lyd_value *val2);
+
+/**
  * @brief Callback for getting the value of the data stored in @p value.
  *
  * Canonical value (@p format of ::LY_VALUE_CANON) must always be a zero-terminated const string stored in
@@ -530,6 +559,7 @@ struct lyplg_type {
     lyplg_type_store_clb store;         /**< store and canonize the value in the type-specific way */
     lyplg_type_validate_clb validate;   /**< optional, validate the value in the type-specific way in data */
     lyplg_type_compare_clb compare;     /**< comparison callback to compare 2 values of the same type */
+    lyplg_type_sort_clb sort;           /**< sorting callback to compare 2 values of the same type */
     lyplg_type_print_clb print;         /**< printer callback to get string representing the value */
     lyplg_type_dup_clb duplicate;       /**< data duplication callback */
     lyplg_type_free_clb free;           /**< optional function to free the type-spceific way stored value */
@@ -562,6 +592,11 @@ struct lyplg_type_record {
  * @brief Implementation of ::lyplg_type_compare_clb for a generic simple type.
  */
 LY_ERR lyplg_type_compare_simple(const struct lyd_value *val1, const struct lyd_value *val2);
+
+/**
+ * @brief Implementation of ::lyplg_type_sort_clb for a generic simple type.
+ */
+int lyplg_type_sort_simple(const struct lyd_value *val1, const struct lyd_value *val2);
 
 /**
  * @brief Implementation of ::lyplg_type_print_clb for a generic simple type.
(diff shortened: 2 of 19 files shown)


choppsv1:
So I did not implement all the sort methods that are possible. Before finishing this off (can be done in an hour or so) I just wanted to get a read from the maintainers on whether they are happy with this addition.

michalvasko:
> So I did not implement all the sort methods that are possible. Before finishing this off (can be done in an hour or so) I just wanted to get a read from the maintainers on whether they are happy with this addition.

Right, in general yes but what about simply improving the compare callback to return standard `int` values instead of `LY_ERR` enum? In theory using 2 callbacks is more efficient but in practice it should make no or very little difference.

choppsv1:
> > So I did not implement all the sort methods that are possible. Before finishing this off (can be done in an hour or so) I just wanted to get a read from the maintainers on whether they are happy with this addition.
> 
> Right, in general yes but what about simply improving the compare callback to return standard `int` values instead of `LY_ERR` enum? In theory using 2 callbacks is more efficient but in practice it should make no or very little difference.

I started off this way, but then thought maybe 2 were better. I'm wondering if there are some types where there really is no "less than" comparison that makes sense, in which case the clb could be left NULL to inidicate this. I'm on the fence though, I think one could always just default to strcmp on the canonical.

What do you think? If you think that default of strcmp on canonical makes sense I have no problem just porting these over to compare and changing the RV on compare to int from LY_ERR and going through the source to fixup uses.

If canonical is not set then i thought I would call print to get it; however, that requires a context arg. This was the other thing that stopped me. Perhaps we could indicate to plugin authors that if they do not set _canonical on store they need to support compare rather than rely on a simple default?

michalvasko:
Yes, we have considered this but since YANG does not define any ordering for its types, they are all implementation-specific. And one can always define some ordering, so it would be best if you simply improved the current callback for the existing types and fix its description, thanks.

For the ordering of some complex types, I would order `bits` based on the position of the first set bit and `instance-identifier` probably just lexicographically on the specific members.

I have also realized this depends quite a lot on the other PR #1536 so you may want to build on that. I will try to look at the tests more today and hopefully find out what the problem is.

choppsv1:
> Yes, we have considered this but since YANG does not define any ordering for its types, they are all implementation-specific. And one can always define some ordering, so it would be best if you simply improved the current callback for the existing types and fix its description, thanks.

I sat down to do this code, and it's exposed an issue that I think warrant's 2 callbacks. Consider the current `lyplg_type_compare_simple` the first thing it does is compare realtype's of the 2 values. This makes sense e.g., when perhaps walking a list to see if you've found a particular value. If the function has to return an ordering decision then we need to do a lot more than just return "false" in this case.

I do think that we need to support ordering based on different realtypes (this can be factored into a common front end function), but it doesn't make sense to run this code (probably comparing a types basetype then the plugin "id"s lexicographically) when all the user wants is to determine if they found an object or not.

I'm going to keep the sort function separate and flush out the rest of the basic implementation. The only really important thing here is to get the callback in before API freeze, the use of it doesn't have to be perfected right now. :)

A slight naming thing, not important, but we could also rename "compare" to "equal" and "sort" to "compare", but again not that important.

choppsv1:
Need to wait to review this until it's rebased on the underlying binary rep changes after they merge to libyang2.

michalvasko:
If the callback will not be used for ordering yet anyway, why bother adding it? That is not a NBC change, we just need to agree on whether there will be one or 2 callbacks.

> I do think that we need to support ordering based on different realtypes (this can be factored into a common front end function), but it doesn't make sense to run this code (probably comparing a types basetype then the plugin "id"s lexicographically) when all the user wants is to determine if they found an object or not.

Once again, it would be best to have the actual numbers (in most cases it is a folly to do optimizations without them) to decide this because even for strings, if `strcmp()` is used, it may take almost the same amount of time in case there is a difference at the beginning of the string. And for types with actual binary values, the comparison should generally be even faster.

choppsv1:
Michal Vasko ***@***.***> writes:

>    If the callback will not be used for ordering yet anyway, why bother
>    adding it? That is not a NBC change, we just need to agree on whether
>    there will be one or 2 callbacks.

I would think that adding a plugin callback would require a version shift which then enters into version checks everywhere. This is a pretty obvious API so why not add it now?

>
>      I do think that we need to support ordering based on different
>      realtypes (this can be factored into a common front end function),
>      but it doesn't make sense to run this code (probably comparing a
>      types basetype then the plugin "id"s lexicographically) when all the
>      user wants is to determine if they found an object or not.
>
>    Once again, it would be best to have the actual numbers (in most cases
>    it is a folly to do optimizations without them) to decide this because
>    even for strings, if strcmp() is used, it may take almost the same

Doing an inline != comparison on a single uint32 value is going to be faster than doing that same compare and then a strcmp function call. We don't have to profile this to know it.

michalvasko:
> Doing an inline != comparison on a single uint32 value is going to be faster than doing that same compare and then a strcmp function call. We don't have to profile this to know it.

The question is how much faster and you could skip the initial comparison when using `strcmp` afterwards. But okay, I think having 2 callbacks should be fine and would allow the `compare` callback (I would rename the current callback to `equal` if we go this way) to be optional, if it made sense.

choppsv1:
Michal Vasko ***@***.***> writes:

>>     Doing an inline != comparison on a single uint32 value is going to
>>     be faster than doing that same compare and then a strcmp function
>>     call. We don't have to profile this to know it.
>
>    The question is how much faster and you could skip the initial
>    comparison when using strcmp afterwards. But okay, I think having 2

Also, consider the case where you are comparing hashed string values (realtypes are the same). You can determine inequality with 2 integer comparisons. If you have to return a sorting result you have to do an actual strcmp. This is going slower.

>    callbacks should be fine and would allow the compare callback (I would
>    rename the current callback to equal if we go this way) to be optional,
>    if it made sense.

I thought about this. The mass rename is a bit annoying I wasn't sold on it in the end. I would prefer this be done in a separate commit if so.

FWIW is "is_equal" a more appropriate replacement (the callbacks are named as verbs, and "equal" is an adjective so it's not a good replacement name).

michalvasko:
> Also, consider the case where you are comparing hashed string values (realtypes are the same). You can determine inequality with 2 integer comparisons. If you have to return a sorting result you have to do an actual strcmp. This is going slower.

Okay, fine.

> I thought about this. The mass rename is a bit annoying I wasn't sold on it in the end. I would prefer this be done in a separate PR if so.
> 
> FWIW is "is_equal" a more appropriate replacement (the callbacks are named as verbs, and "equal" is an adjective so it's not a good replacement name).

Separate commit makes sense, no problem. It is just that `equal` is shorter and quite self-explaining, I think. But `is_equal` will do, too.

michalvasko:
> Great. I'm going to write a few unit-tests, b/c I want to make sure I got the decimal64 one right. :)

Yes, more tests will be added after the release, I suppose, but they never hurt. Just note that I am finalizing a fairly small PR that should probably be merged before any of yours because I removing the hash callback and using the print one with LYB format instead. You changed it this way in some PR and that made me realize it is useless after all, the LYB value can always be used and should return the exact same data.

choppsv1:
> > Great. I'm going to write a few unit-tests, b/c I want to make sure I got the decimal64 one right. :)
> 
> Yes, more tests will be added after the release, I suppose, but they never hurt. Just note that I am finalizing a fairly small PR that should probably be merged before any of yours because I removing the hash callback and using the print one with LYB format instead. You changed it this way in some PR and that made me realize it is useless after all, the LYB value can always be used and should return the exact same data.

I considered this; however, I had a different idea on how to fix it.

1) I think you should leave the hash callback, it's very possible that a plugin author can make a more efficient hash function than what print binary format would have to do.

2) I think we should use a common function defined in plugins_types.h to get the hash, this common function checks for the hash callback, and if it is NULL it calls print requesting binary format.

3) I think that (2) is a pattern that can/should be established for all plugin callbacks. Instead of calling `something->realtype->plugin->FOO()` there should be a `lyplg_FOO(realtype, <rest-of-args>`) for each callback `FOO`. 

The reason you do 3 is so plugin authors only need to implement a minimum set of functions, specialized functions can be added later (e.g., a faster hash!, or a faster is_equal, ...) The front-end function can make choices about what to call based on what is implemented. An example of this exists in python e.g., where one only needs to implement a couple ordering operators (is_equal and less_than) to get an ordered type, but one also can implement the full set of operators (is equal, not equal, less than, greater than ... etc, so that things can be made more efficient if that makes sense.

choppsv1:
> 1. I think you should leave the hash callback, it's very possible that a plugin author can make a more efficient hash function than what print binary format would have to do.

I keep forgetting that the hash callback returns a string and length, I really wish it wouldn't do this. It would be much better if it returned a uint32_t so that it can actually perform a hash that makes sense for it's type. Sometimes this is *not* just some bit twiddling on a string.

michalvasko:
> I keep forgetting that the hash callback returns a string and length, I really wish it wouldn't do this. It would be much better if it returned a uint32_t so that it can actually perform a hash that makes sense for it's type. Sometimes this is not just some bit twiddling on a string.

What do you mean, it returns `const void *` (just like `print()`) which is meant to be a pointer to some binary data, only a string if there are no binary data. That is exactly what the LYB value is supposed to be so I still do not see the point of keeping it. If you can provide a more efficient hash key than LYB value, why did you not define the LYB value that way in the first place? It is not supposed to be nice-to-use, it is supposed to be small and easy to get from the stored value. The stored value should be both efficient and nice-to-work-with so may not correspond to the LYB value but be similar.

michalvasko:
Maybe there is some misunderstanding but *libyang* does not allow to customize the hash function, it only allows (and requires) the types to define their own **hash key** (the data that the hash is calculated from).

choppsv1:
Michal Vasko ***@***.***> writes:

>    Maybe there is some misunderstanding but libyang does not allow to
>    customize the hash function, it only allows (and requires) the types to
>    define their own hash key (the data that the hash is calculated from).


There's no reason for this limitation if you simply change the hash callback to return a uint32_t and handle that.

An example of what I mean by "front-ends" can be found here:

   https://github.com/CESNET/libyang/pull/1554/files

michalvasko:
Like I said, the value hash key is used as only one of the components of the full hash. Are you saying we should be further hashing a hash? That does not sound like a good idea to me.

choppsv1:
> Like I said, the value hash key is used as only one of the components of the full hash. Are you saying we should be further hashing a hash? That does not sound like a good idea to me.

I'm saying a plugin type knows how to hash itself better than your generic function, why are you forcing it? Just have the callback return a uint32_t and work that into your "hash chain".. this doesn't seem that hard.

michalvasko:
> I'm saying a plugin type knows how to hash itself better than your generic function, why are you forcing it? Just have the callback return a uint32_t and work that into your "hash chain".. this doesn't seem that hard.

I am simply asking for an explanation, not saying it is hard. But yes, you are saying what should happen is
```
hash data A -> hash data B -> hash (specific hash of value) -> final hash
```
instead of what happens now
```
hash data A -> hash data B -> hash (specific value) -> final hash
```
The specific hash (no matter how fast) seems completely redundant.

choppsv1:
Michal Vasko ***@***.***> writes:

>      I'm saying a plugin type knows how to hash itself better than your
>      generic function, why are you forcing it? Just have the callback
>      return a uint32_t and work that into your "hash chain".. this
>      doesn't seem that hard.
>
>    I am simply asking for an explanation, not saying it is hard. But yes,
>    you are saying what should happen is
> hash data A -> hash data B -> hash (specific hash of value) -> final hash

Well, I wouldn't have chosen to run the u32 result through the exact same hash function that normally processes binary data, instead:

hash data A -> hash data B ^ u32(specific hash of value) -> final hash.

>    instead of what happens now
> hash data A -> hash data B -> hash (specific value) -> final hash
>
>    The specific hash (no matter how fast) seems completely redundant.

Well I can't do anymore convincing that specific hashes are better than generic ones (and not redundant?!) than to point you at a bunch more research papers.

I'll give up.

michalvasko:
> Well I can't do anymore convincing that specific hashes are better than generic ones (and not redundant?!) than to point you at a bunch more research papers.

I am simply asking you to be exact, you are not. Then you cannot be surprised that I do not understand what you mean. Obviously, I am not saying generic hashes are better than specific ones. Simply changing the callback would mean a specific hash would be hashed again by a generic hash function, in which case the first hash would be redundant.

> hash data A -> hash data B ^ u32(specific hash of value) -> final hash.

So you are proposing to perform `XOR` on the partial hash? Okay, that can work, I suppose. I know nothing about how to combine hashes.

Still, I am not sure all of this is worth it, it does add another callback (albeit optional) and increase the overall complexity. These specific type hashes are used mostly when creating (parsing) data and the actual performance impact may easily get lost and be negligible. Once again, some exact measurements would be best. The papers you posted analyze too specific use-cases, the hashes we are discussing are more generic and, for example, any lookup advantages will get lost when the hash is combined into the final hash, I think.

michalvasko:
Returning to the original topic, I think we agreed that 2 callbacks will be best. So I would rename the `compare` callback to `is_equal` and that is all. Adding a new callback would not break API and we do not plan to implement data node ordering based on it anytime soon. But, I suppose, defining the prototype, adding it to the plugin structure, and setting it to `NULL` in all the plugins will work, too.

Will you do these changes or should I? It should not take too long.

choppsv1:
> @choppsv1 So if I just rename the `compare` callback, will that be fine? Meaning is it the last suggestion and we will be free to make the release? After you have tested it, of course.

Just add the new callback and do the rename, and cut a release. The callback code doesn't matter for the API freeze I guess. You don't need to explicitly set anything to NULL as that's the initialized value of the data.

choppsv1:
Or just add the callback named "sort" as I did and be done (very small change). I don't care about the rename.

michalvasko:
I have pushed the changes, did not rename the current callback as it was likely useless like you said. So, are you fine with the current API and can it be released?

choppsv1:
> I have pushed the changes, did not rename the current callback as it was likely useless like you said. So, are you fine with the current API and can it be released?

Yes.

PR #1552, opened 2021-05-08
Chopps/ly2/pr1536 fixes

choppsv1:


michalvasko:
In principle, the changes are fine, I think, except making it more complex to work with the value. But to properly integrate your commit, it will take me at least the whole Monday.

choppsv1:
> In principle, the changes are fine, I think, except making it more complex to work with the value. But to properly integrate your commit, it will take me at least the whole Monday.

Is there a formatting issue? You can just git cherry pick the commit on top of yours, or you could merge this PR which includes all your commits (with their attributions etc..).

choppsv1:
@michalvasko if it means we get all these commits in faster then why don't you grab the earlier calloc fix: 07f034f07fee5d8f13070488bc38d35c44bafad1 today?

michalvasko:
> if it means we get all these commits in faster then why don't you grab the earlier calloc fix: 07f034f today?

Okay, that should allow to merge my original PR so will probably be best.

> Is there a formatting issue? You can just git cherry pick the commit on top of yours, or you could merge this PR which includes all your commits (with their attributions etc..). (i.e., either case it's a single command, which shouldn't require a day to "integrate").

It is partly a formatting issue. If saving some memory is to be properly supported, the fixed memory size needs to be possible to be adjusted by a `cmake` variable. Then, it must all be properly supported. For example, if you set the size to 4 on a 32b platform, you will need to dynamically allocate even a `uint64` value. And then be able to access it since you could not do so directly.

But perhaps this is simply not worth supporting, we have taken this approach with the hash table, which is now mandatory because even though it requires a lot of memory, the performance gain is huge and making it optional would add redundant complexity of the code. But to learn what exactly is best with these values, I am repeating again, we would need some test-cases and measure the performance changes with them. That was the plan, to design them in the next months. But you are rushing to do all these changes before this release, which is now dragging for way too long and we are several weeks behind in our schedule (seems to me it is the same with you). Why not do all these optimizations later (with other changes included based on the usage of the released version) with having all the relevant numbers and, if need be, releasing a stable and robust *libyang* version 3?

choppsv1:
Michal Vasko ***@***.***> writes:

>> Is there a formatting issue? You can just git cherry pick the commit
>> on top of yours, or you could merge this PR which includes all your
>> commits (with their attributions etc..). (i.e., either case it's a
>> single command, which shouldn't require a day to "integrate").
>
> It is partly a formatting issue. If saving some memory is to be
> properly supported, the fixed memory size needs to be possible to be
> adjusted by a cmake variable. Then, it must all be properly supported.
> For example, if you set the size to 4 on a 32b platform, you will need
> to dynamically allocate even a uint64 value. And then be able to access
> it since you could not do so directly.

No, this is making this overly complex for no reason at all, you simply do not allow it to be smaller than 8.

>    But perhaps this is simply not worth supporting, we have taken this
>    approach with the hash table, which is now mandatory because even
>    though it requires a lot of memory, the performance gain is huge and
>    making it optional would add redundant complexity of the code. But to
>    learn what exactly is best with these values, I am repeating again, we
>    would need some test-cases and measure the performance changes with
>    them. That was the plan, to design them in the next months. But you are

I saw a ~9% improvement in speed with IPv6 addresses. I am out of time to keep doing all this proving.

It's disappointing that you don't wish to take this change. I put a bunch of extra work in so that it was completely selectable at compile time, in case people didn't want to sacrifice the small space for the speed gain.

michalvasko:
> It's disappointing that you don't wish to take this change. I put a bunch of extra work in so that it was completely selectable at compile time, in case people didn't want to sacrifice the small space for the speed gain.

That is the problem, it is not completely selectable. Even if we decide 8B is the smallest allowed size, several types (`union`, `binary`, ...) could be stored in a dynamic or static memory, based on the configured size. So there could not longer be a single way to access them as members of the union in `lyd_value`. That entails adding some functions or macros for accessing these values using API. But then it may also make sense to remove the direct access altogether and design a way to access all the values, always specific for the type.

All of these are rather large changes and that is why I am not sure we are able to come up with a "final" API on such short notice, all of these decisions have too much impact to be able to do this, in my opinion.

choppsv1:
Michal Vasko ***@***.***> writes:

>      It's disappointing that you don't wish to take this change. I put a
>      bunch of extra work in so that it was completely selectable at
>      compile time, in case people didn't want to sacrifice the small
>      space for the speed gain.
>
>    That is the problem, it is not completely selectable. Even if we decide
>    8B is the smallest allowed size, several types (union, binary, ...)
>    could be stored in a dynamic or static memory, based on the configured
>    size.

Your adding features to my idea, and then saying my idea is too complex.

I'm looking for a simple optimization, it's not complex and it works well.

michalvasko:
> You're adding features to my idea, and then saying my idea is too complex.

Then I may be confused as to what is your idea, is changing the static memory size supported or not? It seemed to me it was but simply changing the single `define` would not do the trick, that is all I am saying.

choppsv1:
In any case, please let's get the binary fixes in so that I can rebase my PRs so the diffs are reviewable.

choppsv1:
Michal Vasko ***@***.***> writes:

>      You're adding features to my idea, and then saying my idea is too
>      complex.
>
>    Then I may be confused as to what is your idea, is changing the static
>    memory size supported or not? It seemed to me it was but simply
>    changing the single define would not do the trick, that is all I am
>    saying.


Having and changing the define changes the size of lyd_value. That is all.

Whether that is exported to cmake, or whether things take advantage of that or not is unimportant to the API change. I modified the obvious candidates to take advantage of this adding 3 macros to make the conversion uniform and simple.

The key thing is to get the static space allocation into the API before freeze. Not to optimize everywhere possible. It's also important to pick a good default, and 24 is it IMO. It fits ipv[46]-addresses or prefixes.

michalvasko:
> The key thing is to get the static space allocation into the API before freeze. Not to optimize everywhere possible. It's also important to pick a good default, and 24 is it IMO. It fits ipv[46]-addresses or prefixes.

Not sure how important, in general, handling IPv6 addresses is, but on standard platforms it should be fine (the higher memory requirements), so okay.

What I am worried about is the `lyd_value` union. You have removed values larger than 8B as that would be the smallest allowed size, okay. So all the other values now need to be accessed using the macro you added. That seems functional but the way of accessing the values is not uniform.

I am still not sure why you abandoned the variable-size values at the end because the other uses would work with it, too. In any case, then the whole union would probably be removed as every value would be stored directly and accessing them would be uniform, in addition to always taking only the required memory.

choppsv1:
Michal Vasko ***@***.***> writes:
>    I am still not sure why you abandoned the variable-size values at the
>    end because the other uses would work with it, too. In any case, then
>    the whole union would probably be removed as every value would be
>    stored directly and accessing them would be uniform, in addition to
>    always taking only the required memory.

Predicates required a bunch of work b/c they use an array of lyd_value (not lyd_value *), and then I found some "struct lyd_value val = {}" sitting on the stack in various functions and decided this was more than I wanted to take on so late in the game.

michalvasko:
Okay, so the minimal allowed value size will be 8B and these types will be accessible in the union. All other types will have their own structure that should be accessed by the provided macro because they could be allocated in a dynamic or static memory. That could work, I suppose. Now I have all the information I need to finish these changes by some cosmetic/doc improvements.

choppsv1:
> Okay, so the minimal allowed value size will be 8B and these types will be accessible in the union. All other types will have their own structure that should be accessed by the provided macro because they could be allocated in a dynamic or static memory. That could work, I suppose. Now I have all the information I need to finish these changes by some cosmetic/doc improvements.
> 
> By the way, there may not be any actual conflicts because github thought so for the other PR as well but I merged without any conflicts on the command line.

I had another branch locally that didn't squash the intermediate calloc commit that you used, so it rebased cleanly on libyang2 post your merge. I then hard reset this PR branch to that and force-pushed. So this PR is now good for review.

PR #1554, opened 2021-05-10
plugins types REFACTOR front-ends allow for more functionality

choppsv1:
Many advantages to using front-ends for accessing plugin callbacks,
including allowing for optional callback functionality, future expansion of
plugin types, default implementations, etc..

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -1435,7 +1435,7 @@ lyd_diff_merge_create(struct lyd_node *diff_match, enum lyd_diff_op cur_op, cons
                 sleaf = (struct lysc_node_leaf *)diff_match->schema;
             }
 
-            if (sleaf && sleaf->dflt && !sleaf->dflt->realtype->plugin->compare(sleaf->dflt,
+            if (sleaf && sleaf->dflt && !lyplg_type_compare(sleaf->dflt->realtype, sleaf->dflt,
                     &((struct lyd_node_term *)src_diff)->value)) {
                 /* we deleted it, so a default value was in-use, and it matches the created value -> operation NONE */
                 LY_CHECK_RET(lyd_diff_change_op(diff_match, LYD_DIFF_OP_NONE));
--- a/src/path.c
+++ b/src/path.c
@@ -1067,7 +1067,7 @@ ly_path_dup(const struct ly_ctx *ctx, const struct ly_path *path, struct ly_path
                 case LY_PATH_PREDTYPE_LEAFLIST:
                     /* key-predicate or leaf-list-predicate */
                     (*dup)[u].predicates[v].key = pred->key;
-                    pred->value.realtype->plugin->duplicate(ctx, &pred->value, &(*dup)[u].predicates[v].value);
+                    lyplg_type_duplicate(pred->value.realtype, ctx, &pred->value, &(*dup)[u].predicates[v].value);
                     ++((struct lysc_type *)pred->value.realtype)->refcount;
                     break;
                 case LY_PATH_PREDTYPE_NONE:
@@ -1098,7 +1098,7 @@ ly_path_predicates_free(const struct ly_ctx *ctx, enum ly_path_pred_type pred_ty
         case LY_PATH_PREDTYPE_LIST:
         case LY_PATH_PREDTYPE_LEAFLIST:
             if (predicates[u].value.realtype) {
-                predicates[u].value.realtype->plugin->free(ctx, &predicates[u].value);
+                lyplg_type_free(predicates[u].value.realtype, ctx, &predicates[u].value);
                 lysc_type_free((struct ly_ctx *)ctx, (struct lysc_type *)predicates[u].value.realtype);
             }
             break;
--- a/src/plugins_types.c
+++ b/src/plugins_types.c
@@ -791,7 +791,7 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
             &set, 0);
     if (ret) {
         ret = LY_ENOTFOUND;
-        val_str = lref->plugin->print(lref->cur_mod->ctx, value, LY_VALUE_CANON, NULL, NULL, NULL);
+        val_str = lyplg_type_print((const struct lysc_type *)lref, lref->cur_mod->ctx, value, LY_VALUE_CANON, NULL, NULL, NULL);
         if (asprintf(errmsg, "Invalid leafref value \"%s\" - XPath evaluation error.", val_str) == -1) {
             *errmsg = NULL;
             ret = LY_EMEM;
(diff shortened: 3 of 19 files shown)


choppsv1:
This is provided as an example. Flushing it out would be moving doc comments from the `*_clb` functions and actually using these front-ends instead of directly calling the clb in general ly2 code.

choppsv1:
Consider the recent sort/compare/is_equal thread. In this case the `compare` front-end can call the `sort` callback if the plugin does not provide a `compare` callback. This makes sense e.g., when the `sort` function has the same overhead as the `compare` version to determine equality or not.

michalvasko:
> Many advantages to using front-ends for accessing plugin callbacks,
including allowing for optional callback functionality,

We have that, the `validate` callback. Being defined or not actually decides some functionality so a front-end simply makes no sense, the missing callback cannot be hidden.

> future expansion of plugin types,

How exactly does this help? We will have to change some internal code no matter what and these callbacks are not meant to be used directly (meaning the separate header is meant for new plugin implementations, not as part of the main library).

> default implementations, etc..

Yes, we have that. And I think I prefer the plugin having to explicitly use the default callback after recent experiences. Should make the author more aware about what exactly they are doing.

Please stop adding new and new changes, I would like to finish this at some point.

choppsv1:
> > Many advantages to using front-ends for accessing plugin callbacks,
> > including allowing for optional callback functionality,
> 
> We have that, the `validate` callback. Being defined or not actually decides some functionality so a front-end simply makes no sense, the missing callback cannot be hidden.

You miss the point. Once again, consider the 2 callbacks `compare` and `sort` being proposed. A plugin only needs to implement `sort` if there's a front end that uses it if `compare` doesn't exist:

```
static inline LY_ERR
lyplg_type_compare(const struct lysc_type *rt, const struct lyd_value *val1, const struct lyd_value *val2)
{
    if(rt->plugin->compare)
        return rt->plugin->compare(val1, val2);
    return rt->plugin->sort(val1, val2) ? LY_ENOT : LY_SUCCESS;
}
```

This is an example of how you can refine functionality down the road w/o having to change all the plugins

> 
> > future expansion of plugin types,
> 
> How exactly does this help? We will have to change some internal code no matter what and these callbacks are not meant to be used directly (meaning the separate header is meant for new plugin implementations, not as part of the main library).

See above.

> 
> > default implementations, etc..
> 
> Yes, we have that. And I think I prefer the plugin having to explicitly use the default callback after recent experiences. Should make the author more aware about what exactly they are doing.

That's fine, it was only one of the advantages. The point is to get the checks for this type of crap out of all the callers and into a single place though.

> Please stop adding new and new changes, I would like to finish this at some point.

This was in direct response to you removing the hash function. I already thought of this 3 days ago and didn't propose it b/c I didn't want to block the release or to "keep adding new and new changes". But then you went and proposed removing the hash function which opened this up to discussion.

michalvasko:
We will just not agree on this. If a plugin does not want to implement a specific `compare` callback, it can simply call its `sort` callback in the `compare` callback, I just do not see the harm. On the contrary, I like the fact that the plugin must explicitly do this in its own code so that these "rules" (such as "if you do not define `compare` callback, the `sort` callback will be called) are not outside the plugin code and are not enforced in any way.

> This was in direct response to you removing the hash function. I already thought of this 3 days ago and didn't propose it b/c I didn't want to block the release or to "keep adding new and new changes". But then you went and proposed removing the hash function which opened this up to discussion.

All I did was remove the hash callback because I realized it has absolutely no advantages over using the LYB value in the **current** design. You are proposing some design changes.

choppsv1:
> We will just not agree on this. If a plugin does not want to implement a specific `compare` callback, it can simply call its `sort` callback in the `compare` callback, I just do not see the harm. On the contrary, I like the fact that the plugin must explicitly do this in its own code so that these "rules" (such as "if you do not define `compare` callback, the `sort` callback will be called) are not outside the plugin code and are not enforced in any way.

There's a saying, missing the forest for the trees. I'm not talking about any specific use case here, I was just trying to give illustrative examples.

What I've proposed is abstracting the Interface from the Implementation. This is a very useful, well understood, and common abstraction in computer science. Especially for things like plugins. Sometimes this is referred to as "Northbound" and "Southbound".

> 
> > This was in direct response to you removing the hash function. I already thought of this 3 days ago and didn't propose it b/c I didn't want to block the release or to "keep adding new and new changes". But then you went and proposed removing the hash function which opened this up to discussion.
> 
> All I did was remove the hash callback because I realized it has absolutely no advantages over using the LYB value in the **current** design. You are proposing some design changes.

Right and instead it should have been fixed to be useful.

choppsv1:
Leaving aside the hash discussion this change is not API affecting (other than telling people not to directly call the plugin callbacks).

michalvasko:
> There's a saying, missing the forest for the trees. I'm not proposing/defending any specific use case here, I was just trying to give illustrative examples.

Okay, to use that, I do not want to have forests without trees. Once part of the library, it will need to be maintained, adjusted, and improved to still work and make sense with respect to the whole library or at least the type plugins. That will be my long-term task so I am keeping close watch on what gets to that stage. I have my experiences and paid the price for not doing this, several times.

> Right and instead it should have been fixed to be useful.

I tried to argue with reasons in some other PR why I do not believe it to be the case.

> Leaving aside the hash discussion this change is not API affecting (other than telling people not to directly call the plugin callbacks).

It does not affect the API at all, the plugin callback declarations are not even public if you do not specially include `plugins_types.h`, which (as hinted in the name) should be done only in some type plugin sources.

So, it is actually not relevant to the FRR use-case, it does not affect API or the performance and can be understood as an internal design suggestion. And like I said, I prefer it the way it is now and do not really see why you should care either way.

michalvasko:
> > Okay, to use that, I do not want to have forests without trees.
> 
> The saying means one is focusing on a tree (a detail), and this is causing one to not see the forest (missing the bigger picture)

Okay, my answer would basically be the same. If I do not see the bigger picture, I cannot be sure that there ever will be.

> > Once part of the library, it will need to be maintained, adjusted, and improved to still work and make sense with respect to the whole library or at least the type plugins. That will be my long-term task so I am keeping close watch on what gets to that stage. I have my experiences and paid the price for not doing this, several times.
> 
> So you think adding an interface abstraction is going to create _more_ work for you? You just removed hash functions and replaced them with direct calls to print(), if you'd like to back that change out and put back a useful hash at some point (or just find the cases where you are getting a hash from a type), you've now lost the info. If instead you had used a "hash" front end you would not have made this change you would have just changed the front end to call print() (as is done in this PR). And so the code still self-documents the cases where it's getting a hash vs getting the binary representation of the data.
> 
> It's OK, as you say you're the maintainer, if you don't like abstractions like north and south bound I guess that's your prerogative.

You may be right and it may be added later, once it makes sense. When adding it before that, you risk having an abstraction with no use. And I see no harm in doing it the other way, especially since all of this is internal.

PR #1555, opened 2021-05-10
libyang BUGFIX ly_getutf8 return code must be checked

lePici:
For all uses of the ly_getutf8 function, the return code must be checked.


src/common.h
```diff
@@ -432,13 +432,14 @@ ly_strlen(const char *s)
...
+ * @param[out] bytes Number of bytes (length) of the found prefix/substring starting at @p str_begin.
  * @param[out] is_prefix Type of substring found. Set to True for prefix, otherwise False.
  * @param[out] str_next Remaining string starting after prefix/substring and ending with @p str_end.
  * If the @p is_prefix is set to True then the colon character is skipped.
  * If no string remains, it is set to NULL.
- * @return Number of bytes (length) of the found prefix/substring starting at @p str_begin.
+ * @return LY_ERR value.
  */
-uint32_t ly_value_prefix_next(const char *str_begin, const char *str_end, ly_bool *is_prefix, const char **str_next);
+LY_ERR ly_value_prefix_next(const char *str_begin, const char *str_end, uint32_t *bytes, ly_bool *is_prefix, const char **str_next);
```
michalvasko, line 442: The description is fine but the name is not clear on its own. `len` would be better as it is obvious what it means.

michalvasko (changes requested):

PR #1560, opened 2021-05-11
build BUGFIX use LESS to support cmake 2.8 (centos 7)

choppsv1:


michalvasko:
The way it is now it also catches any non-number values, after your change it will not. But then the compilation fails immediately so I suppose no real harm done.

PR #1562, opened 2021-05-12
union BUGFIX bypass calloc for zero length value

lePici:


src/plugins_types/union.c
```diff
@@ -176,9 +176,13 @@ lyplg_type_store_union(const struct ly_ctx *ctx, const struct lysc_type *type, c
...
-        subvalue->original = calloc(1, value_len);
-        LY_CHECK_ERR_GOTO(!subvalue->original, ret = LY_EMEM, cleanup);
-        memcpy(subvalue->original, value, value_len);
+        if (value_len) {
+            subvalue->original = calloc(1, value_len);
+            LY_CHECK_ERR_GOTO(!subvalue->original, ret = LY_EMEM, cleanup);
+            memcpy(subvalue->original, value, value_len);
+        } else {
+            subvalue->original = "";
+        }
```
michalvasko, line 185: For the code to be a bit better organized, you could use `if () ... else if () ... else ...`. With the `else if` condition being `value_len` (execute the current code) and `else` with simply `subvalue->original = strdup("");` with error check. That way the member value is unified and always dynamic. You also get rid of one level of depth.

src/plugins_types/union.c
```diff
@@ -311,9 +315,13 @@ lyplg_type_dup_union(const struct ly_ctx *ctx, const struct lyd_value *original,
...
 
-    dup_val->original = calloc(1, orig_val->orig_len);
-    LY_CHECK_ERR_GOTO(!dup_val->original, LOGMEM(ctx); ret = LY_EMEM, cleanup);
-    memcpy(dup_val->original, orig_val->original, orig_val->orig_len);
+    if (orig_val->orig_len) {
+        dup_val->original = calloc(1, orig_val->orig_len);
+        LY_CHECK_ERR_GOTO(!dup_val->original, LOGMEM(ctx); ret = LY_EMEM, cleanup);
+        memcpy(dup_val->original, orig_val->original, orig_val->orig_len);
+    } else {
+        dup_val->original = "";
```
michalvasko, line 323: Connected to the previous comment, this should be allocate dynamically.

michalvasko (changes requested):

src/plugins_types/union.c
```diff
@@ -339,7 +347,9 @@ lyplg_type_free_union(const struct ly_ctx *ctx, struct lyd_value *value)
             val->value.realtype->plugin->free(ctx, &val->value);
         }
         lyplg_type_prefix_data_free(val->format, val->prefix_data);
-        free(val->original);
+        if (val->orig_len) {
+            free(val->original);
+        }
```
michalvasko, line 352: And of course this change will then be removed.

michalvasko (approved):

PR #1566, opened 2021-05-13
json parser BUGFIX duplicate in lydjson_subtree_r

lePici:
This PR fixes issue 32281 from oss-fuzz.


michalvasko (approved):
Cannot say I understand what is going on here but it seems sensible.

lePici:
> Cannot say I understand what is going on here but it seems sensible.

PR #1579, opened 2021-05-20
distro BUGFIX remove unnecessary build dependencies

idryzhov:
Tests and docs are not built by default in Release build mode so doxygen
and cmocka are not needed.


michalvasko:
Right, thanks you.

PR #1581, opened 2021-05-21
cmake BUGFIX do not error out on custom CMAKE_BUILD_TYPE

jsmolic:
* Discovered while trying to package version 2.0.0 on Gentoo Linux, where we use `CMAKE_BUILD_TYPE=Gentoo`, so don't error out on `CMAKE_BUILD_TYPE` different than the ones specified in `CMakeLists.txt`


michalvasko:
Okay, no problem.

PR #1590, opened 2021-05-25
[Minor]Change recursion limit error type from LYE_XML_INVAL to LYE_SPEC

fredgan:
This change helps the error message to print 'Recursion limit 10000 reached' instead of 'Recursion limit %d reached'


michalvasko:
Okay, thanks.

PR #1591, opened 2021-05-26
parser yang/json BUGFIX stack-oveflow protection

lePici:
By tracking the right and left braces.
This PR fixes issues 31630, 31858, 31998, 32066, 32934, 33439, 32392 from oss-fuzz.

### DIFF
--- a/src/json.c
+++ b/src/json.c
@@ -545,6 +545,8 @@ lyjson_object(struct lyjson_ctx *jsonctx)
     LY_CHECK_RET(skip_ws(jsonctx));
 
     if (*jsonctx->in->current == '}') {
+        assert(jsonctx->depth);
+        jsonctx->depth--;
         /* empty object */
         ly_in_skip(jsonctx->in, 1);
         lyjson_ctx_set_value(jsonctx, NULL, 0, 0);
@@ -626,6 +628,12 @@ lyjson_value(struct lyjson_ctx *jsonctx)
         LY_CHECK_RET(lyjson_array(jsonctx));
 
     } else if (*jsonctx->in->current == '{') {
+        jsonctx->depth++;
+        if (jsonctx->depth > LY_MAX_BLOCK_DEPTH) {
+            LOGERR(jsonctx->ctx, LY_EINVAL,
+                    "The maximum number of block nestings has been exceeded.");
+            return LY_EINVAL;
+        }
         /* object */
         ly_in_skip(jsonctx->in, 1);
         LY_CHECK_RET(lyjson_object(jsonctx));
@@ -697,6 +705,7 @@ lyjson_ctx_backup(struct lyjson_ctx *jsonctx)
     jsonctx->backup.value_len = jsonctx->value_len;
     jsonctx->backup.input = jsonctx->in->current;
     jsonctx->backup.dynamic = jsonctx->dynamic;
+    jsonctx->backup.depth = jsonctx->depth;
     jsonctx->dynamic = 0;
 }
 
@@ -712,6 +721,7 @@ lyjson_ctx_restore(struct lyjson_ctx *jsonctx)
     jsonctx->value_len = jsonctx->backup.value_len;
     jsonctx->in->current = jsonctx->backup.input;
     jsonctx->dynamic = jsonctx->backup.dynamic;
+    jsonctx->depth = jsonctx->backup.depth;
     jsonctx->backup.dynamic = 0;
 }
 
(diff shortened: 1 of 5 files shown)


src/json.c
```diff
@@ -626,6 +627,12 @@ lyjson_value(struct lyjson_ctx *jsonctx)
         LY_CHECK_RET(lyjson_array(jsonctx));
 
     } else if (*jsonctx->in->current == '{') {
+        (jsonctx->depth)++;
+        if (jsonctx->depth > LY_MAX_BLOCK_DEPTH) {
+            LOGERR(jsonctx->ctx, LY_EINT,
```
michalvasko, line 632: Not an internal error, they are only those that should never occur under normal circumstances (and are hence independent of any user input). This is probably `LY_EINVAL`.

src/parser_yang.c
```diff
@@ -751,7 +751,19 @@ get_keyword(struct lys_yang_parser_ctx *ctx, enum ly_stmt *kw, char **word_p, si
     word_start = ctx->in->current;
     *kw = lysp_match_kw(ctx->in, &ctx->indent);
 
-    if ((*kw == LY_STMT_SYNTAX_SEMICOLON) || (*kw == LY_STMT_SYNTAX_LEFT_BRACE) || (*kw == LY_STMT_SYNTAX_RIGHT_BRACE)) {
+    if (*kw == LY_STMT_SYNTAX_SEMICOLON) {
+        goto success;
+    }
+    if (*kw == LY_STMT_SYNTAX_LEFT_BRACE) {
```
michalvasko, line 757: Would expect `} else if {` here as well.

src/json.c
```diff
@@ -545,6 +545,7 @@ lyjson_object(struct lyjson_ctx *jsonctx)
     LY_CHECK_RET(skip_ws(jsonctx));
 
     if (*jsonctx->in->current == '}') {
+        jsonctx->depth = jsonctx->depth ? jsonctx->depth - 1 : 0;
```
michalvasko, line 548: Is the check for `0` really needed? How can the depth be 0 if the closing curly bracket was found? Or is this before the validation of these brackets is performed? But it seems it should be before.

michalvasko (changes requested):

src/json.c
michalvasko: Just put an `assert()` there if it should never be `0` to prevent possible underflow.

michalvasko: The point is that we are talking strictly about JSON here, which must have `}` for each `{`, otherwise it is not valid JSON. And invalid JSON should be detected in the generic JSON parser (`json.c`) and not left up to the higher-level parsers. But if you check where is `lyjson_object()` called, you will see that only after a `{` was parsed so I believe `assert()` is the way to go.

src/json.c
```diff
@@ -768,6 +775,9 @@ lyjson_ctx_next(struct lyjson_ctx *jsonctx, enum LYJSON_PARSER_STATUS *status)
             ret = lyjson_value(jsonctx);
         }
     } else if (((prev == LYJSON_OBJECT) && (*jsonctx->in->current == '}')) || ((prev == LYJSON_ARRAY) && (*jsonctx->in->current == ']'))) {
+        if (*jsonctx->in->current == '}') {
+            jsonctx->depth = jsonctx->depth ? jsonctx->depth - 1 : 0;
```
michalvasko, line 779: The zero-check `assert()` should also be here, after checking that it really must not occur and of it can, handle it accordingly.

src/xml.h
```diff
@@ -81,6 +81,7 @@ struct lyxml_ctx {
 
     enum LYXML_PARSER_STATUS status; /* status providing information about the last parsed object, following attributes
                                         are filled based on it */
+    uint32_t depth;         /* current number of nested blocks, see ::LY_MAX_BLOCK_DEPTH */
```
michalvasko, line 84: I am a bit unsure as to whether to add this explicit variable or not. But since you can directly use `xmlctx->elements.count` since the elements are always [removed](https://github.com/CESNET/libyang/blob/libyang2/src/xml.c#L547) and [added](https://github.com/CESNET/libyang/blob/libyang2/src/xml.c#L601), I think it is redundant. Comparing directly on the open element count has also the advantage of being obvious that it is actually limit on the open elements since there are no blocks in XML.

src/parser_yang.c
```diff
@@ -751,7 +751,18 @@ get_keyword(struct lys_yang_parser_ctx *ctx, enum ly_stmt *kw, char **word_p, si
     word_start = ctx->in->current;
     *kw = lysp_match_kw(ctx->in, &ctx->indent);
 
-    if ((*kw == LY_STMT_SYNTAX_SEMICOLON) || (*kw == LY_STMT_SYNTAX_LEFT_BRACE) || (*kw == LY_STMT_SYNTAX_RIGHT_BRACE)) {
+    if (*kw == LY_STMT_SYNTAX_SEMICOLON) {
+        goto success;
+    } else if (*kw == LY_STMT_SYNTAX_LEFT_BRACE) {
+        (ctx->depth)++;
```
michalvasko, line 757: I wanted to leave this be but have noticed that you are not consistent, a few lines below there is `ctx->depth--;`. Since the parentheses are redundant, please remove them.

michalvasko (changes requested):

michalvasko (approved):

PR #1592, opened 2021-05-26
libyang2: parser yang BUGFIX parsing revisions longer than LY_REV_SIZE

jvijtiuk:
Hello,

this PR fixes a heap buffer overflow in `parse_revision`

This bug was found by OSS-Fuzz in `lys_parse_mem`.
The file that caused the issue is in `tests/fuzz/corpus/lys_parse_mem/pull1592.yang` also added in this PR.
The issue can be tested by running `tests/fuzz/lys_parse_mem/regress_fuzz_lys_parse_mem` with the file as input.

The issue would appear when a revision longer than `LY_REV_SIZE` was
in the input model. `get_argument()` would parse the string and return
a length that is significantly bigger than `LY_REV_SIZE`.
The `word_len` is then cast into a `uint8_t` when `get_argument()` was called, and since `word_len` is a
`size_t` an overflow would occur, where only the first 10 characters of
the string would be read, resulting in the date appearing as valid in
`lysp_check_date()`. The `strncpy()` into `rev->date` would then result in an
overflow, as `word_len` bytes are copied, which in this case was larger than `LY_REV_SIZE`.


jvijtiuk:
It looks like the check should be for `(word_len > LY_REV_SIZE - 1)` and not `(word_len > LY_REV_SIZE)`, as LY_REV_SIZE includes the null byte.  I'll fix that and push it in a minute.

michalvasko:
Thanks for the fix but I think you have left another similar use-case when calling the function in `parser_yang.c`. To fix all such use-cases, it should be enough to simply change `lysp_check_date` parameter `date_len` from `uint8_t` to `size_t`.

jvijtiuk:
Thanks, you are correct, I missed the other case. The fix you suggested seems to be working, and the tests are passing too. I've edited the commit and force pushed on this branch again.

michalvasko (approved):

michalvasko:
Alright, thanks.

PR #1614, opened 2021-06-09
Fix <sys/cdefs.h> inclusion on non-BSD, non-glibc systems

wesleycoakley:
Fixes this build failure seen since #1534  on any non-glibc system:

```
/ly2/libyang-2.0.7/build # ldd --version
musl libc (x86_64)
Version 1.2.2
Dynamic Program Loader
Usage: /lib/ld-musl-x86_64.so.1 [options] [--] pathname
/ly2/libyang-2.0.7/build # make
[  1%] Building C object CMakeFiles/yangobj.dir/src/log.c.o
/ly2/libyang-2.0.7/src/log.c:16:10: fatal error: sys/cdefs.h: No such file or directory
   16 | #include <sys/cdefs.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/yangobj.dir/build.make:95: CMakeFiles/yangobj.dir/src/log.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:211: CMakeFiles/yangobj.dir/all] Error 2
make: *** [Makefile:149: all] Error 2
```
`<sys/cdefs.h>` is a part of BSD's libc ABI and is necessary to explicitly include only in that case. This worked incidentally on glibc systems because `<sys/cdefs.h>` is an internal header and *not* part of its ABI. For non-glibc systems (like the musl-based one above) attempting to include this file caused the breakage above (see [the musl FAQ entry about this](https://wiki.musl-libc.org/faq.html#Q:-When-compiling-something-against-musl,-I-get-error-messages-about-%3Ccode%3Esys/cdefs.h%3C/code%3E))

Solution is to only include this header on the BSD systems where it is needed.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -16,8 +16,8 @@
 /* realpath */
 #define _XOPEN_SOURCE 1
 #define _XOPEN_SOURCE_EXTENDED 1
-#endif
 #include <sys/cdefs.h>
+#endif
 
 #include "context.h"
 
--- a/src/diff.c
+++ b/src/diff.c
@@ -12,7 +12,9 @@
  *     https://opensource.org/licenses/BSD-3-Clause
  */
 #define _GNU_SOURCE /* asprintf, strdup */
+#if defined (__NetBSD__) || defined (__OpenBSD__)
 #include <sys/cdefs.h>
+#endif
 
 #include "diff.h"
 
--- a/src/log.c
+++ b/src/log.c
@@ -13,7 +13,9 @@
  */
 
 #define _GNU_SOURCE /* asprintf, strdup */
+#if defined (__NetBSD__) || defined (__OpenBSD__)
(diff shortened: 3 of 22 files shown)


michalvasko:
I was worried that some platforms may not have this header but it seemed fine, at first. So I have removed the include, it should not really be necessary. Note that the fix is only in `devel` and will get to `master` on the next release (which I do not know when will be, few weeks to few months).

wesleycoakley:
Does removing the `#include` all together affect BSD builds? I'll admit I don't have a box to check at the moment but it seemed like #1534 fixed it up so building on BSD worked again.

michalvasko:
I have kept most of the changes, just the explicit include was removed. It should be included in almost all the standard headers anyway so it should be fine.

PR #1623, opened 2021-06-15
Compile refactor

michalvasko:


src/schema_compile.c
```diff
@@ -1761,7 +1758,6 @@ lys_compile(struct lys_module *mod, uint32_t options, ly_bool recompile, struct
     LOG_LOCBACK(0, 0, 1, 0);
     lys_compile_unres_mod_erase(&ctx, ret);
     if (ret) {
-        lys_precompile_augments_deviations_revert(ctx.ctx, mod);
```
rkrejci, line 1764: how is this related to the commit f2a22c6?

src/schema_compile.c
```diff
@@ -1547,9 +1547,6 @@ lys_recompile(struct ly_ctx *ctx, ly_bool log)
             lysc_module_free(mod->compiled);
             mod->compiled = NULL;
         }
-
-        /* free precompiled iffeatures */
-        lys_free_feature_iffeatures(mod->parsed);
```
rkrejci, line 1552: commit message says "unused functions", but this function is used, why it can be removed now?

src/schema_compile.c
```diff
@@ -1486,35 +1486,40 @@ lys_compile_unres_mod_erase(struct lysc_ctx *ctx, ly_bool error)
...
- * @param[in] ctx Compile context.
+ * @param[in] mod Module to process.
  * @return LY_ERR value.
  */
 static LY_ERR
-lys_compile_identities(struct lysc_ctx *ctx)
+lys_compile_identities(struct lys_module *mod)
 {
-    LY_ERR ret = LY_SUCCESS;
+    struct lysc_ctx ctx = {0};
```
rkrejci: don't you want to initiate members directly here? Now they are initiated twice actually.

src/schema_compile.c
michalvasko: The message was not clear and now it should be better. I suppose it could have been a separate commit but the reasoning is the same, it is not needed to call this function here. Also, looking at it now, it is irrelevant because augments and deviations are not applied in `lys_compile()` anymore.

src/schema_compile.c
michalvasko: Yes, the message should be improved now.

src/schema_compile.c
michalvasko: Well, it would be possible but I prefer it the way it is now because this initialization simply zeros the memory (like `calloc`) and the later code fills in only the non-NULL members in an orderly fashion.

PR #1624, opened 2021-06-15
json BUGFIX integer under/overflow checks

lePici:
... in ::lyjson_exp_number().

This PR fixes issue 35193 from oss-fuzz.


michalvasko (commented):
I have no problem with the changes but it is quite difficult to check for underflow/overflow in general and I am not sure you have covered all the cases.

michalvasko (approved):

PR #1625, opened 2021-06-16
don't attempt to parse dlerror() in utests

wesleycoakley:
Remove some behavior in a unit test which depends on knowing the exact output of `dlerror()` when a bad plugin path is given.

Since this human-readable string difers between `libc` implementations (as seen by the previous `#ifdef __APPLE__`) and it would be too exhaustive to list all possible strings, only verify correct behavior by testing if we see a `LY_ESYS`.

I was seeing this issue specifically on FreeBSD and Alpine; I tried coding out all their `dlerror()` outputs originally but decided this was simpler :)


michalvasko:
Sure, no problem.

PR #1626, opened 2021-06-16
json parser BUGFIX duplicate in parser_json

lePici:
This PR fixes issue 34347 from oss-fuzz.


src/parser_json.c
```diff
@@ -666,8 +694,7 @@ lydjson_metadata(struct lyd_json_ctx *lydctx, const struct lysc_node *snode, str
     enum LYJSON_PARSER_STATUS status;
     const char *expected;
     ly_bool in_parent = 0;
-    const char *name, *prefix = NULL;
-    char *dynamic_prefname = NULL;
+    const char *name, *prefix = NULL, *dynamic_prefname = NULL;
```
michalvasko, line 697: Why did you changed the type? I would definitely not expect `dynamic_` variable to be `const` and see no reason to.

lePici: Because **::lyjson_ctx_submit_dynamic_value()** must have a **const** parameter, and because inside is **jsonctx->value** assignment, which is **const char**.

michalvasko: I disagree, `lyjson_ctx_submit_dynamic_value()` should not have the second parameter `const` since you are always assigning a dynamically-allocated value into it. `jsonctx->value` is `const` only because the value can be both `const` and non-`const` when we favor defining it as `const` and casting it where needed.

michalvasko (approved):

PR #1628, opened 2021-06-17
plugin types BUGFIX ::ly_err_new() format emphasis

lePici:
This commit prevents an error where err_msg could contain a conversion
specifier (character '%'), but the variadic arguments are empty, which
causes undefined behavior. Therefore, in the function, the err_msg
parameter has been renamed to err_format to emphasize that the function
expects a format. A compiler attribute has also been added to the
function to check the correct number of variadic arguments.

This PR fixes issue 35228 from oss-fuzz.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ as well as YANG 1.1 ([RFC 7950](https://tools.ietf.org/html/rfc7950)).
 
 * C compiler
 * cmake >= 2.8.12
-* libpcre2 >= 10.30 (including devel package)
+* libpcre2 >= 10.21 (including devel package)
  * note, that PCRE is supposed to be compiled with unicode support (configure's options
    `--enable-utf` and `--enable-unicode-properties`)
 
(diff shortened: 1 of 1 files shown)


src/plugins_types.h
```diff
@@ -213,13 +213,14 @@ struct lysc_type_leafref;
  * @param[in] vecode Validity error code in case of LY_EVALID error code.
  * @param[in] path Path to the node causing the error.
  * @param[in] apptag Error-app-tag value.
- * @param[in] err_msg error formating message.
+ * @param[in] err_format Format string (same like at printf) or string literal.
```
michalvasko, line 216: Mention the danger of using unknown variable (not in direct control of the caller) as format.

src/plugins_types.h
```diff
@@ -213,13 +213,14 @@ struct lysc_type_leafref;
...
  * @return The given @p ecode value if the @p err is successfully created. The structure can be freed using ::ly_err_free()
  * or passed back from callback into libyang.
  * @return LY_EMEM If there is not enough memory for allocating error record, the @p err is not touched in that case.
  * @return LY_SUCCESS if @p ecode is LY_SUCCESS, the @p err is not touched in this case.
  */
-LY_ERR ly_err_new(struct ly_err_item **err, LY_ERR ecode, LY_VECODE vecode, char *path, char *apptag, const char *err_msg, ...);
+LY_ERR ly_err_new(struct ly_err_item **err, LY_ERR ecode, LY_VECODE vecode, char *path, char *apptag, const char *err_format, ...)
+__attribute__((format (printf, 6, 7)));
```
michalvasko, line 223: Make the attribute conditional, look into `compat/compat.h.in` and do it the same way as `__unused__` or `__packed__`.

michalvasko (changes requested):

PR #1634, opened 2021-06-22
parse BUGFIX releasing sets in ::lys_parser_ctx

lePici:
Also added collision tests for grouping.
This PR fixes issue 35192 from oss-fuzz.

### DIFF
--- a/src/parser_yang.c
+++ b/src/parser_yang.c
@@ -2693,7 +2693,7 @@ checks:
     }
 
     /* store data for collision check */
-    if (parent && !(parent->nodetype & (LYS_GROUPING | LYS_RPC | LYS_ACTION | LYS_INPUT | LYS_OUTPUT | LYS_NOTIF))) {
+    if (parent) {
         assert(ctx->main_ctx);
         LY_CHECK_RET(ly_set_add(&ctx->main_ctx->tpdfs_nodes, parent, 0, NULL));
     }
--- a/src/parser_yang.c
+++ b/src/parser_yang.c
@@ -2977,7 +2977,7 @@ parse_grouping(struct lys_yang_parser_ctx *ctx, struct lysp_node *parent, struct
     grp->nodetype = LYS_GROUPING;
     grp->parent = parent;
 
-    YANG_READ_SUBSTMT_FOR(ctx, kw, word, word_len, ret, return LY_SUCCESS, return ret) {
+    YANG_READ_SUBSTMT_FOR(ctx, kw, word, word_len, ret, goto checks, return ret) {
         switch (kw) {
         case LY_STMT_DESCRIPTION:
             LY_CHECK_RET(parse_text_field(ctx, LY_STMT_DESCRIPTION, 0, &grp->dsc, Y_STR_ARG, &grp->exts));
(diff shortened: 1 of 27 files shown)


src/parser_yang.c
```diff
@@ -4678,6 +4678,8 @@ yang_parse_module(struct lys_yang_parser_ctx **context, struct ly_in *in, struct
     LOG_LOCBACK(0, 0, 0, 1);
     if (ret) {
         lysp_module_free(mod_p);
+        ly_set_erase(&(*context)->tpdfs_nodes, NULL);
+        ly_set_erase(&(*context)->grps_nodes, NULL);
         yang_parser_ctx_free(*context);
```
michalvasko, line 4683: Briefly looking at the code, I think the 2 set erases can be added to `yang_parser_ctx_free()`.

src/tree_schema.c
```diff
@@ -1607,6 +1607,7 @@ lys_parse_in(struct ly_ctx *ctx, struct ly_in *in, LYS_INFORMAT format,
 cleanup:
     if (pctx) {
         ly_set_erase(&pctx->tpdfs_nodes, NULL);
+        ly_set_erase(&pctx->grps_nodes, NULL);
```
michalvasko, line 1610: With the previous change, the whole `if` can be removed.

michalvasko (changes requested):
I think context freeing can be refactored a bit to make it simpler and also fix this problem, specifically always call free on only the particular context kind.

src/parser_yang.c
lePici: First, I thought the same thing, but it leads to Segfault. Look at function `yang_parse_submodule`  here:
https://github.com/CESNET/libyang/blob/c5722f39afc494d5c465ecdbad216f0e329b37c3/src/parser_yang.c#L4571
There are two contexts which points to the same dynamic memory and in the event of an error, it must not release sets in `context`. Maybe it would only work if `memcpy` were replaced by duplication of sets by `::ly_set_dup()`. This could increase code readability, but at the expense of unnecessary memory allocation.

Actually, current solution is not so complicated. Functions `::yin_parse_module()` or `yang_parse_module()` are those that creates sets and in the case of some error, they clean up after themselves. In other cases, the release is on `::lys_parse_in()`.

michalvasko: You are right, it is confusing and should be refactored including this code. There is a mechanism of having the main module context and submodule context in case a submodule is being parsed. It probably was not when the contexts were defined and it was obviously not implemented well. So, what you could do is make this principle obvious.

You could add `struct lys_parser_ctx *main_ctx;` to `struct lys_parser_ctx` that would be NULL when parsing the main module and point to the main context when parsing a submodule. When parsing a submodule, all the members `tpdfs_nodes`, `grps_nodes`, and `new_mods` would be unused and those in the main context accessed instead. It may make sense to define separate submodule parser structure but that would add refactoring work and should be fine without it.

That should simplify freeing (you always free the specific context so when freeing submodule context, the main one is left untouched) and also it should no longer be required to pass 2 contexts into `yang_parse_submodule`, for example. Lastly, it should prevent any more similar errors as it would be much better organized.

What I do not like about your solution is that it just adds the specific code to more places (where it is needed), ideally, no specific code should be needed and if, it should be made as transparent as possible (and not force separate freeing of these 2 sets based on the (hidden) knowledge of whether the context is for a module or submodule.

michalvasko (approved):

PR #1651, opened 2021-07-12
schema compile FEATURE leafref target checks

lePici:


src/schema_compile.c
```diff
@@ -1199,6 +1199,46 @@ lys_compile_unres_llist_dflts(struct lysc_ctx *ctx, struct lysc_node_leaflist *l
...
+    if (leaf_type->basetype == LY_TYPE_UNION) {
+        array_count = LY_ARRAY_COUNT(((struct lysc_type_union *)leaf_type)->types);
+        for (i = *index; !ret && (i < array_count); i++) {
+            union_type = ((struct lysc_type_union *)leaf_type)->types[i];
+            if (union_type->basetype == LY_TYPE_LEAFREF) {
+                ret = (struct lysc_type_leafref *)union_type;
+            }
+        }
+
+        *index = ret ? i : UINT64_MAX;
```
michalvasko, line 1231: This is a redundant mechanism. Generally, `index` value should always be the index of the next item to return. In case the index is equal to the count of the items, you do not return anything and can call the function repeatedly, it will never return anything.

src/schema_compile.c
```diff
@@ -1231,27 +1271,15 @@ lys_compile_unres_depset(struct ly_ctx *ctx, struct lys_glob_unres *unres)
...
-                if (((struct lysc_type_union *)type)->types[v]->basetype == LY_TYPE_LEAFREF) {
-                    lref = (struct lysc_type_leafref *)((struct lysc_type_union *)type)->types[v];
-                    ret = lys_compile_unres_leafref(&cctx, node, lref, unres);
-                    if (ret) {
-                        break;
-                    }
-                }
-            }
+        v = 0;
+        while ((ret == LY_SUCCESS) && (lref = lys_type_leafref_next(node, &v))) {
```
michalvasko, line 1301: I would personally keep this condition only for non-NULL `lref` and `break` in case the function in the loop returns an error. But you are free to keep it this way if you think it is more readable.

src/schema_compile.c
```diff
@@ -1255,22 +1253,27 @@ lys_compile_unres_depset(struct ly_ctx *ctx, struct lys_glob_unres *unres)
...
+    struct ly_path *path;
     LY_ARRAY_COUNT_TYPE v;
-    uint32_t i;
+    uint32_t i, processed_leafrefs = 0;
 
     ds_unres = &unres->ds_unres;
 
     /* fake compile context */
+again:
+    cctx = (struct lysc_ctx) {
```
michalvasko, line 1266: Please, no constructs like this, use `memset` instead.

src/schema_compile.c
```diff
@@ -1255,22 +1253,27 @@ lys_compile_unres_depset(struct ly_ctx *ctx, struct lys_glob_unres *unres)
...
     struct lys_depset_unres *ds_unres;
+    struct ly_path *path;
     LY_ARRAY_COUNT_TYPE v;
-    uint32_t i;
+    uint32_t i, processed_leafrefs = 0;
 
     ds_unres = &unres->ds_unres;
 
     /* fake compile context */
+again:
```
michalvasko, line 1265: Too vague, use something like `resolve_all` or at least `retry`.

src/schema_compile.c
```diff
@@ -1354,6 +1360,29 @@ lys_compile_unres_depset(struct ly_ctx *ctx, struct lys_glob_unres *unres)
 
         lysc_node_free(ctx, node, 1);
     }
+    /* also check if the leafref target has not been disabled */
```
michalvasko, line 1381: Add a newline before the comment.

michalvasko (changes requested):

src/schema_compile.c
lePici: ok

src/schema_compile.c
lePici: I think it's more readable because the code is shorter and the condition only consists of two parts, so when debugging it shouldn't be such a problem to check why the loop ended.

src/schema_compile.c
lePici: ok

src/schema_compile.c
lePici: ok

src/schema_compile.c
lePici: ok

src/schema_compile.c
michalvasko: Why I prefer the other solution is more because I think the conditions are different enough not to be in a single expression. In other words, I like error conditions to always be handled separately. But no real problem, just something to think about.

src/schema_compile.c
```diff
@@ -1199,6 +1197,43 @@ lys_compile_unres_llist_dflts(struct lysc_ctx *ctx, struct lysc_node_leaflist *l
...
+ * @return Pointer to the leafref or next leafref, otherwise NULL.
+ */
+static struct lysc_type_leafref *
+lys_type_leafref_next(const struct lysc_node *node, uint64_t *index)
+{
+    uint64_t array_count;
+    struct lysc_type *leaf_type, *union_type;
+    struct lysc_type_leafref *ret = NULL;
+
+    leaf_type = ((struct lysc_node_leaf *)node)->type;
```
michalvasko, line 1217: An assert should probably precede this line, that `node` is a term (`LYD_NODE_TERM`) node.

src/schema_compile.c
```diff
@@ -1199,6 +1197,43 @@ lys_compile_unres_llist_dflts(struct lysc_ctx *ctx, struct lysc_node_leaflist *l
...
+    if (leaf_type->basetype == LY_TYPE_UNION) {
+        array_count = LY_ARRAY_COUNT(((struct lysc_type_union *)leaf_type)->types);
+        for (uint64_t i = *index; !ret && (i < array_count); i++) {
+            union_type = ((struct lysc_type_union *)leaf_type)->types[i];
+            if (union_type->basetype == LY_TYPE_LEAFREF) {
+                *index = i + 1;
+                ret = (struct lysc_type_leafref *)union_type;
+            }
+        }
```
michalvasko, line 1226: It still seems redundantly complex to me. There are probably more ways of improving it.

michalvasko (changes requested):

michalvasko (approved):

PR #1655, opened 2021-07-14
Testing multiple/chained error messages

rkrejci:
Use the ability of libyang to provide a chain of error messages and check for all/more the messages in unit tests.

### DIFF
--- a/src/tree_schema.c
+++ b/src/tree_schema.c
@@ -1203,6 +1203,11 @@ lys_parse_submodule(struct ly_ctx *ctx, struct ly_in *in, LYS_INFORMAT format, s
     return LY_SUCCESS;
 
 error:
+    if (!submod || !submod->name) {
+        LOGERR(ctx, ret, "Parsing submodule failed.");
+    } else {
+        LOGERR(ctx, ret, "Parsing submodule \"%s\" failed.", submod->name);
+    }
     lysp_module_free((struct lysp_module *)submod);
     if (format == LYS_IN_YANG) {
         yang_parser_ctx_free(yangctx);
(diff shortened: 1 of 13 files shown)


src/tree_schema.c
```diff
@@ -1552,13 +1558,14 @@ lys_parse_in(struct ly_ctx *ctx, struct ly_in *in, LYS_INFORMAT format,
...
+        LY_CHECK_GOTO(ret = lys_parsed_add_internal_ietf_netconf(mod->parsed), cleanup);
     } else if (!strcmp(mod->name, "ietf-netconf-with-defaults")) {
-        LY_CHECK_GOTO(ret = lys_parsed_add_internal_ietf_netconf_with_defaults(mod->parsed), free_mod_cleanup);
+        LY_CHECK_GOTO(ret = lys_parsed_add_internal_ietf_netconf_with_defaults(mod->parsed), cleanup);
     }
 
     /* add the module into newly created module set, will also be freed from there on any error */
-    LY_CHECK_GOTO(ret = ly_set_add(new_mods, mod, 1, NULL), free_mod_cleanup);
+    LY_CHECK_GOTO(ret = ly_set_add(new_mods, mod, 1, NULL), cleanup);
+    module_created = 1;
```
michalvasko, line 1568: Would personally prefer the flag to be `module_in_ctx`, for example, and set it after it is added to the context a few lines below. Not sure what it means that "module is created".

michalvasko (changes requested):
Seems that formatting tests did not pass.

src/tree_schema.c
rkrejci: Changing the variable name is not a problem. 

Regarding setting the variable - it should be actually correct as it is. The critical for unres revert is the creating list. The module there is also tried to be removed from the context, but it is not an issue to not have the module there. Any other handling here will just complicate the code (it cannot be simply moved after adding to the context since it can fail and then we still need to remove the module from the creating list).

michalvasko: Okay then, I suppose it can stay this way.

michalvasko (approved):

PR #1669, opened 2021-08-04
identity BUGFIX compilation of identities moved

lePici:


src/tree_schema.h
```diff
@@ -2321,12 +2321,21 @@ struct lys_module {
...
 };
 
+/**
+ * @defgroup latestrevflags Options for ::lys_module.latest_revision.
+ *
+ * Various information bits of ::lys_module.latest_revision.
+ *
+ * @{
+ */
+#define LYS_MOD_LATEST_REV          0x01 /**< The latest revision in searchdirs was not searched yet
```
michalvasko, line 2348: My idea was to make this flag completely independent simply signifying that this is the latest revision currently in the context. No need to make it redundantly complex and force that only single of these flags can be set.

tests/utests/basic/test_context.c
```diff
@@ -313,12 +313,12 @@ test_imports(void **state)
     ly_ctx_set_module_imp_clb(UTEST_LYCTX, test_imp_clb, "module a {namespace urn:a; prefix a; revision 2019-09-17;}");
     assert_int_equal(LY_SUCCESS, lys_parse_mem(UTEST_LYCTX, "module a {namespace urn:a;prefix a;revision 2019-09-16;}",
             LYS_IN_YANG, &mod1));
-    assert_int_equal(1, mod1->latest_revision);
+    assert_int_equal(LYS_MOD_LATEST_REV, mod1->latest_revision);
```
michalvasko, line 319: Flags now, should ideally be tested with `&`.

src/tree_schema_helpers.c
```diff
@@ -804,94 +804,117 @@ lys_parse_localfile(struct ly_ctx *ctx, const char *name, const char *revision,
     return ret;
 }
 
-LY_ERR
-lys_parse_load(struct ly_ctx *ctx, const char *name, const char *revision, struct ly_set *new_mods,
-        struct lys_module **mod)
+static LY_ERR
```
michalvasko, line 807: Static functions should be commented.

michalvasko (changes requested):

michalvasko (approved):

PR #1685, opened 2021-08-24
lyb FEATURE printer/parser works with LY_VALUE_LYB

lePici:


src/plugins_types.c
```diff
@@ -285,11 +285,8 @@ lyplg_type_compare_simple(const struct lyd_value *val1, const struct lyd_value *
 
 API const void *
 lyplg_type_print_simple(const struct ly_ctx *UNUSED(ctx), const struct lyd_value *value, LY_VALUE_FORMAT UNUSED(format),
-        void *UNUSED(prefix_data), ly_bool *dynamic, size_t *value_len)
+        void *UNUSED(prefix_data), ly_bool *UNUSED(dynamic), size_t *value_len)
```
michalvasko, line 288: I do not like the changes in this commit. It somewhat simplifies the plugins but requires `dynamic` to always be initialized to 0. That is not convenient and actually changes to parameter to `in,out` instead of `out`.

src/plugins_types/enumeration.c
```diff
@@ -71,6 +71,13 @@ lyplg_type_store_enum(const struct ly_ctx *ctx, const struct lysc_type *type, co
                     *(int32_t *)value);
             goto cleanup;
         }
+
+        if (options & LYPLG_TYPE_STORE_DYNAMIC) {
+            free((void *)value);
+            options &= ~LYPLG_TYPE_STORE_DYNAMIC;
+        }
```
michalvasko, line 78: What is the purpose of this code here? It is always executed in cleanup.

michalvasko (changes requested):

src/plugins_types/binary.c
```diff
@@ -15,9 +15,6 @@
 #include "plugins_types.h"
 
 #include <ctype.h>
-#include <stdint.h>
```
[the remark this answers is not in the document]
lePici: What is the benefit of doing this?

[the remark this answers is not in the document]
michalvasko: Because the local headers may change or their includes may change, which would result in the header missing in this file.

src/plugins_types.c
lePici: I believe it is safer to not trust plugins because they can be created by users. There can be dozens of plugins, and if one of them does not set the **dynamic** flag to 0, a problem may arise in **libyang** that will be hard to find.
For example, let's look at the _lyd_validate_unique()_ function. The _print callback_ is called in a cycle. If the plugin with the wrong code does not set the flag to 0 or 1, then the _free_ function can be applied (because the previous plugin set the dynamic flag to 1) and deallocate a string from the _dictionary_, for example, and then a segfault occurs in some unrelated place.

On the other hand, if the change in the commit is applied, then if the user does not set the **dynamic** flag to 1 because it has allocated memory, it will not cause a **libyang** crash, but "only" a memory leak.

I don't understand why after the commit there should be a dynamic parameter of type `[in, out]`. The intent is that the user must set the **dynamic** flag if the return value is dynamically allocated. So it is still of type `[out]`.

src/plugins_types/enumeration.c
lePici: yes, it could be done better. It will be fixed.

src/plugins_types.c
michalvasko: The problem is that for the `dynamic` parameter to have a correct value after the print callback finishes, it must be initialized to `0`, otherwise it is impossible to determine whether it was set or not. Meaning it has some input value in addition to being changed and returned, so it is `in,out`.

As for not trusting the plugins, it is simply not possible. The print callback is described and all correct plugins must fully conform to the description. If they do not, they are invalid and should be fixed.

PR #1692, opened 2021-08-30
Fix some build warnings with GCC11

jktjkt:
## ly_time_ts2str: check against a negative value as well
    
This was found by GCC11:
    
```
[11/205] Building C object CMakeFiles/yangobj.dir/src/tree_data_helpers.c.o
/home/jkt/work/cesnet/gerrit/github/CESNET/libyang/src/tree_data_helpers.c: In function ‘ly_time_ts2str’:
/home/jkt/work/cesnet/gerrit/github/CESNET/libyang/src/tree_data_helpers.c:903:28: warning: ‘%09ld’ directive writing between 9 and 20 bytes into a region of size 10 [-Wformat-overflow=]
  903 |         sprintf(frac_buf, "%09ld", ts->tv_nsec);
      |                            ^~~~~
/home/jkt/work/cesnet/gerrit/github/CESNET/libyang/src/tree_data_helpers.c:903:27: note: directive argument in the range [-9223372036854775808, 999999999]
  903 |         sprintf(frac_buf, "%09ld", ts->tv_nsec);
      |                           ^~~~~~~
In file included from /nix/store/iwd8ic5hhwdxn5dga0im55g5hjl270cd-glibc-2.33-47-dev/include/stdio.h:866,
                 from /home/jkt/work/cesnet/gerrit/github/CESNET/libyang/src/tree_schema.h:23,
                 from /home/jkt/work/cesnet/gerrit/github/CESNET/libyang/src/tree_data.h:31,
                 from /home/jkt/work/cesnet/gerrit/github/CESNET/libyang/src/context.h:22,
                 from /home/jkt/work/cesnet/gerrit/github/CESNET/libyang/src/common.h:23,
                 from /home/jkt/work/cesnet/gerrit/github/CESNET/libyang/src/tree_data_helpers.c:24:
/nix/store/iwd8ic5hhwdxn5dga0im55g5hjl270cd-glibc-2.33-47-dev/include/bits/stdio2.h:38:10: note: ‘__builtin___sprintf_chk’ output between 10 and 21 bytes into a destination of size 10
   38 |   return __builtin___sprintf_chk (__s, __USE_FORTIFY_LEVEL - 1,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   39 |                                   __glibc_objsize (__s), __fmt,
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   40 |                                   __va_arg_pack ());
      |                                   ~~~~~~~~~~~~~~~~~
```
## build: Enable some optimizations even in a debug build
    
My distribution sets `_FORTIFY_SOURCE`, and the glibc headers issue a warning when optimizations are completely disabled. Apart from that, even the GCC manual suggests to use `-Og` instead of `-O0` for debugging.

## build: Require C11
    
This project uses anonymous structs, and modern versions of GCC and clang both warn about these unless the C standard version is properly set.
    
For some reason I was getting these warnings even though the CMakeLists.txt tries to set the `-std=` parameter manually. Let's use a native CMake feature for this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ set(LIBYANG_MICRO_SOVERSION 15)
 set(LIBYANG_SOVERSION_FULL ${LIBYANG_MAJOR_SOVERSION}.${LIBYANG_MINOR_SOVERSION}.${LIBYANG_MICRO_SOVERSION})
 set(LIBYANG_SOVERSION ${LIBYANG_MAJOR_SOVERSION})
 
-set(CMAKE_C_FLAGS                "${CMAKE_C_FLAGS} -Wall -Wextra -Wno-missing-field-initializers -std=c99")
+set(CMAKE_C_FLAGS                "${CMAKE_C_FLAGS} -Wall -Wextra -Wno-missing-field-initializers -std=c11")
 set(CMAKE_C_FLAGS_RELEASE        "-DNDEBUG -O2")
 set(CMAKE_C_FLAGS_DEBUG          "-g3 -O0")
 set(CMAKE_C_FLAGS_ABICHECK       "-g -Og")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,17 +23,21 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE Debug)
 endif()
 # normalize build type string
+# see https://github.com/CESNET/libyang/pull/1692 for why CMAKE_C_FLAGS_<type> are not used directly
 string(TOUPPER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_UPPER)
 if ("${BUILD_TYPE_UPPER}" STREQUAL "RELEASE")
     set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Build Type" FORCE)
+    set(CMAKE_C_FLAGS "-DNDEBUG -O2 ${CMAKE_C_FLAGS}")
 elseif("${BUILD_TYPE_UPPER}" STREQUAL "DEBUG")
     set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Build Type" FORCE)
+    set(CMAKE_C_FLAGS "-g3 -O0 ${CMAKE_C_FLAGS}")
 elseif("${BUILD_TYPE_UPPER}" STREQUAL "RELWITHDEBINFO")
     set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Build Type" FORCE)
 elseif("${BUILD_TYPE_UPPER}" STREQUAL "RELWITHDEBUG")
     set(CMAKE_BUILD_TYPE "RelWithDebug" CACHE STRING "Build Type" FORCE)
 elseif("${BUILD_TYPE_UPPER}" STREQUAL "ABICHECK")
     set(CMAKE_BUILD_TYPE "ABICheck" CACHE STRING "Build Type" FORCE)
+    set(CMAKE_C_FLAGS "-g -Og ${CMAKE_C_FLAGS}")
 elseif("${BUILD_TYPE_UPPER}" STREQUAL "DOCONLY")
     set(CMAKE_BUILD_TYPE "DocOnly" CACHE STRING "Build Type" FORCE)
 endif()
@@ -68,9 +72,6 @@ set(LIBYANG_SOVERSION_FULL ${LIBYANG_MAJOR_SOVERSION}.${LIBYANG_MINOR_SOVERSION}
 set(LIBYANG_SOVERSION ${LIBYANG_MAJOR_SOVERSION})
 
 set(CMAKE_C_FLAGS                "${CMAKE_C_FLAGS} -Wall -Wextra -Wno-missing-field-initializers -std=c11")
-set(CMAKE_C_FLAGS_RELEASE        "-DNDEBUG -O2")
-set(CMAKE_C_FLAGS_DEBUG          "-g3 -O0")
-set(CMAKE_C_FLAGS_ABICHECK       "-g -Og")
 
 include_directories(${PROJECT_BINARY_DIR}/src ${PROJECT_SOURCE_DIR}/src)
 
--- a/src/tree_data_helpers.c
+++ b/src/tree_data_helpers.c
@@ -896,7 +896,7 @@ ly_time_ts2str(const struct timespec *ts, char **str)
 {
     char frac_buf[10];
 
-    LY_CHECK_ARG_RET(NULL, ts, str, ts->tv_nsec <= 999999999, LY_EINVAL);
+    LY_CHECK_ARG_RET(NULL, ts, str, ((ts->tv_nsec <= 999999999) && (ts->tv_nsec >= 0)), LY_EINVAL);
 
     /* convert nanoseconds to fractions of a second */
     if (ts->tv_nsec) {


michalvasko:
Thanks for the effort but regarding each commit:
1. `CMAKE_C_STANDARD` is not available in `cmake` 2.8.12, we still need to support this version.
2. Not sure why `gcc` manual suggests using `-Og`, it makes debugging a pain and it is not used on purpose.
3. Right, these should have been fixed long time ago but somehow there are always other things to do. They have a purpose and we will try to fix ASAP.

jktjkt:
>  `CMAKE_C_STANDARD` is not available in `cmake` 2.8.12, we still need to support this version.

Sorry, the docs didn't mention that. I'll make this depend on a cmake version, then.

> Not sure why `gcc` manual suggests using `-Og`, it makes debugging a pain and it is not used on purpose.

What the code in `devel` is doing is pretty surprising because it silently overrides the user's choice with an ad-hoc choice. You might have wasted time debugging optimized code, others might have wasted time debugging performance issues in code where optimizations were siltnly turned off (I have been there in another project, so I know what I'm talking about here). Are you sure that your experience is **recent** and that it was about `-Og` and not something higher like `-O2`?

Anyway, if you want to "default to some optimizaiton level", at least put that into the `CFLAGS` *before* the user-provided options. It might be a bit painful in CMake, something like `set(CMAKE_C_FLAGS "-O0 ${CMAKE_C_FLAGS}")` made conditional upon an appropriate `CMAKE_BUILD_TYPE`. If you just put that into `CMAKE_C_FLAGS_DEBUG`, these bits override sane systemwide defaults.

> Right, these should have been fixed long time ago but somehow there are always other things to do. They have a purpose and we will try to fix ASAP.

I just moved them to the appropriate `#if` block.

Should I force-push here again, or would you prefer to cherry-pick what you want now and let me fix up the cmake bits later?

michalvasko:
> Sorry, the docs didn't mention that. I'll make this depend on a cmake version, then.

We have tried to avoid all such conditions. Is it really necessary to add it? If the `std=...` parameter is not used then there is something else wrong, which should be fixed. But you are the first to bring this up so I am not sure what to fix.

> What the code in devel is doing is pretty surprising because it silently overrides the user's choice with an ad-hoc choice. You might have wasted time debugging optimized code, others might have wasted time debugging performance issues in code where optimizations were siltnly turned off (I have been there in another project, so I know what I'm talking about here). Are you sure that your experience is recent and that it was about -Og and not something higher like -O2?

Yes, I am sure. I had the same experience as the user in this fairly recent [question](https://stackoverflow.com/questions/63386189/whats-the-difference-between-a-compilers-o0-option-and-og-option).

> Anyway, if you want to "default to some optimizaiton level", at least put that into the CFLAGS before the user-provided options. It might be a bit painful in CMake, something like set(CMAKE_C_FLAGS "-O0 ${CMAKE_C_FLAGS}") made conditional upon an appropriate CMAKE_BUILD_TYPE. If you just put that into CMAKE_C_FLAGS_DEBUG, these bits override sane systemwide defaults.

Feel free to improve the current way of doing this but it seemed to be working even if I used some custom C flags. Although it was never any `-Ox` option.

> I just moved them to the appropriate #if block.

Fine, I guess, if it bothers you that much.

> Should I force-push here again, or would you prefer to cherry-pick what you want now and let me fix up the cmake bits later?

Please address all the comments and you can force-push, it is not a problem.

jktjkt:
> We have tried to avoid all such conditions. Is it really necessary to add it? If the `std=...` parameter is not used then there is something else wrong, which should be fixed. But you are the first to bring this up so I am not sure what to fix.

This one turned out to be a PEBKAC on my side -- the problem only shows up when building sysrepo, not libyang. Sorry. I'll be dropping this patch (and fixing sysrepo instead).

> > Anyway, if you want to "default to some optimizaiton level", at least put that into the CFLAGS before the user-provided options. It might be a bit painful in CMake, something like set(CMAKE_C_FLAGS "-O0 ${CMAKE_C_FLAGS}") made conditional upon an appropriate CMAKE_BUILD_TYPE. If you just put that into CMAKE_C_FLAGS_DEBUG, these bits override sane systemwide defaults.
> 
> Feel free to improve the current way of doing this but it seemed to be working even if I used some custom C flags. Although it was never any `-Ox` option.

Yeah, my personal trade-off is to enable optimization and get an idea about how well the code works *in practice*, and I eat the cost of variables being optimized out and add `printf`s. I understand that your preferences makes sense as well.

I looked into the CMake options, and I'm afraid this will result in a lot of rather ugly code (see, e.g., [here](https://stackoverflow.com/questions/28732209/change-default-value-of-cmake-cxx-flags-debug-and-friends-in-cmake)). The problem once again is that `CMAKE_C_FLAGS_DEBUG` gets *unconditionally appended* to the effective value of `$CFLAGS`, and that in case of optimization flags, the last one seen by the compiler wins. So if you put `-O0` into `CMAKE_C_FLAGS_DEBUG` via `CMakeLists.txt`, the user cannot override that -- even if they invoked CMake via something like `CFLAGS=-O2 cmake ..`.

I think that it's rather confusing that the `CMAKE_<LANG>_FLAGS_<CONFIG>` cannot be used for this purpose -- to make sure that **unless specified otherwise**, a debug build gets `-O0`. It seems that [CMake's presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) are marketed for this purpose, or, alternatively, the [toolchain files](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html) could be (ab)used for that as well.

I would prefer a patch which simply drops these optimization levels and relies on "whatever systemwide defaults there are". You will then have to either `export CFLAGS=-O0`, or invoke CMake as, e.g., `cmake CFLAGS=-O0 ..` when you build to keep the current behavior. As a bonus, your users will be able to configure libyang similarly to any other C+CMake project that they are building. Will that work for you? If not I can get you a preset file which should do the right thing -- but you'll need CMake 3.19+ for that, and you'll have to invoke CMake via something like `cmake --preset=debug`. If that doesn't work for you either then I'll open-code this, which will be ugly. Which one do you prefer?

> > I just moved them to the appropriate #if block.
> 
> Fine, I guess, if it bothers you that much.

(I could not care less about these trivial warnings in the test suite, but they are noise which prevents me from benefiting from the *useful* warnings; that's why I'm fixing these.)

michalvasko:
So we are left with the `-Ox` parameter problem. The simplest fix to me seems to analyze what `CMAKE_C_FLAGS` there are and for this particular flag leave it as it is if set, otherwise set based on the build type. Not sure whether to just modify `CMAKE_C_FLAGS` or just drop `-Ox` from `CMAKE_C_FLAGS_<TARGET>` if it is found in `CMAKE_C_FLAGS`. What do you think about this and the solution?

jktjkt:
> EDIT: Simpler yet, the build-type-specific C flag variable should probably not be used and just prepend the parameters to `CMAKE_C_FLAGS`. That way you can always override them.

Done. If you''re happy with the solution, I'll fix it in the same way in sysrepo as well.

I put back the `-std=c11`, and here's the explanation from the commit message:

This project uses anonymous structs and anonymous unions which are not part of C99, but are available since C11. By default, compilers tend to enable these, but they backfire when building with `-pedantic` -- and sysrepo gets built with `-pedantic`, meaning that libyang causes extra warnings there.

Note that we still cannot enable `-pedantic` for libyang itself; some extra warnings are still being reported:

- casts between `void*` and function pointers,
- forward references to enums,
- void pointer arithmetics,
- `return` in a function returning void,
- zero-sized arrays,
- string literals longer than 4kB,
- empty initializer lists,
- macros with zero variadic arguments.

michalvasko:
Okay, thanks, merging this first before #1700.

PR #1694, opened 2021-09-06
decrease sublevel when skipping empty arrays and objects in lydjson_data_skip()

jvijtiuk:
Hello,

this PR should fix an assert violation found by OSS-Fuzz in `lyd_parse_mem_json`.

The input file looks like this: 
```
{"types:cont":{"":"","":{}}}
```

valgrind output looks like this:
```
$ valgrind ./tests/fuzz/lyd_parse_mem_json/regress_fuzz_lyd_parse_mem_json < ./crash
==19702== Memcheck, a memory error detector
==19702== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==19702== Using Valgrind-3.17.0 and LibVEX; rerun with -h for copyright info
==19702== Command: ./tests/fuzz/lyd_parse_mem_json/regress_fuzz_lyd_parse_mem_json
==19702==
regress_fuzz_lyd_parse_mem_json: /home/user/ly2/src/parser_json.c:1291: lydjson_subtree_r: Assertion `status == LYJSON_OBJECT' failed.
==19702==
==19702== Process terminating with default action of signal 6 (SIGABRT): dumping core
==19702==    at 0x4BFAD22: raise (in /usr/lib/libc-2.33.so)
==19702==    by 0x4BE4861: abort (in /usr/lib/libc-2.33.so)
==19702==    by 0x4BE4746: __assert_fail_base.cold (in /usr/lib/libc-2.33.so)
==19702==    by 0x4BF3615: __assert_fail (in /usr/lib/libc-2.33.so)
==19702==    by 0x146763: lydjson_subtree_r (parser_json.c:1291)
==19702==    by 0x146141: lydjson_parse_instance (parser_json.c:1188)
==19702==    by 0x146C9F: lydjson_subtree_r (parser_json.c:1407)
==19702==    by 0x147272: lyd_parse_json (parser_json.c:1542)
==19702==    by 0x12C9AD: lyd_parse (tree_data.c:363)
==19702==    by 0x12CF24: lyd_parse_data (tree_data.c:435)
==19702==    by 0x12CFA3: lyd_parse_data_mem (tree_data.c:446)
==19702==    by 0x1128A2: LLVMFuzzerTestOneInput (lyd_parse_mem_json.c:77)
```

The issue appears in `lydjson_data_skip()` when it encountered an open object with a nested empty object. 
It would increment the sublevel when the empty object was opened, but wouldn't decrement the sublevel when it was closed, as it only checked for `LYJSON_OBJECT_CLOSE` and not `LYJSON_EMPTY_OBJECT`.

This could result in `lydjson_data_skip()` continuing to parse the input, resulting in invalid state in its caller. 
In the case of the input found by OSS-Fuzz, `lydjson_data_skip()` would read the input until the end, which would then result in state being set to `LYSJON_END` which would violate the assert in `lydjson_subtree_r`.

This fix should also fix the same issue for nested empty arrays.


src/parser_json.c
```diff
@@ -290,7 +290,7 @@ lydjson_data_skip(struct lyjson_ctx *jsonctx)
         LY_CHECK_RET(lyjson_ctx_next(jsonctx, &current));
         if (current == status) {
             sublevels++;
-        } else if (current == status + 1) {
+        } else if (current == status + 1 || current == status + 2) {
```
michalvasko, line 293: Every boolean condition should be in parentheses so for tests to pass this should be
```C
} else if ((current == status + 1) || (current == status + 2)) {
```

michalvasko (changes requested):

michalvasko:
Thanks for this fix, I have tried to reproduce all standing issues in oss-fuzz but failed. HOwever, if you are fixing this code, please use the enum values instead of `status + x` please, it is quite unreadable this way.

jvijtiuk:
> Thanks for this fix, I have tried to reproduce all standing issues in oss-fuzz but failed.

This is issue 36501. I've managed to reproduce it by cloning libyang and building with:
`cmake -DCMAKE_BUILD_TYPE=Debug -DENABLE_FUZZ_TARGETS=ON ..`

To run the example I executed
```
./tests/fuzz/lyd_parse_mem_json/regress_fuzz_lyd_parse_mem_json < ~/Downloads/clusterfuzz-testcase-minimized-lyd_parse_mem_json-6182547028574208.fuzz
```
from within `build`.

The assert failed on both master and devel.

> HOwever, if you are fixing this code, please use the enum values instead of status + x please, it is quite unreadable this way.

Thanks for the suggestion, I've changed the code to use enums. It looks like the original `current == status + 1` code was written without enums to catch both the array and object case with one test. I split it into separate checks for arrays and objects, as having it all on one line results in a really long line.

michalvasko (approved):

michalvasko:
I have tried reproducing just using `yanglint`, I will try it properly next time. And I now also understand why the enum was not used but still think it is better this way. Merging.

PR #1700, opened 2021-09-22
Fix some warnings for pedantic compilation

michalvasko:
Refs #1692


jktjkt (approved):
Looks good to me on a quick glance. The result will still fail with `-std=c11 -pedantic` because of a zero-sized array in `CHECK_LYD_VALUE_BITS`, but that's not a regression.

michalvasko:
Yes, I wanted all the checks to be run because I did not manage to fix the last problem yesterday but did so today so now it should be fine with pedantic compilation enabled.

PR #1705, opened 2021-10-04
validation BUGFIX set error message and app tag when must not satisfied

idryzhov:
Model-supplied error message and app tag should be used when returning
an error to the user. This works correctly in libyang1.


michalvasko:
The whole `error-app-tag` support was mostly missing so I have added it not just for `must`.

idryzhov:
I was going to submit an issue about not-working `error-app-tag` but now everything works perfect – must/pattern/length/range. Thanks!

PR #1707, opened 2021-10-05
parser BUGFIX allow empty pattern

idryzhov:
Empty regular expression is explicitly allowed by the standard:

```
+---------------------------------+----------------------------------+
| For all branches S, and for all | Denoting the set of strings L(R) |
| regular expressions T, valid    | containing:                      |
| regular expressions R are:      |                                  |
+---------------------------------+----------------------------------+
| (empty string)                  | the set containing just the      |
|                                 | empty string                     |
+---------------------------------+----------------------------------+
| S                               | all strings in L(S)              |
+---------------------------------+----------------------------------+
| S|T                             | all strings in L(S) and          |
|                                 | and all strings in L(T)          |
+---------------------------------+----------------------------------+
```

### DIFF
--- a/src/parser_yang.c
+++ b/src/parser_yang.c
@@ -2057,12 +2057,6 @@ parse_type_pattern(struct lys_yang_parser_ctx *ctx, struct lysp_restr **patterns
     /* get value */
     LY_CHECK_RET(get_argument(ctx, Y_STR_ARG, NULL, &word, &buf, &word_len));
 
-    /* empty pattern is not valid */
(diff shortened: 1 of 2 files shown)


michalvasko:
You are right, I have re-read the specs and somehow missed that it is allowed.

michalvasko:
I see, that is a rather specific use and I have not seen it in any modules, YANG 1.1 is here for some years now. But fine, it is supported now and even released.

PR #1728, opened 2021-11-03
docs: Mention some flags in the transition manual

syyyr:
I was missing these in libyang2, so I added them to the transition manual, should other people need them too.

Feel free to change this if you don't like the wording.


michalvasko:
Alright, no problem.

jktjkt (approved):

PR #1733, opened 2021-11-11
update documentation to add reference to lyd_free when destroying ctx.

### LINKED ISSUE #1732: Add clarity in freeing data before destroying the context in docs
Hello,

I think a small addition to the documentation around destroying the context would be helpful. It confused me for a little while as it seemed to contradict itself. 

```
 * All instance data are supposed to be freed before destroying the context.
 * Data models are destroyed automatically as part of ::ly_ctx_destroy() call.
```
To me (probably confusion on my part of terminology), those two contradict, I think a reference to lyd_free would help.

```
 * All instance data are supposed to be freed before destroying the context.
 * Use ::lyd_free_all() call to free all data nodes.
 * Data models are destroyed automatically as part of ::ly_ctx_destroy() call.
```

Thanks

monotok:
Closes #1732

### DIFF
--- a/src/context.h
+++ b/src/context.h
@@ -603,6 +603,7 @@ LY_ERR ly_ctx_get_yanglib_data(const struct ly_ctx *ctx, struct lyd_node **root,
  * multiple contexts, the function should be called for each used context.
  *
  * All instance data are supposed to be freed before destroying the context.
+ * Use ::lyd_free_all() call to free all data nodes.
  * Data models are destroyed automatically as part of ::ly_ctx_destroy() call.
  *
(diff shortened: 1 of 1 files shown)


michalvasko:
Should be fixed now, we do not accept PRs to `master`.

monotok:
> Should be fixed now, we do not accept PRs to `master`. It will be merged there on the next release.

Oops my mistake. I guess pull requests should go to "devel" for future reference? Thank you!

michalvasko:
Yes, to `devel` only.

PR #1741, opened 2021-11-29
Get rid of volatile in favor of _Atomic

syyyr:
Hi, these things sometimes caused errors in our CI with ThreadSanitizer.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -20,6 +20,7 @@
 #include <stdint.h>
 #include <string.h>
 
+#include "compat.h"
 #include "context.h"
 #include "hash_table.h"
 #include "log.h"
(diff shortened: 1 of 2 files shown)


jktjkt (approved):

michalvasko (changes requested):
Using `_Atomic` directly may cause compatibility problems (even if using `-std=c11`) so please use macros from `compat.h` instead. Also, you should then use compat macros for any access of these variables, relaxed (no) memory barrier should be fine.

syyyr:
>Using _Atomic directly may cause compatibility problems (even if using -std=c11) so please use macros from compat.h instead. Also, you should then use compat macros for any access of these variables, relaxed (no) memory barrier should be fine.

Done, hopefully I got it right.

michalvasko (changes requested):
Alright, seems fine to me except missing include for `compat.h` in `common.h`.

syyyr:
I added the include

michalvasko (approved):

michalvasko:
Fine, I have no more problems with it, merging.

PR #1749, opened 2021-12-09
compat: remove unused `pthread_mutex_timedlock`

jktjkt:
It looks like this has never been used in libyang2.


michalvasko:
But it is used in *libnetconf2*, for example. We try to keep the `compat` mini-library the same in all the projects, makes maintenance easier.

jktjkt:
It would be nice to have a README which explains this.

michalvasko:
Well, I think this is developer information, not user information, only the latter is found in the README. We do not really have any central place for developer information, there are bits and pieces in the relevant parts of the code but that is all.

PR #1750, opened 2021-12-12
Sync the debian packaging into apkg format

oerdnj:
- Drop the versioned dependency on pcre2 (the oldest still supported Debian has
  pcre2 10.22 and the oldest still supported Ubuntu has pcre2 10.31)
- Update the symbols file up to upstream 2.0.7 version
- Update the yanglint autopkgtest for libyang2 packaging


michalvasko:
Alright, thanks.

PR #1754, opened 2021-12-14
Windows build

jktjkt:
# Building on Windows under MSVC
This patch series makes it possible to build libyang on Windows using MSVC. Important changes which I had to make:

- dynamic loading of plugins is not supported (check the commit description below)
- all dated are considered as "invalid timezone" for purposes of canonization (`struct tm` has no TZ info on Windows)
- there's no interactive shell in `yanglint`, and similarly there's no `yangre` at all due to missing libraries
- multi-dir module search path is specified via `;` due to Windows path format

The rest is the usual porting churn, compatibility functions, CRLF line ends, backslashes in paths, the usual fun. I tried to limit myself to `compat/`, but there's a bunch of `#ifdef`s where it made better sense.

---
## tests: do not hardcode paths under `/tmp`

This is a classic case of insecure file creation, aka CWE-377 (albeit "just" affecting developers, not users of libyang). This path is also not available on Windows.

## include: make self-sufficient

This was using the `LY_ERR` which lives in `log.h` without including anything.

## Do not try to compare strings using C preprocessor

This code apparently tried to feed a CMake-level string into a C preprocessor for comparison via `==`, which won't work. For example, it evaluates to true on MSVC. Fix this by checking compiler-specific macros.

## Allow building on platforms without `struct tm.tm_gmtoff`, `timezone`, `daylight`

I'm not happy about this because the results are non-intuitive, but that looks like something which originates in libyang's attempts to use local TZ everywhere instead of normalizing to UTC time.

But hey, this only matters on a platform which cannot "easily" represent timezones in `struct tm`, or for which we cannot determine the TZ/DST status without going the extra mile of writing non-POSIX stuff. So as long as we mark the returned value with "I cannot really vouch for the TZ", this *might* even be standard-compliant.

## MSVC: handle symbol visibility

On MSVC, the exported symbols have to be marked both in the header and in the implementation as `dllexport`, whereas on GCC the `-fvisibility=normal` attribute is only supposed to be used in the implementation. As a result, we need two types of macros here, one for the public headers, and the other for TUs with implementation. Also, the symbol name cannot be generic ("`API`") because we only want to mark as dllexport those symbols which are in a library that we're building *now*. Otherwise this will break havoc in any libraries which also use the `ABI` macro *and* also use libyang.

## MSVC: POSIX compatibility files

There's a bunch of packages on vcpkg, let's use them this way. I wasn't able to find anything for `<unistd.h>`, so I took that from https://stackoverflow.com/a/826027/2245623 .

## compat: `<unistd.h>`: make `X_OK` a no-op on Windows

There's nothing to check via `access()` on Windows for directory permissions, and I have no clue if the underlying `_access()` ignores unknown mode flags, so let's play it safe. The worst case is that `access()` will be too optimistic, which wouldn't really hart because of TOCTOU races, anyway. In other words, if anything breaks by this patch, then it's broken already.

## compat: there's no `<sys/time.h>` on MSVC

...but we won't be really using `gettimeofday()` there because `pthread_mutex_timedlock` is provided by the pthread compatibility library that we depend on.

## compat: make sure `ssize_t` is available

Normally this comes via `<sys/types.h>` on POSIX, but we're faking that on MSVC via our `<unistd.h>` shim. Make sure it's available.

## MSVC: compat: implement `localtime_r`

The Windows version of `localtime_s` swaps the argument order and has a different return type for increased level of fun.

## compat: MUSL implementation of `strptime()`

On Windows there's no `strptime()`. A similar functionality is available from C++11's `std::get_time` in `<iomanip>`, but there are subtle differences, including actual behavioral differences when parsing invalid dates such as "January 41st" (which is invalid everywhere, and could become "January 4th", eating one char less), or Feb 29th, 2018 (which does not exist).

Let's just import the MIT-licensed implementation from the MUSL C library because it's reasonably self-contained and appears to work the same as the glibc one.

## compat: strptime: lobotomize to not depend on `nl_langinfo`

...which is not available on Windows, and neither is `<strings.h>`.

## MSVC: compat: disable `pthread_mutex_timedlock` on Windows

On Windows there's no `gettimeofday()`/`nanosleep()`, and the usual suggestions come with a big disclaimer saying that it is not intended for high-resolution timing. Since this is not needed by the initial port of libyang to MSVC, let's just comment this out.

## MSVC: implement some of the atomic operations via MSVC' Interlocked ops

These actually look sane; the outliers are the load/store pairs, but given that the "alternative implementation" is just a plain old variable load...

## MSVC: compat: make the POSIX shims available to all targets

Some of them have been already defined, but others are being added via submodules. Best just make them available everywhere.

## MSVC: there's no `mmap()` on Windows

There's a compatibility library, the mman-win32, which implement a mmap-like semantics. However, on Windows one cannot map a read-only file into memory for a length that is larger than the size of that file, and you also cannot overwrite an existing mapping.

The easiest course of action is to just read the file into memory and be done with it.

See-also: https://stackoverflow.com/a/46014637/2245623

## MSVC: do not link with `-lm`

...because the Internet says that math functions are directly in MSVCR.

## MSVC: tools: disable interactive tools

I don't feel like porting `linenoise` to Windows, there's no `getline`, etc.

## MSVC: compat: `strcasecmp()` in `<strings.h>`

The `#ifdef` is for `_MSC_VER`, not `_WIN32`, because it's apparently available on MinGW.

## MSVC: do not build `fuzz/` and `style/` tests on Windows

Stuff like `waitpid` is a sure sign that `fuzz/` is better skipped when building on Windows. The `style/` is about compiling like C++ and about making sure that all files which need some functions include `compat.h`.  I don't feel like researching how to do `-Werror=c++-compat` on MSVC, and if the `compat.h` is ever missing, the MSVC build will yell loudly.

## MSVC: tests: link with all the compat libraries

This dance only has to be done because tests link against `yangobj` instead of simply using `-lyang` for linking against the library. I wonder why it's done like that.

## MSVC: tests: fix linking with `yangobj`

This fixes the LNK4217 linker warning. This project for some reasons creates an object library (`yangobjs` target in CMake) and proceeds to link tests against that instead of the usual way of linking against `libyang` the library properly. This is a problem on MSVC because the tests would normally process the API symbols as `__declspec(dllimport)`, while at the same time linking the object library with an implementation of these symbols into the same image. At the same time we cannot simply set `-DLIBYANG_BUILD` globally ebcause we're *also* building tools such as `yanglint` which does *not* embed these object files, but links against `-lyang` normally, and that one needs `__declspec(dllimport)`.

The alternative would be to get rid of the `yangobj` altogether, which would make the code more straightforward IMHO. That would also make the previous commit (`MSVC: tests: link with all the compat libraries`) superfluous because the usual rules of CMake transitive linking of libraries would take over.

See-also: https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-warning-lnk4217

## Windows: use binary I/O everywhere

...because newlines matter, at least in the test suite, and because having a stable file format really helps. No LF/CRLF changes for me, please.

## Windows: tests: adapt path handling

There's no exec bit on Windows, so don't test for that.

## Use semicolon as a directory delimiter on Windows

Windows path names have ':' inside as a drive letter separator. The widespread convention on that platform is to use ';' as a list separator in file paths, so let's follow suite.

It looks like this multi-value searchdir is only used in `ly_ctx_new()`, and that all other functions operate on one directory at a time.

## Windows: tests: do not invoke syscalls on an invalid FDs

On Windows, calling `fstat()` on a FD that was already closed results in calling the "invalid parameter handler". Set up on so that we can catch this properly. I'm doing this instead of just commenting out everything because I would prefer to keep that coverage to make sure we won't leak the FD in question.

## MSVC: tests: fix paths to utests and test plugins

On Unix, libyang plugins are not subject to `RUNTIME_OUTPUT_DIRECTORY`, instead requiring `LIBRARY_OUTPUT_DIRECTORY`. On Windows, the MSVC generator is a multi-config one, which means that it will happily put files into an extra subdirectory of `RUNTIME_OUTPUT_DIRECTORY`. This is super-hard to fix, so instead just query CMake for whatever the generator thinks is suitable, and use *that*.

This means that we have to distinguish between "the test binary directory", "directory with utests binaries" and "directory with test plugin binaries", so while we're at that, make sure that these paths are different than they used to be before this change -- simply to make sure that all bugs will be visible even on single-config generators like make or ninja. The tests still use the "binary dir" for things like `CMakeFiles` (!), which is super implementation-specific, but at least now the actual binaries live elsewhere.

Querying the `xxx_OUTPUT_DIRECTORY`, however, is not enough because that property does *not* have the config-specific suffix. Generators to the rescue, but that also means that the generated file suddenly has to be generator-config specific as well, which means that we have to massage the incldue path even more. Yay.

Also, on windows the test executables have the `.exe` suffix.

## MSVC: tests: use backslashes in expected normalized directory names

...and because `__FILE__` already has The Correct™ format even on Windows, use that for a non-existing file as well.

## tests: Plugins are not supported on Windows, yet

For some reason the `dlopen()` always fails, like this:

```
[ RUN      ] test_add_simple
libyang[2]: Searching for "ietf-yang-types" in "D:\a\oopt-gnpy-libyang\build\libyang\tests\utests".
libyang[2]: Newer revision than "ietf-yang-types@2013-07-15" not found, using this as the latest revision.
libyang[2]: Searching for "ietf-inet-types" in "D:\a\oopt-gnpy-libyang\build\libyang\tests\utests".
libyang[2]: Newer revision than "ietf-inet-types@2013-07-15" not found, using this as the latest revision.
libyang[0]: Loading "D:/a/oopt-gnpy-libyang/build/libyang/out-plugins/Release/plugin_simple.dll" as a plugin failed ("D:\a\oopt-gnpy-libyang\build\libyang\out-plugins\Release\plugin_simple.dll": The specified module could not be found.).
[  ERROR   ] --- 0x0 != 0x2
[   LINE   ] --- D:\a\oopt-gnpy-libyang\oopt-gnpy-libyang\libyang\tests\utests\basic\test_plugins.c:50: error: Failure!
[  FAILED  ] test_add_simple
```

Rather than trying to debug this on a platform that I don't really know, let's declare plugins as not supported on Windows.

## Allow building on Windows under MSVC as well

Legend of warnings:
- `/wd4711`, don't care about inline
- `/w14013`, missing prototype is a serious thing

## MSVC: source code is in UTF-8

This is important because the test suite uses `\Uxxxx` for some Unicode chars.

## CI: remove an unused job
    
This was used for the coverity job before, but that one got moved to another file.

Fixes: 7f55a30c actions REFACTOR separate devel push actions

## CI: start building on Windows

The differences compared to a POSIX build are too big even in the build script, mainly because CMake considers the Visual Studio build file generator a multiconfig one, and that means that all invocations of cmake/ctest have to have a specification of the current config. There's no way around that in CMake (and boy did I try!), so let's give up and do everything from scratch.


src/common.c
```diff
@@ -391,16 +394,48 @@ ly_mmap(struct ly_ctx *ctx, int fd, size_t *length, void **addr)
...
+        } else if (n < 0) {
+            if (errno == EINTR) {
+                continue; // can I get this on Windows?
+            }
+            LOGERR(ctx, LY_ESYS, "ly_mmap: read() failed (%s).", strerror(errno));
+        }
+        to_read -= n;
+        buf += n;
+    }
+#endif
```
michalvasko, line 424: I don't particularly like such a long code in `#ifdef` directly in the main source file. Maybe you could create a Windows-specific source file with functions called from the main source.

tools/lint/CMakeLists.txt
```diff
@@ -1,7 +1,12 @@
 # yanglint
 
+if(WIN32)
+    set(YANGLINT_INTERACTIVE OFF)
+else()
+    set(YANGLINT_INTERACTIVE ON)
+endif()
```
michalvasko, line 7: This is adding a fair amount of complexity for `yanglint` compilation, I do not like this. But I am not sure how to improve it so I am let this one go.

compat/posix-shims/strings.h
```diff
@@ -0,0 +1,7 @@
+#include <compat.h>
+
+#ifndef HAVE_STRCASECMP
+#ifdef _MSC_VER
+#define strcasecmp _stricmp
+#endif
```
michalvasko, line 6: Missing `#else`.

compat/compat.c
```diff
@@ -287,3 +287,20 @@ char *dirname(char *path)
...
+    if (!overwrite) {
+        size_t envsize = 0;
+        errcode = getenv_s(&envsize, NULL, 0, name);
+        if (errcode || envsize) {
+            return errcode;
+        }
+    }
+    return _putenv_s(name, value);
+}
+#endif
```
michalvasko, line 305: Add `#else`.

tests/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ function(ly_add_utest)
     endforeach()
 
     add_executable(${TEST_NAME} ${TEST_SOURCES} $<TARGET_OBJECTS:yangobj>)
+    target_compile_definitions(${TEST_NAME} PRIVATE LIBYANG_BUILD)
```
michalvasko, line 17: Comment for the commit adding this (saying that `yangobj` target should be removed altogether): I totally agree but the reason for it being written this way is that there are test for some functions in the internal *libyang* API. I was against this from the beginning (it was not my call then) and the latest plans were to fix this and rewrite the tests to use only public API but we did not make this change yet.

src/context.c
```diff
@@ -258,7 +258,7 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
         search_dir_list = strdup(search_dir);
         LY_CHECK_ERR_GOTO(!search_dir_list, LOGMEM(NULL); rc = LY_EMEM, cleanup);
 
-        for (dir = search_dir_list; (sep = strchr(dir, ':')) != NULL && rc == LY_SUCCESS; dir = sep + 1) {
+        for (dir = search_dir_list; (sep = strchr(dir, PATH_SEPARATOR_CHAR)) != NULL && rc == LY_SUCCESS; dir = sep + 1) {
```
michalvasko, line 261: Could just be `PATH_SEPARATOR_STR` renamed to `PATH_SEPARATOR` and here `PATH_SEPARATOR[0]` used.

tests/utests/basic/test_inout.c
```diff
@@ -44,6 +44,18 @@ test_input_mem(void **UNUSED(state))
...
+void empty_invalid_parameter_handler(const wchar_t* expression,
+   const wchar_t* function,
+   const wchar_t* file,
+   unsigned int line,
+   uintptr_t pReserved)
+{
+    ++invalid_parameter_handler_called;
+}
+#endif
+
```
michalvasko, line 58: For all I care you can remove this test instead of this wild code.

tests/CMakeLists.txt
```diff
@@ -51,3 +52,8 @@ if(ENABLE_PERF_TESTS)
 endif()
 
 set(format_sources ${format_sources} PARENT_SCOPE)
+
+get_target_property(TESTS_BIN_UTESTS utest_context RUNTIME_OUTPUT_DIRECTORY)
+get_target_property(TESTS_BIN_PLUGINS plugin_simple RUNTIME_OUTPUT_DIRECTORY)
+file(GENERATE OUTPUT "$<TARGET_FILE_DIR:utest_context>/tests_config2.h"
+    CONTENT "#define TESTS_BIN_UTESTS \"$<TARGET_FILE_DIR:utest_context>\"\n#define TESTS_BIN_PLUGINS \"$<TARGET_FILE_DIR:plugin_simple>\"\n")
```
michalvasko, line 59: `MSVC: tests: fix paths to utests and test plugins` This is another huge layer of complexity, why not just skip these tests instead?

tests/utests/basic/test_context.c
```diff
@@ -22,6 +22,36 @@
...
+{
+    char *tmp = strdup(value);
+#ifdef _WIN32
+    slashes_to_backslashes(tmp);
+#endif
+    LY_ERR ret = ly_ctx_unset_searchdir(ctx, tmp);
+    free(tmp);
+    return ret;
+}
+
```
michalvasko, line 54: `MSVC: tests: use backslashes in expected normalized directory names` Again, perhaps rather skip this test altogether.

src/plugins_types/ipv4_address.c
```diff
@@ -16,10 +16,15 @@
 
 #include "plugins_types.h"
 
-#include <arpa/inet.h>
-#if defined (__FreeBSD__) || defined (__NetBSD__) || defined (__OpenBSD__)
-#include <netinet/in.h>
-#include <sys/socket.h>
+#ifndef _WIN32
```
michalvasko, line 19: General minor note, I like more having `#ifdef` instead of `#ifndef`.

CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
-cmake_minimum_required(VERSION 2.8.12)
+if(WIN32)
+    cmake_minimum_required(VERSION 3.22.0)
+else()
+    cmake_minimum_required(VERSION 2.8.12)
+endif()
```
michalvasko, line 5: Don't really like this but I suppose if the newer version is strictly required not much to do about it.

michalvasko (changes requested):

src/build.h
```diff
@@ -0,0 +1,32 @@
...
+#else
+/*
+ * If the compiler supports attribute to mark objects as hidden, mark all
+ * objects as hidden and export only objects explicitly marked to be part of
+ * the public API.
+ */
+#  define LIBYANG_API_IMPL __attribute__((visibility("default")))
+#  define LIBYANG_API
+#endif
+
```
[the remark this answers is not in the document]
jktjkt: I realize that this is a "noisy change", but then that's why I've done that in a single commit over all of libyang. It's just the way that platform works; all symbols that are exported to a DLL must be marked as such by "method A" in the public header, and then by the same "method A" in the implementation, and then when user code includes these symbols, the first "method A" must become "method B". I tried to explain that in the commit message, but you're right that this needs docs.

I'll be happy to move it to any other header you indicate. Which one should I use? The only constrain is that it must be included by all public headers *and* by all `.c` files which implement any of these public symbols. Once I know the location, I'll add the docs as well.

compat/compat.c
```diff
@@ -250,3 +250,15 @@ pthread_mutex_timedlock(pthread_mutex_t *mutex, const struct timespec *abstime)
...
+#ifndef HAVE_REALPATH
+#ifdef _WIN32
+char *realpath(const char *path, char *resolved_path)
+{
+    char *resolved = _fullpath(resolved_path, path, PATH_MAX);
+    if (_access(resolved, 0) == -1 && errno == ENOENT)
+        return NULL;
+    return resolved;
+}
+#endif
```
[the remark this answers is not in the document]
jktjkt: Thanks, I'll change all implementations in `compat.c` (that's `realpath`, `localtime_r`, `dirname`, `setenv`).

src/build.h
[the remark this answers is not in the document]
michalvasko: Maybe use `config.h`? Also, when adding some comments, please rename `LIBYANG_API_IMPL` to `LIBYANG_API_DEF` (definition) and `LIBYANG_API` to `LIBYANG_API_DECL` (declaration), it is more clear to me.

tools/lint/CMakeLists.txt
jktjkt: Thanks. The alternative would be bringing in "something like `linenoise`", but for Windows, and I do not really prefer that :).

src/build.h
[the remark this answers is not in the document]
jktjkt: Done.

CMakeLists.txt
jktjkt: Agreed, which is why it's only for Windows. Some of the fancier target tricks are definitely a no-go on CMake 2.8.12, and since this is a new port and I can only test on the latest, it makes sense to use that IMHO.

tests/CMakeLists.txt
jktjkt: that makes sens,e I've adjusted the commit message.

src/common.c
jktjkt: Done.

tests/utests/basic/test_inout.c
jktjkt: It was there to check that the FD is not leaked, but I agree, that's a super-minor thing. Gone.

src/build.h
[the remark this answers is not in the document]
jktjkt: Just FYI, this requested change broke compilation under Windows due to the way how their headers are structured. I'll solve that via bringing the `compat.h` earlier in the include chain in a later commit (this is once again related to their [dreaded winsock implementation](https://stackoverflow.com/questions/1372480/c-redefinition-header-files-winsock2-h)).

tests/CMakeLists.txt
jktjkt: OK. I've added a trivial test for Windows to make sure that everything still works (this helped me before to catch that `:` path separator).

tests/utests/basic/test_context.c
jktjkt: Simplified now.

jktjkt:
OK, I think that all feedback has been incorporated now, and I made one final `uncrustify` sweep to make the CI happy (the changes appear minor and also touch `strptime.c` which is a third-party import, so I chose to make that sweep in an extra commit at the end).

jktjkt:
I don't think that the "NetDEF-CI" failures are specific to this patch series (e.g., the `yangre` manpage doesn't look relevant). Should I pay attention to them?

michalvasko:
No, it is fine, they are not relevant.

compat/faux_mmap.c
```diff
@@ -0,0 +1,80 @@
...
+ * Copyright (c) 2015-2021 CESNET, z.s.p.o.
+ *
+ * This source code is licensed under BSD 3-Clause License (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://opensource.org/licenses/BSD-3-Clause
+ */
+
+#include "common.h"
```
michalvasko, line 14: `compat` is meant to have only implementations of standard functions, this is *libyang* specific (the `common.h` include). Sorry for not mentioning any details, please either put these functions in a separate `common_win.c` or something like that to `libyang/src` (if you anticipate more functions may be added there later) or put these functions directly in `common.c`. I do not mind having `#ifdef WIN32` with a single function call in another Linux function, I am trying to avoid having large Windows-specific code block in a Linux function, that is all. So you may keep `ly_mmap` as it was and just call Windows code as needed with code `ly_win_mmap();`, for example.

tools/lint/main_ni_only.c
```diff
@@ -0,0 +1,23 @@
+/**
+ * @file main.c
+ * @author Radek Krejci <rkrejci@cesnet.cz>
+ * @brief libyang's yanglint tool
```
michalvasko, line 4: Update doxygen metadata.

CMakeLists.txt
```diff
@@ -71,8 +70,14 @@ set(LIBYANG_MICRO_SOVERSION 1)
 set(LIBYANG_SOVERSION_FULL ${LIBYANG_MAJOR_SOVERSION}.${LIBYANG_MINOR_SOVERSION}.${LIBYANG_MICRO_SOVERSION})
 set(LIBYANG_SOVERSION ${LIBYANG_MAJOR_SOVERSION})
 
-# global C flags
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wpedantic -std=c11")
+if(NOT WIN32)
```
michalvasko, line 73: Please switch to `if(WIN32)`.

michalvasko (changes requested):
Mostly minor thing left, do not worry if I do not merge this today, I will take a look and finish this tomorrow.

compat/faux_mmap.c
jktjkt: OK, moved back to `common.c`. The reason why I have not done it like that in my first series is because there's some overlap in the leading code of `ly_mmap` and I want to use that.

jktjkt:
> I made one final `uncrustify` sweep to make the CI happy

I've now separated out all the required changes to appropriate commits and checked that there's no coding style issue at any point (except for these two `strptime` commits which are fixed right after that). In order to make this checkable automatically, I had to fix an error in CMake setup (another `make` assumption while I'm on `ninja`) -- so there's a new commit right at the beginning of this patch series.

michalvasko (approved):

michalvasko:
Alright, thanks for the contribution, it will hopefully be helpful for more users.

PR #1757, opened 2021-12-17
README: updates for Windows, C++ bindings, etc

jktjkt:
See the commit messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,18 @@ the `distro` directory.
 
 * libpcre2 >= 10.21
 
+### Windows
+
+* Visual Studio 17 (2022)
+* cmake >= 3.22.0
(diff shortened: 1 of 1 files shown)


michalvasko (changes requested):
Thanks but I was thinking a separate section for WIN would be better, just those `Build Requirements`. If nothing else, `cmake` requirement is also higher and I do not think it would compile with any compiler. So the majority of the dependencies are different, anyway.

jktjkt:
> Thanks but I was thinking a separate section for WIN would be better, just those `Build Requirements`. If nothing else, `cmake` requirement is also higher and I do not think it would compile with any compiler. So the majority of the dependencies are different, anyway.

Done.

michalvasko:
Okay, thanks, I will do some final refactoring.

PR #1775, opened 2022-01-19
Check presence of list keys in subsequent entries in a list when parsing JSON input

metaswitch-ak10:
List's keys are mandatory in YANG. There is currently a bug where the presence of keys is only checked on the first entry in a list when parsing JSON input.

The key check happens in `lyv_data_content` at
https://github.com/CESNET/libyang/blob/6884c02a010ddf6b7d98f9725a2c4e53ecfe9fdb/src/validation.c#L715-L717
, which is only called if the `LYD_VAL_MAND` flag is set for the node. 

This flag is cleared from a node after successful validation at 
https://github.com/CESNET/libyang/blob/6884c02a010ddf6b7d98f9725a2c4e53ecfe9fdb/src/validation.c#L788-L789

When parsing JSON data, `lyv_data_content` is called iteratively on each entry in a list at 
https://github.com/CESNET/libyang/blob/6884c02a010ddf6b7d98f9725a2c4e53ecfe9fdb/src/parser_json.c#L1351
When moving onto the next entry in the list, the validation flags are copied from the previous entry
https://github.com/CESNET/libyang/blob/6884c02a010ddf6b7d98f9725a2c4e53ecfe9fdb/src/parser_json.c#L1363-L1364
, but at this point the `LYD_VAL_MAND` flag on the previous entry has been cleared, so the key check won't be performed on subsequent list entries.

To fix this, I have set the validation flags on a subsequent entry to their initial value.

I have also added a test which fails without this fix and made sure that the existing tests pass.

### DIFF
--- a/src/parser_json.c
+++ b/src/parser_json.c
@@ -1360,8 +1360,8 @@ attr_repeat:
                 new->prev = list;
                 list->next = new;
 
-                /* copy the validity and when flags */
-                new->validity = list->validity;
+                /* set the validity and when flags */
+                new->validity = ly_new_node_validity(list->schema);
                 new->when_status = list->when_status;
 
                 /* fix the "last" pointer */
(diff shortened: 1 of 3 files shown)


michalvasko:
Alright, thanks.

PR #1791, opened 2022-02-14
Fix windows time handling

jktjkt:
## MSVC: preserve date-and-time with unspecified TZ
    
This compatibility code was not stable for datetime values with unknown TZ. I think that the old behavior was not in violation of the YANG standard, but it was silly anyway -- if there's a timestamp with "unknown TZ", adding the *local* TZ before storing it doesn't make a lot of sense.
    
After this commit, the date-and-time handling on Windows becomes saner. These timestamps still won't remember their original TZ, but at least the absolute time will remain unchanged if we interpret "unknown" as "UTC", *and* there's now roundtrip stability.
    
Also expand the test coverage a bit.
    
Fixes: e182a27c Allow building on platforms without `struct tm.tm_gmtoff`, `timezone`, `daylight`

## MSVC: compat: implement `gmtime_r()`
    
Technically, there's no need for `gmtime_r()` because Windows `gmtime()` is thread-safe already, but we're using these thread-safe versions almost everywhere already.

### DIFF
--- a/CMakeModules/UseCompat.cmake
+++ b/CMakeModules/UseCompat.cmake
@@ -53,6 +53,7 @@ macro(USE_COMPAT)
 
     check_symbol_exists(realpath "stdlib.h" HAVE_REALPATH)
     check_symbol_exists(localtime_r "time.h" HAVE_LOCALTIME_R)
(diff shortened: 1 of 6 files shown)


michalvasko:
Great, thanks.

michalvasko (approved):

PR #1799, opened 2022-02-28
lys_getnext: Clarify iteration order

syyyr:
The doc hints that the order is schema order, but the order is not specifically written.
Also, I am not aware exactly how this function works:
>[lys_getnext()](https://github.com/CESNET/libyang/compare/group__schematree.html#gaeb9fc3101ead041d1e34e657c4cfb498) is supposed to be called sequentially. In the first call, the last parameter is usually NULL and function starts returning i) the first parent's child or ii) the first top level element of the module. Consequent calls suppose to provide the previously returned node as the last parameter and still the same parent and module parameters.

What happens if I specify both module and parent? 🤔 I think the "i)" and "ii)" should be changed, but I'm no exactly sure how.


michalvasko:
I have tried to improve the description.

PR #1819, opened 2022-03-19
Add stdint.h to utests.h

rwmacleod:
Needed to cross-build using Yocto.


michalvasko:
No problem but we do not accept PRs into `master` and the history seems not fully shared with `devel` so please create a new PR targeting the `devel` branch, thanks.

PR #1821, opened 2022-03-23
Accept empty lists in json

VandanaAkhramka:
Related issue - #1822 

If the node type is a `LYS_LEAFLIST`, then the code checks for `'['` to ensure the input has the begin array square brackets and then the code checks for the subsequent elements of the list. 
```
if (leaf->schema->nodetype == LYS_LEAFLIST) {
        /* expecting begin-array */
        if (data[len++] != '[') {
...
```
The code checks if the list contains the `null` json keyword, however, it erroneously checks for `[` again.
```
else if (data[len] == '[') {
        /* empty '[' WSP 'null' WSP ']' */
        for (r = len + 1; isspace(data[r]); ++r);
        if (strncmp(&data[r], "null", 4)) {
            goto inval;
        }
...
```
This allows `[[null]]` to be a valid entry, but not `[null]`. 
Additionally, the code never checks for an empty list without the keyword `[]` which is valid json. 

To fix this, I've added checks for `null]` and `]` after we check for `[` and if not present, we carry on checking for subsequent list elements as normal.

### DIFF
--- a/src/parser_json.c
+++ b/src/parser_json.c
@@ -456,22 +456,22 @@ json_get_value(struct lyd_node_leaf_list *leaf, struct lyd_node **first_sibling,
         unsigned int empty_list_len = len;
         // Skip whitespaces
         empty_list_len += skip_ws(&data[empty_list_len]);
-        bool is_empty_list = false;
+        unsigned int is_empty_list = 0;
         // Check if an empty list [] was provided
         if (data[empty_list_len] == ']') {
-            is_empty_list = true;
+            is_empty_list = 1;
         } 
         // Check if [null] was provided
         if (!strncmp(&data[empty_list_len], "null", 4)) {
-            ++empty_list_len
+            ++empty_list_len;
             // Skip whitespaces
             empty_list_len += skip_ws(&data[empty_list_len]);
             if (data[empty_list_len] != ']') {
                 goto inval;
             }
-            is_empty_list = true;
+            is_empty_list = 1;
         }
-        if is_empty_list {
+        if (is_empty_list != 0) {
             leaf->value_str = lydict_insert(ctx, "", 0);
             ++empty_list_len;
             empty_list_len += skip_ws(&data[empty_list_len]);
--- a/src/parser_json.c
+++ b/src/parser_json.c
@@ -456,22 +456,8 @@ json_get_value(struct lyd_node_leaf_list *leaf, struct lyd_node **first_sibling,
         unsigned int empty_list_len = len;
         // Skip whitespaces
         empty_list_len += skip_ws(&data[empty_list_len]);
-        unsigned int is_empty_list = 0;
         // Check if an empty list [] was provided
         if (data[empty_list_len] == ']') {
-            is_empty_list = 1;
-        } 
-        // Check if [null] was provided
-        if (!strncmp(&data[empty_list_len], "null", 4)) {
-            empty_list_len += 4;
-            // Skip whitespaces
-            empty_list_len += skip_ws(&data[empty_list_len]);
-            if (data[empty_list_len] != ']') {
-                goto inval;
-            }
-            is_empty_list = 1;
-        }
-        if (is_empty_list != 0) {
             leaf->value_str = lydict_insert(ctx, "", 0);
             ++empty_list_len;
             empty_list_len += skip_ws(&data[empty_list_len]);
@@ -549,6 +535,18 @@ repeat:
         }
         leaf->value_str = lydict_insert(ctx, &data[len], r);
         len += r;
+    } else if (data[len] == '[') {
+        /* empty '[' WSP 'null' WSP ']' */
+        for (r = len + 1; isspace(data[r]); ++r);
+        if (strncmp(&data[r], "null", 4)) {
+            goto inval;
+        }
+        for (r += 4; isspace(data[r]); ++r);
+        if (data[r] != ']') {
+            goto inval;
+        }
+        leaf->value_str = lydict_insert(ctx, "", 0);
+        len = r + 1;
     } else {
 inval:
         /* error */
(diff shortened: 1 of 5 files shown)


src/parser_json.c
```diff
@@ -452,6 +452,29 @@ json_get_value(struct lyd_node_leaf_list *leaf, struct lyd_node **first_sibling,
         if (data[len++] != '[') {
             LOGVAL(ctx, LYE_XML_INVAL, LY_VLOG_LYD, leaf, "JSON data (expected begin-array)");
             return 0;
+        } else {
+            unsigned int len2 = len;
```
johnbatty, line 456: doc: I think you need to have a comment here explaining what you are doing.
e.g.
```
// Check for an empty LEAFLIST here.
// This may either be '[ WS ]' or '[ WS none WS ]' (WS = whitespace).
```

johnbatty: suggestion: rename `len2` to `empty_list_len` to make it clearer what it is tracking.

src/parser_json.c
```diff
@@ -452,6 +452,29 @@ json_get_value(struct lyd_node_leaf_list *leaf, struct lyd_node **first_sibling,
...
+            // Strip whitespaces
+            for (r = len2 + 1; isspace(data[r]); ++r);
```
johnbatty, line 458: I think it would be simpler to just use skip_ws here?

src/parser_json.c
```diff
@@ -452,6 +452,29 @@ json_get_value(struct lyd_node_leaf_list *leaf, struct lyd_node **first_sibling,
         if (data[len++] != '[') {
             LOGVAL(ctx, LYE_XML_INVAL, LY_VLOG_LYD, leaf, "JSON data (expected begin-array)");
             return 0;
+        } else {
+            unsigned int len2 = len;
+            // Strip whitespaces
```
johnbatty, line 457: doc => "Strip whitespaces" => "Skip whitespace"
Strip implies removing something, but you are just moving past the whitespace.

src/parser_json.c
```diff
@@ -452,6 +452,29 @@ json_get_value(struct lyd_node_leaf_list *leaf, struct lyd_node **first_sibling,
...
+            for (r = len2 + 1; isspace(data[r]); ++r);
+            // Check if an empty list [] was provided
+            if (data[r] == ']') {
+                leaf->value_str = lydict_insert(ctx, "", 0);
+                len2 = r + 1;
+                len2 += skip_ws(&data[len2]);
+                return len2;
```
johnbatty, line 464: deduplication: This code is duplicated again just below.  I suggest you combine the two, and simply set a flag "is_empty_list" in here, plus in the next branch.  Then test the flag to determine whether to run this code.

src/parser_json.c
```diff
@@ -452,6 +452,29 @@ json_get_value(struct lyd_node_leaf_list *leaf, struct lyd_node **first_sibling,
...
+            } 
+            // Check if [null] was provided
+            else if (!strncmp(&data[r], "null", 4) {
+                // Strip whitespaces
+                for (r += 4; isspace(data[r]); ++r);
```
johnbatty, line 469: clarity: Again, simpler to use `skip_ws(...)` here?

src/parser_json.c
```diff
@@ -452,6 +452,29 @@ json_get_value(struct lyd_node_leaf_list *leaf, struct lyd_node **first_sibling,
...
+            for (r = len2 + 1; isspace(data[r]); ++r);
+            // Check if an empty list [] was provided
+            if (data[r] == ']') {
+                leaf->value_str = lydict_insert(ctx, "", 0);
+                len2 = r + 1;
+                len2 += skip_ws(&data[len2]);
+                return len2;
+            } 
+            // Check if [null] was provided
+            else if (!strncmp(&data[r], "null", 4) {
```
johnbatty, line 467: bug: Missing a closing parenthesis on the if expression.

src/parser_json.c
```diff
@@ -452,8 +452,18 @@ json_get_value(struct lyd_node_leaf_list *leaf, struct lyd_node **first_sibling,
...
             return 0;
-        }
-
+        } 
+        unsigned int empty_list_len = len;
+        // Skip whitespaces
+        empty_list_len += skip_ws(&data[empty_list_len]);
+        // Check if an empty list [] was provided
+        if (data[empty_list_len] == ']') {
+            leaf->value_str = lydict_insert(ctx, "", 0);
```
michalvasko, line 461: Setting "empty" value will not be enough, you want to free the whole `leaf` but that is probably not expected in this function so it will need more changes.

tests/data/test_json.c
```diff
@@ -449,6 +449,11 @@ static const char *string_data_024 =
 "}"
 ;
 
+static const char *json_empty_ll = 
+"{"
+  "\"null-leaf-list:ll\":[]"
```
michalvasko, line 454: You wanted to use prefix `null_leaflist_json`.

michalvasko (changes requested):

tests/data/test_json.c
```diff
@@ -449,6 +449,11 @@ static const char *string_data_024 =
 "}"
 ;
 
+static const char *json_empty_ll = 
+"{"
+  "\"empty_leaf_list:ll\":[]"
```
michalvasko, line 454: Still wrong, these are JSON data so prefix is the module name meaning the prefix should be `empty_leaflist_json`.

tests/data/test_json.c
```diff
@@ -694,6 +699,23 @@ test_parse_string(void **state)
     assert_ptr_equal(st->dt, NULL);
 }
 
+static void
+test_parse_ll_empty(void **state)
+{
+    struct state *st;
+    const char *modules[] = {"empty_leaf_list"};
```
michalvasko, line 706: Here you must also put the module name, there is no module with this name.

VandanaAkhramka: Apologies, I renamed the files and missed the exact name.

Issue #1829, opened 2022-03-31
resolve_unres_data cannot resolve all the when-stmt in some cases then cause lyd_parse_mem failure


happyisthebest:
In function resolve_unres_data, 
`while (progress && resolved < stmt_count)`
is used to judge whether all when-stmt are resolved. However, as resolved counts some non-when-stmt in subtree deleting, the loop
will end before all when-stmt are resolved.  This bug will cause lyd_parse_mem failure in some cases.  To fix this bug, the code should be modified as bellow.
```
    struct lys_when *when;
    int non_when_items = 0;
```
```
                        /* test if the node is in subtree to be deleted */
                        for (parent = unres->node[j]; parent; parent = parent->parent) {
                            if (parent == unres->node[i]) {
                                /* yes, it is */
                                if (unres->type[i] != UNRES_WHEN) {
                                    non_when_items++;
                                }
                                unres->type[j] = UNRES_RESOLVED;
                                resolved++;
                                break;
                            }
```
```
    } while (progress && resolved - non_when_items < stmt_count);

    /* do we have some unresolved when-stmt? */
    if (stmt_count > resolved - non_when_items) {
        goto error;
    }
```

michalvasko:
Yes, it seems so, feel free to provide a PR to the `libyang1` branch. Note that *libyang* v1 is obsolete and should not be used anymore.

PR #1830, opened 2022-04-01
Fix resolve_unres_data to resolve all the when-stmt

happyisthebest:
https://github.com/CESNET/libyang/issues/1829


michalvasko:
Fine, merging.

PR #1833, opened 2022-04-12
lyd_diff: check LYS_USERORDERED is valid for the node type

deastoe:
A LYD_DIFF_MOVEDAFTER2 diff was incorrectly being generated for a
leaf node which was a unique child of a list, since LYS_USERORDERED
and LYS_UNIQUE overload the same value (0x100).


michalvasko:
Fine, merging.

PR #1865, opened 2022-06-07
windows: fix path handling in the schema mount

jktjkt:
On Windows, libyang uses semicolon as path separator.

Fixes: 0fca45cf schema mount UPDATE use all the searchdirs
See-also: a3ea193c Use semicolon as a directory delimiter on Windows


michalvasko:
I see, thanks.

PR #1868, opened 2022-06-09
Fix compilation for FreeBSD/NetBSD

ton31337:
```
libyang-source/src/plugins_types/date_and_time.c:122:19: warning: arithmetic on a pointer to the function type 'char *(int, int)' is a GNU extension [-Wpointer-arith]
 	        val->time += timezone;
 	                  ^  ~~~~~~~~
libyang-source/src/plugins_types/date_and_time.c:122:19: warning: incompatible pointer to integer conversion assigning to 'time_t' (aka 'long') from 'char *(*)(int, int)' [-Wint-conversion]
 	        val->time += timezone;
 	                  ^  ~~~~~~~~
 	/usr/home/ci/cibuild.167/libyang-source/src/plugins_types/date_and_time.c:125:13: error: use of undeclared identifier 'daylight'
 	        if (daylight) {
 	            ^
2 warnings and 1 error generated.
```

* localtime_r() tells about DST already. In FreeBSD `daylight` variable
is not exposed. I believe we can use DST from tm?

Signed-off-by: Donatas Abraitis <donatas@opensourcerouting.org>

### DIFF
--- a/compat/compat.c
+++ b/compat/compat.c
@@ -265,7 +265,18 @@ realpath(const char *path, char *resolved_path)
     }
     return resolved;
 }
+#elif defined (__NetBSD__)
+char *
+realpath(const char *path, char *resolved_path)
+{
+    ssize_t nbytes;
 
+    nbytes = readlink(path, resolved_path, PATH_MAX);
+    if (nbytes == -1)
+        return NULL;
+
+    return resolved_path;
+}
 #else
 #error No realpath() implementation for this platform is available.
 #endif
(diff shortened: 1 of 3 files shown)


michalvasko:
Thanks, the patch looks fine and `daylight` was being checked only to save the call of `localtime_r`. So I think it can be merged but we do not accept PRs into `master` so please change the target branch and adjust your branch accordingly.

ton31337:
> Thanks, the patch looks fine and `daylight` was being checked only to save the call of `localtime_r`. So I think it can be merged but we do not accept PRs into `master` so please change the target branch and adjust your branch accordingly.

Oh, I missed reading https://github.com/CESNET/libyang/blob/master/CONTRIBUTING.md first :) Thanks.

ton31337:
Re-created https://github.com/CESNET/libyang/pull/1869

PR #1871, opened 2022-06-14
Do not include POSIX-specific bits from public headers

jktjkt:
On MSVC, there's no `<unistd.h>`. The only relevant types used directly via this header is `size_t` (which is already available via other headers, so let's use one of these instead), and `ssize_t`. The latter appears to be a POSIX extension. We already have a workaround in our compattibility, faux `unistd.h`, but it's problematic to install that one "publicly" -- and we should not use private bits from a public interface, anyway.

Fix this via using a MSVC `typedef`, and adjust the compat header as well. This means that there's now an `#ifdef` in a public header, but hey, it's just one thing, and there's no way around that.

### DIFF
--- a/compat/posix-shims/unistd.h
+++ b/compat/posix-shims/unistd.h
@@ -39,11 +39,7 @@
 #define timegm _mkgmtime
 /* read, write, and close are NOT being #defined here, because while there are file handle specific versions for Windows, they probably don't work for sockets. You need to look at your app and consider whether to call e.g. closesocket(). */
 
-#ifdef _WIN64
-#define ssize_t __int64
-#else
-#define ssize_t long
-#endif
+#define ssize_t SSIZE_T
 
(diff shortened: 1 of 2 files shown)


michalvasko:
I am a bit confused, the only change I see you did is add include for `sys/types.h`, why not keep the `unistd.h` include and use your header (from `posix-shims`) on Windows?

jktjkt:
That would mean installing the `unistd.h` as a part of libyang, and that's a big no-no because our tiny `unistd.h` shim does not contain all the features of a standard POSIX `unistd.h`. If we installed that, suddenly build systems would find our shim, and assume that all the functionality is there. I do not want to open that particular can of worms.

The other half of the change is a `typedef` for `SSIZE_T` which gets the job done on MSVC. That once again cannot be done via a compat header because that compat header is not currently being installed.

PR #1878, opened 2022-07-07
Fix EOL issues on Windows

jktjkt:
The test suite was broken on Windows, unless a particular git configuration was used during checkout. Since there are just two files where LF as the EOL absolutely have to be preserved, let's mark them as such in `.gitattributes`.

The `ly_mmap` was a bigger problem; I was getting parse errors on Windows when loading YANG modules via libyang: 
```
libyang[0]: Trailing garbage ":node-id";
   ..." after module, expected end-of-input. (path: Line number 193.)
```
This error message was especially entertaining; it seems to me that something, somewhere in the IO layer is getting out-of-sync. Rather than debugging this, prevent that in the Windows-compatibility `mmap` implementation by always configuring the FD to a binary mode.

## Windows: ensure we use binary IO
    
There's no point in behaving differently on Linux compared to Windows, so make sure we disable any behind-the-scenes IO character translation.

## tests: git: no CRLF substitutions are allowed

The `utest_parser_xml` and `utest_parser_json` check for some verbatim string literals from these on-disk YANG models, so make sure these are kept as-is by git. The standard says that these should not be normalized, and that whatever is in these multiline string literals is what should really be part of the resulting attribute.

## CI: Windows: test with CRLF transformations disabled as well

Since the code really, really assumes that CR and LF are independent characters, make sure that we test with various repository settings when building on Windows.


michalvasko:
No problem, thanks.

PR #1895, opened 2022-08-11
Printing features in a new format in yanglint 

Roytak:
Added the ability to print all features in a format, which can later be used to specify, which features a model should use, for both interactive and non-interactive versions of yanglint. Interactive version also supports doing this for all implemented modules at once.

### DIFF
--- a/tools/lint/cmd_feature.c
+++ b/tools/lint/cmd_feature.c
@@ -27,25 +27,14 @@
 void
 cmd_feature_help(void)
 {
-    printf("Usage: feature [-h] <module> [<module>]*\n"
-            "                  Print features of all the module with state of each one.\n");
-}
-
-int
-collect_features(const struct lys_module *mod, struct ly_set *set)
-{
-    struct lysp_feature *f = NULL;
-    uint32_t idx = 0;
-
-    while ((f = lysp_feature_next(f, mod->parsed, &idx))) {
-        if (ly_set_add(set, (void *)f->name, 1, NULL)) {
-            YLMSG_E("Memory allocation failed.\n");
-            ly_set_erase(set, NULL);
-            return 1;
-        }
-    }
-
-    return 0;
+    printf("Usage: feature [-f] <module> [<module>]*\n"
+            "       feature -a [-f]\n"
+            "                  Print features of all the modules with state of each one.\n\n"
+            "  -f <module1, module2, ...>, --feature-param <module1, module2, ...>\n"
+            "                  Generate features parameter for the command \"add\" \n"
+            "                  in the form of -F <module-name>:<features>\n"
+            "  -a, --all \n"
+            "                  Print features of all implemented modules.\n");
 }
 
 void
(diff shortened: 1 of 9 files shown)


tools/lint/main_ni.c
```diff
@@ -944,6 +962,7 @@ main_ni(int argc, char *argv[])
 cleanup:
     /* cleanup */
     erase_context(&c);
+    free(features_output);
```
lePici, line 965: I think there is missing to release `set` when `generate_features_output` fails.

tools/lint/main_ni.c
```diff
@@ -576,6 +577,8 @@ fill_context(int argc, char *argv[], struct context *c)
             } else if (!strcasecmp(optarg, "lyb")) {
                 c->schema_out_format = 0;
                 c->data_out_format = LYD_LYB;
+            } else if (!strcasecmp(optarg, "feature-param")) {
+                c->feature_param_format = 1;
```
lePici, line 581: What about `c->data_out_format`? All `if` branches set it somehow, but this one doesn't. Is it intentional?

tools/lint/main_ni.c
```diff
@@ -895,7 +900,20 @@ main_ni(int argc, char *argv[])
         /* print the list of schemas */
         print_list(c.out, c.ctx, c.data_out_format);
     } else {
-        if (c.schema_out_format) {
+        if (c.feature_param_format) {
+            for (uint32_t u = 0; u < c.schema_modules.count; u++) {
```
lePici, line 904: We usually declare a variable at the beginning of a function. But here it seems not to be followed. Maybe you can also add a refactoring commit and fix it. I'll leave it up to you.

tools/lint/common.c
```diff
@@ -209,6 +209,137 @@ fill_cmdline_file(struct ly_set *set, struct ly_in *in, const char *path, LYD_FO
...
+            }
+        }
+
+        /* print features */
+        for (j = 0; j < set->count; ++j) {
+            name = set->objs[j];
+            printf("\t%-*s (%s)\n", (int)max_len, name, lys_feature_value(mod, name) ? "off" : "on");
+        }
+    } else {
+        printf("\t(none)\n");
```
lePici, line 255: Maybe it would be more appropriate:
`if (!set->count) {  printf("\t(none)\n\n"); return; }`
This would prevent the `if` immersion.

lePici (commented):
I only did a code review. I haven't tried the printing functionality yet.

tools/lint/cmd_feature.c
```diff
@@ -27,65 +27,74 @@
...
         }
     }
 
+    if (print_all) {
+        if (print_all_features(*ctx, generate_features, &features_output)) {
+            YLMSG_E("Printing all features failed.\n");
+            goto cleanup;
+        }
+        if (generate_features) {
+            printf("%s\n", features_output);
```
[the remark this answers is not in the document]
Roytak: So, if you run `feature <module>`, the module's features will get printed and their status (on/off).
If you run `feature -a`, it will do the same as above, but for all the implemented modules.
If you run `feature -g <module>`, a parameter generated from the given module's features will be printed (e.g., `feature -g ietf-tls-common` -> ` -F ietf-tls-common:tls10,tls11,tls12,tls13,hello-params,public-key-generation`).
Lastly, if you run `feature -g -a`, it will be a combination of the second and third case, meaning all the implemented modules' features will be printed, along with their status, and at the end the generated parameter.

Hence I chose `Usage: feature [-g] <module> [<module>]* | [-a]`.

tools/lint/main_ni.c
Roytak: I omitted it, because it is already set to 0 (in function main_ni there is `struct context c = {0};`).

tools/lint/cmd_feature.c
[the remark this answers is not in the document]
lePici: Ok, thanks for the clarification. I would choose:
`Usage: feature [-g] <module> [<module>]*`
`...... feature -a [-g]`  (replace the dots with spaces)

to avoid confusion, if the operator `|` refers to `<module> [<module>]*` or `[-g] <module> [<module>]*`

michalvasko (approved):

PR #1900, opened 2022-08-22
tests: do not assume that `/usr/bin/expect` exists

jktjkt:
On my system, there's no shared `/usr/bin/`. Since the code already tries to check whether `expect` is in the path, let's use that information instead of hardcoding the path.

(The `shunit2` tests are OK because they go via `/usr/bin/env`, but that one doesn't support passing multiple arguments, and that's actually needed for the `expect -f` shebang. Unless we could depend on a [non-portable argument `--split-string` to GNU coreutils 8.30+](https://stackoverflow.com/a/52979955/2245623)).


michalvasko:
Interesting solution but fine, merging.

PR #1901, opened 2022-08-22
Yanglint schema mount

ekinzie:
This makes a couple changes to yanglint so that (1) tree diagrams can include mount points and (2) a schema context (yang-library instance, etc.) can be specified for mount points when validating data.


src/tree_schema.h
```diff
@@ -2392,6 +2392,7 @@ struct lys_module {
     ly_bool implemented;             /**< flag if the module is implemented, not just imported */
     ly_bool to_compile;              /**< flag marking a module that was changed but not (re)compiled, see
                                           ::LY_CTX_EXPLICIT_COMPILE. */
+    ly_bool internal;                /**< flag indicating that this is an automatically loaded internal module */
```
michalvasko, line 2395: This will not be accepted because you are changing the ABI. But I do not understand what it is needed for, anyway. Why do you want to skip internal modules in the tree output?

src/plugins_exts.h
```diff
@@ -214,6 +224,7 @@ struct lyplg_ext {
     lyplg_ext_data_snode_clb snode;         /**< callback to get schema node for nested YANG data */
     lyplg_ext_data_validate_clb validate;   /**< callback to validate parsed data instances according to the extension
                                                  definition */
+    lyplg_ext_get_context_clb context;      /**< callback to get extension instance context */
```
michalvasko, line 227: Adding a new callback is the last resort because it changes plugins API and is useless for the great majority of extensions. But if you make sure that an extension instance is a `mount-point`, you should be able to safely access its data directly and get the context.

michalvasko (changes requested):
Thanks for the effort but there are some serious problems with your solution so it will need some more work.

src/tree_schema.h
ekinzie: I don't have a strong opinion on this and don't mind removing it.

The reason I filtered these out to begin with was because it looked to me like yanglint was reporting on libyang implementation details.  The internal modules appear in the diagram even if they are not specified in the module-set instance for the mount point.  Some of the modules are generally necessary, but for example, libyang loads ietf-yang-schema-mount even if none of the modules in the context has a mount point.  I wasn't sure that it really makes sense to display this.

src/plugins_exts.h
ekinzie: The logic to retrieve the context in this case is slightly less straightforward than it sounds.  It's buried in a structure defined locally in the schema mount plugin (struct lyplg_ext_sm), so applications don't have access to that.  Perhaps that structure definition and some supporting functionality could be moved from the plugin into the library, but that sort of defeats the idea of a plugin.

src/tree_schema.h
michalvasko: I think it is because you can always instantiate the internal modules data in the mounted data so they technically are always part of the yang-library data.

src/plugins_exts.h
michalvasko: I would expect something similar as in the case of type plugins. There are lots of plugin-specific functions in the public header `plugins_types.h`. So, write a new API function `lyplg_ext_schema_mount_get_context(const struct lysc_ext_instance *ext)`, put the implementation into `schema_mount.c` and the declaration into `plugins_exts.h`. The function should then check that it is in a `mount-point` extension instance and if so, return the context. Of course, there is still the problem of which context to return (which your callback has not solved either) because there can be lots of data parsed with the context and each of those could have created a different context.

ekinzie: I have implemented the new API function and removed the "internal" marking for modules.  The context returned is based on the yang-library data returned by the extension callback function.  Perhaps I'm missing something.

michalvasko: Good, there is progress but your callback does not retrieve any existing context, it creates a new one every time it is called instead. I do not think that was your intention and it is most likely not the right thing to do. So you will have to solve the problem of returning a particular context somehow.

michalvasko (commented):
After another look at this, I have some more comments. Firstly, you wanted to skip internal modules and you are probably right about that, I mainly did not like the way you implemented it. There is function `ly_ctx_internal_modules_count()` that you can use to skip all the internal modules without any changes to the API. For the iteration over the next modules then use `ly_ctx_get_module_iter()`.

Secondly, about the context problem, there is probably no other way of implementing it except the way you did it meaning always creating a new context. So you can keep it that way but rename the schema mount function to `lyplg_ext_schema_mount_create_context()`, for example, to make the functionality obvious. Thanks,

lePici:
To make YANG Schema Mount Tree Diagrams (printer_tree) complete, please add support for printing `/` and `@` special characters. See [opts](https://www.rfc-editor.org/rfc/rfc8340.html#section-2.6).
Please also edit your [example](https://github.com/CESNET/libyang/pull/1901/commits/41d8bdeabc9a6c2c317d370890690e6b49d74b40) to show that these symbols print correctly.

ekinzie:
@michalvasko internal modules are again skipped, using ly_ctx_get_module_iter().  I have also renamed the one function to lyplg_ext_schema_mount_create_context().

@lePici I have added the missing "opts" to the tree printer.

michalvasko (changes requested):
Thanks, it seems we are quite close for this PR to be ready to be merged but please address my last larger suggestions first. After that you will still need to adjust the formatting using `uncrustify` before it can be merged. If you don't, I will format it myself.

src/printer_tree.c
```diff
@@ -4337,3 +4418,131 @@ tree_print_parsed_submodule(struct ly_out *out, const struct lysp_submodule *sub
 
     return erc;
 }
+
+/* @param ext[in] context allocated for extension */
+/* @param refs[out] set of lysc nodes matching parent-refernce xpaths? */
+static LY_ERR
+trb_parent_references(const struct lysc_ext_instance *ext, struct ly_set **refs)
```
michalvasko, line 4425: Looking at this again, please put this function directly into `schema_mount.c` and make it publicly available similarly to `lyplg_ext_schema_mount_create_context()`, you can call it `lyplg_ext_schema_mount_get_parent_ref()`. Also, it is already implemented there, in function `schema_mount_dup_parent_ref()` so it would be best if you could refactor it to not duplicate the code.

src/printer_tree.c
```diff
@@ -4337,3 +4418,131 @@ tree_print_parsed_submodule(struct ly_out *out, const struct lysp_submodule *sub
...
+out:
+    if (ext_data_free) {
+        lyd_free_all(ext_data);
+    }
+    ly_set_free(pref_set, NULL);
+    return res;
+}
+
+static LY_ERR
+trb_print_mount_point(struct trt_node *node, struct trt_wrapper wr, struct trt_printer_ctx *pc,
```
lePici, line 4484: To check whether the parsed module prints correctly (only the `mp` symbol is checked there), I prepared a unit-test in `/tests/utests/schema/test_printer_tree.c`, which you can add to PR.
[test_printer_tree_edited.txt](https://github.com/CESNET/libyang/files/9615944/test_printer_tree_edited.txt)

src/printer_tree.c
```diff
@@ -4337,3 +4418,131 @@ tree_print_parsed_submodule(struct ly_out *out, const struct lysp_submodule *sub
...
+        }
+
+        *newtc = tmptc;
+        ly_set_add(modules, (void *)newtc, 1, NULL);
+    }
+
+    for (uint32_t v = 0; v < modules->count; ++v) {
+        newtc = modules->objs[v];
+        tmpwr = (v == modules->count-1) ? wr : trp_wrapper_set_mark_top(wr);
+        trb_print_family_tree(tmpwr, pc, newtc);
```
lePici, line 4539: As I mentioned in a previous comment, the entire subtree should not be printed. The  `trb_print_parents` would be maybe better to use but it will need to be modified a bit. Unfortunately, such a function is missing for a module that is only parsed and not compiled and must be added.

tools/lint/examples/sm-main.yang
```diff
@@ -0,0 +1,32 @@
+module sm-main {
```
lePici, line 1: When running the command:
`../../../build/yanglint -f tree sm-main.yang`
I am getting an error:
`libyang err : Extension plugin "libyang 2 - Schema Mount, version 1": Failed to get extension data, no callback set.`

src/printer_tree.c
ekinzie: OK, I will move that to schema_mount.c and refactor so that the xpath evaluation is shared, but schema_mount_dup_parent_ref() duplicates data nodes and there is no data context when printing the tree.  There may not be a huge savings resulting from the refactoring.

michalvasko: Hm, you are right. But please move the function there anyway, I would say that file is more suitable because it is schema-mount functionality that is not specific to the tree printer and could theoretically be used for something else in the future.

src/printer_tree.c
```diff
@@ -4337,3 +4418,131 @@ tree_print_parsed_submodule(struct ly_out *out, const struct lysp_submodule *sub
...
+     * modules and modules with no nodes (e.g., iana-if-types).
+     */
+    uint32_t v = ly_ctx_internal_modules_count(ext_ctx);
+    while ((mod = ly_ctx_get_module_iter(ext_ctx, &v))) {
+        uint32_t siblings;
+
+        if ((mod->ctx->flags & LY_CTX_SET_PRIV_PARSED) && mod->compiled)
+            trm_lysc_tree_ctx(mod, pc->out, pc->max_line_length, pc, &tmptc);
+        else
+            trm_lysp_tree_ctx(mod, pc->out, pc->max_line_length, pc, &tmptc);
```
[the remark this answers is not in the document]
ekinzie: Displaying the "mp" flag is no problem and there is enough information to figure out when `/` and `@` should be displayed when working with the parsed module, but displaying the mounted tree itself might require substantial work.  The schema_mount plugin API is dependent on `struct lysc_ext_instance`, presumably because it has module and plugin data pointers which the lysp_* version of that structure is missing.  The extension callback function (ly_ext_data_clb) also requires the lysc variant.

Maybe I have overlooked something.  What are your thoughts?

src/printer_tree.c
```diff
@@ -4337,3 +4418,131 @@ tree_print_parsed_submodule(struct ly_out *out, const struct lysp_submodule *sub
...
+
+        term = (struct lyd_node_term *)pref_set->dnodes[i];
+        LYD_VALUE_GET(&term->value, xp_val);
+        LY_CHECK_GOTO(res = lyplg_type_print_xpath10_value(xp_val, LY_VALUE_JSON, NULL, &value, &err), out);
+        LY_CHECK_ERR_GOTO(res = lys_find_xpath(ext->module->ctx, NULL, value, 0, &snode_set), free(value), out);
+        free(value);
+        for (uint32_t sn = 0; sn < snode_set->count; sn++) {
+            struct lysc_node *top = snode_set->snodes[sn];
+
+            while (top->parent)
```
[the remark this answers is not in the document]
ekinzie: And in a case like this:
```
<parent-reference>/if:interfaces/if:interface/if:name</parent-reference>
<parent-reference>/if:interfaces/if:interface/ip:ipv4/ip:address/ip:ip</parent-reference>
```
you would expect a single trimmed tree containing both name and ip?  That's probably too simple of an example, but I think we could have some awkward diagrams.  The tree diagram RFC and the network instance RFC (8529) both show these parent reference top-level nodes with collapsed trees ("...").   Maybe it is safer to mimic this behavior since there are examples of it?

src/context.c
```diff
@@ -389,7 +389,8 @@ ly_ctx_new_yl_legacy(struct ly_ctx *ctx, const struct lyd_node *yltree)
 }
 
 LIBYANG_API_DEF LY_ERR
-ly_ctx_new_ylpath(const char *search_dir, const char *path, LYD_FORMAT format, int options, struct ly_ctx **ctx)
+ly_ctx_new_ylpath_ext(const char *search_dir, const char *path, LYD_FORMAT format, int options, struct ly_ctx **ctx,
```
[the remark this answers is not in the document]
ekinzie: Maybe you can help clarify this.  Are you suggesting a way to get rid of the new `ly_ctx_new_yl*()` variants, or just an alteration to how the `ctx` argument is used?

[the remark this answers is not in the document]
ekinzie: I removed the new _ext() variants and adjusted the behavior of ly_ctx_new_yldata() based on whether *ctx is NULL.  This works, but one thing I'll point out is that it is a little bit fragile since existing applications have previously had no reason to initialize ctx to NULL

src/printer_tree.c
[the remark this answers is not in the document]
lePici: Yes the schema_mount plugin API needs a compiled module. If there is a module that is only parsed, I would just print the `mp` flag for now and not do any additional compilation. So if the mounted module is only parsed, its mount points are not resolved (only the `mp` flags are printed).
I will try to summarize the situation:
1. The compiled module (lysc) has a mount point:
1.1 The mounted module is compiled and it can be printed.
1.2 The mounted module is only parsed but it can be printed (_not currently implemented in PR_ and we agreed that `mp`, `@`, `/` symbols can be printed.).
2. If the parsed module (lysp) has a mount point:
2.1 The mounted module cannot be printed.

src/context.c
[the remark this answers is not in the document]
michalvasko: Yes, you are right but I do not think this function was used that much and it is worth it to not clutter the API.

src/printer_tree.c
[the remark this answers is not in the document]
michalvasko: > show these parent reference top-level nodes with collapsed trees ("...")

I think `...` was meant to just be a simplification so the example does not have to have the whole schema trees printed, they are supposed to be there. In any case, printing the whole schema trees is misleading because they are not actually accessible so it does not make sense to me.

> you would expect a single trimmed tree containing both name and ip?

Exactly but when we thought about it, "merging" the paths is a non-trivial problem that we wanted to solve later ourselves although you are free to give it a shot yourself. For now we would be fine with printing the path of each parent reference separately.

tools/lint/examples/sm-main.yang
ekinzie: This is fixed, now.

src/printer_tree.c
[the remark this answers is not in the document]
ekinzie: For now each parent reference is printed separately.

src/printer_tree.c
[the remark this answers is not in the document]
ekinzie: I think I have accounted for all of these.  Scenario 1.2 requires some manual testing because, as it is, the schema mount plugin will always create a context for the mounted modules so that they are compiled.  I tried this out with a temporary change
```
--- a/src/printer_tree.c
+++ b/src/printer_tree.c
@@ -4469,6 +4469,9 @@ trb_print_mount_point(struct trt_node *node, struct trt_wrapper wr, struct trt_p
         return LY_SUCCESS;
     }
 
+    if ((err = ly_ctx_new(".", 0, &ext_ctx)))
+        return err;
+
     if (lyplg_ext_schema_mount_create_context(node->mount, &ext_ctx)) {
         /* Void mount point */
         return LY_SUCCESS;
```
to force the plugin to start with an existing (empty) context.

tools/lint/main_ni.c
```diff
@@ -191,6 +200,10 @@ help(int shortout)
     printf("  -s SUBMODULE, --submodule=SUBMODULE\n"
             "                Print the specific submodule instead of the main module.\n\n");
 
+    printf("  -S CONTEXT\n"
+            "                Name of a file containing an XML schema context to be passed to\n"
+            "                the schema-mount extension.\n\n");
+
```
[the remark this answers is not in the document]
ekinzie: I updated the help text.

tools/lint/main_ni.c
```diff
@@ -645,6 +689,10 @@ fill_context(int argc, char *argv[], struct context *c)
             c->submodule = optarg;
             break;
 
+        case 'x': /* --ext-context */
```
michalvasko, line 692: s/ext-context/ext-data

src/context.c
```diff
@@ -453,7 +453,11 @@ ly_ctx_new_yldata(const char *search_dir, const struct lyd_node *tree, int optio
     LY_CHECK_ARG_RET(NULL, tree, ctx, LY_EINVAL);
 
     /* create a new context */
-    LY_CHECK_GOTO(ret = ly_ctx_new(search_dir, options, &ctx_new), cleanup);
+    if (*ctx == NULL) {
```
michalvasko, line 456: API function functionality changed, please update its documentation accordingly.

michalvasko (changes requested):
Alright, thanks for this larger update and I have only a few last comments, otherwise I think it is ready.

src/context.c
ekinzie: I also updated the documentation for ly_ctx_new_ylmem and ly_ctx_new_ylpath, since these call ly_ctx_new_yldata internally and are affected by the change.

michalvasko (approved):
Okay, seems fine to me, thanks. @lePici will also take a look and then we can merge it.

src/printer_tree.c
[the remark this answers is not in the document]
lePici: When I tried this temporary change, I didn't see any result. Nevermind. You can test scenario 1.2 by adding to `sm-context-extension.xml` something like:
```
     <import-only-module>
       <name>sm-mod</name>
       <revision>2017-01-26</revision>
       <namespace>urn:yanglint:sm-mod</namespace>
     </import-only-module>
```
Also add `import sm-mod {...` to the `sm-extension` module. And also of course you need to add a module `sm-mod.yang` with some nodes. The parsed module print should be visible using your command:
`../../../build/yanglint -f tree -p ../../../models -p . -Y ./sm-context-main.xml -x ./sm-context-extension.xml sm-main.yang`
When I tried such an example, the `printer_tree` failed. Please check it out.

src/printer_tree.c
lePici: Ups sorry, I did not call `make format` when I sent this file. Then the `uncrustify` tool will take care of that. But you don't have to deal with that.

lePici:
I just have a [suggestion](https://github.com/CESNET/libyang/pull/1901#discussion_r986830023) on how to modify your example. Please try to fix the error that appears when printing the parsed module.

src/printer_tree.c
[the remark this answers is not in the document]
ekinzie: I can reproduce the error and will take a look at it.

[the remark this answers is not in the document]
ekinzie: I just pushed an update to address this.

lePici:
I just have a [suggestion](https://github.com/CESNET/libyang/pull/1901#discussion_r986830023) on how to modify your example. Please try to fix the error that appears when printing the parsed module.

michalvasko:
Thank you for the contribution, we will finish the last minor problems ourselves.

PR #1902, opened 2022-08-22
tests: skip `expect` on Apple Mac OS

jktjkt:
***Note:** I'm basing this on top of #1900 to prevent a simple conflict in `tools/lint/CMakeLists.txt`.*

When I run these tests via CI, they appear to hang, and the debugging output (via `exp_internal 1`) indicates that there are some raw escape sequences related to terminal width querying which are for some reason pattern-matched against the expected output. That feels wrong, and there's nothing like that on Linux (but `isatty()` reports the same thing on both platforms):
```
expect: does "YANGLINT[W]: Configuration directory "/Users/runner/.yanglint" does not exist, creating it.\r\nYANGLINT[W]: No saved history.\r\n\u001b[6n" (spawn_id exp6) match glob pattern "> "? no
expect: timed out
send: sending "list\r" to { exp6 }
```
I don't feel like debugging a readline reimplementation for a platform that I have no access to. Also, the `linenoise` library fork that we're using has been modified before, and it seems that simply syncing with upstream won't be trivial anymore.

Given that libyang itself has decided to skip these tests on its own CI (see 682d543541d401209e40c45504b16f7ec231cfb1, 8d6e65ff8b256115d495f1174b22d480c748beb6 and dca634647b97ac1dd28c2879a4d270420c3e16d4), let's solve this for everybody and make it clear that the interactive thingy is totally untested on OS X.

Fixes: 682d54354 (actions UPDATE skip expect tests on mac)


michalvasko:
> Also, the linenoise library fork that we're using has been modified before, and it seems that simply syncing with upstream won't be trivial anymore.

I would have no problem resolving some conflicts and/or merging some changes manually if I knew it will help but I have no reason to think so.

Also, please fix the conflicts with your previous PR.

jktjkt:
> Also, please fix the conflicts with your previous PR.

Oops, sorry, done.

michalvasko (commented):
Actually, I am wondering whether to merge this. I have disabled these tests in the CI because they do not work there but they passed fine when I have tried it on a MacOS 11 myself.

jktjkt:
I cannot debug the difference between GitHub's action runner and an actual OS X environment since I don't have access to the HW. Is there perhaps a different version of `expect`, or some funny env flag, or a different `$TERM`, perhaps?

michalvasko:
I have no idea :( But things like this are exactly why I hate developing on Mac and so I would prefer not to spend more time on this than necessary. It seems to me it is fine the way it is, the tests disabled only in the CI.

PR #1915, opened 2022-08-30
xpath BUGFIX overwrite ly_set only after comparison is done

Haititi:
Fixes CESNET#1913

### DIFF
--- a/src/xpath.c
+++ b/src/xpath.c
@@ -9253,7 +9253,8 @@ eval_equality_expr(const struct lyxp_expr *exp, uint32_t *tok_idx, uint32_t repe
             set_scnode_clear_ctx(set, LYXP_SET_SCNODE_ATOM_VAL);
         } else {
             ly_bool result;
-            rc = moveto_op_comp(set, &set2, &exp->expr[exp->tok_pos[this_op]], &result);
+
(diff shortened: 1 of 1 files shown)


michalvasko:
Alright, the fix seems correct and I have no problems with it except failing format check because you need to leave a space after variable declarations.

src/xpath.c
```diff
@@ -9260,8 +9252,11 @@ eval_equality_expr(const struct lyxp_expr *exp, uint32_t *tok_idx, uint32_t repe
             lyxp_set_scnode_merge(set, &set2);
             set_scnode_clear_ctx(set, LYXP_SET_SCNODE_ATOM_VAL);
         } else {
-            rc = moveto_op_comp(set, &set2, &exp->expr[exp->tok_pos[this_op]]);
+            ly_bool result;
+
+	    rc = moveto_op_comp(set, &set2, &exp->expr[exp->tok_pos[this_op]], &result);
```
michalvasko, line 9257: Had to run `uncrustify` myself, you have a TAB here.

michalvasko (changes requested):

src/xpath.c
Haititi: Sorry, couldn't install `uncrustify` on my system due to the minimum required version being to high.

michalvasko (approved):

PR #1928, opened 2022-10-18
Printer tree and mount-point

lePici:
This pull request partly refactored PR #1901.
Printer tree and mount-point tests have also been added.
The biggest change concerns the printing of the shared-schema tree based on parent-references. Individual paths to parent-reference nodes are not printed, but instead are combined into a single tree if they have a common parents.

### DIFF
--- a/src/plugins_exts/schema_mount.c
+++ b/src/plugins_exts/schema_mount.c
@@ -690,24 +690,24 @@ cleanup:
 LY_ERR
 lyplg_ext_schema_mount_get_parent_ref(const struct lysc_ext_instance *ext, struct ly_set **refs)
 {
-    LY_ERR res;
+    LY_ERR rc;
     struct ly_set *pref_set = NULL;
-    struct ly_set *snode_set;
+    struct ly_set *snode_set = NULL;
     struct ly_set *results_set = NULL;
     struct lyd_node *ext_data;
     ly_bool ext_data_free;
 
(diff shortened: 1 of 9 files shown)


src/plugins_exts/schema_mount.c
```diff
@@ -692,7 +692,7 @@ lyplg_ext_schema_mount_get_parent_ref(const struct lysc_ext_instance *ext, struc
 {
     LY_ERR res;
```
michalvasko, line 693: We use `ret` or `rc`, we definitely do not need another variable `res` for this purpose, enough that we have to. I write `rc` in all new code.

src/plugins_exts/schema_mount.c
```diff
@@ -717,29 +717,29 @@ lyplg_ext_schema_mount_get_parent_ref(const struct lysc_ext_instance *ext, struc
...
     if (ext_data_free) {
         lyd_free_all(ext_data);
     }
     ly_set_free(pref_set, NULL);
     return res;
+
+error:
+    ly_set_free(results_set, NULL);
+    goto out;
 }
```
michalvasko, line 743: Merge into `cleanup` and start it with `if (res) { ly_set_free(results_set, NULL); }`.

michalvasko (changes requested):

michalvasko (approved):

PR #1933, opened 2022-10-26
resolve Segmentation fault when get extensions plugin

jc1995wade:
If ext_plugins[u].revision is NULL, revision is not NULL, resulting in a Segmentation fault


michalvasko (approved):

michalvasko:
Okay, no problem, but you really should stop using this deprecated *libyang* version.

jc1995wade:
ok

PR #1966, opened 2023-01-06
Debian packaging update

rfc1036:
I have uploaded these changes to the Debian archive.


rfc1036:
This is slight embarrassing, but I have been working on the packaging for the old 1.x branch instead of the current one...
Merging this will not hurt, but it's not going to end up in the Debian archive. Up to you. :-)

michalvasko:
I can merged if you care about it, but only if there are no extra commits of mine, just yours.

rfc1036:
I do not think that other commits are needed.

michalvasko:
That is my point, but they are included in your PR. You have not properly synchronized with `debian/master`.

debian/control
```diff
@@ -8,7 +8,7 @@ Priority: optional
 Standards-Version: 4.5.0
 Build-Depends: bison,
                cmake,
-               debhelper (>= 9.20150101),
+               debhelper-compat (= 13),
```
jruzicka-nic, line 11: This would break build on Debian 10, 11.

jruzicka-nic:
Thanks for the extra effort to get the packaging changes upstream, but I'll take care of that on next `libyang` release.

I'm still waiting for my key to be added to Debian Developer, uploading keyring, but I hope to be able to package next releases in Debian myself :pray: Once I'm able to, I'll keep the upstream packaging in sync with Debian.

Thanks again for updating the Debian package!

debian/control
rfc1036: Debian 11 has debhelper 13, and for Debian 10 for this kind of package you can safely downgrade the compat level to 12. Compat level 9 is terribly old and should not be used for any reason at this point.

Please do not forget to enable the tests in the next upload.

jruzicka-nic: In Knot DNS and Knot Resolver, I just depend on `debhelper` without version specification which should result in using the latest `debhelper` package for target system, no? Why specify the particular `debhelper` version for upstream packaging which targets multiple releases including supported Ubuntu LTS, currently 18.04 has `debhelper` 11 for example. I've never hit any problems with upstream packages without the `debhelper` version spec.

Which tests do you mean? I thought the tests are already enabled during compilation, I see
```
Build-depends: libcmocka-dev <!nocheck>
```
which is required for tests and also [this](https://salsa.debian.org/netconf-team/libyang2/-/blob/debian/master/debian/rules#L13)
```
		-DENABLE_TESTS=ON
```

There are even `autopkgtests`. We have some basic autopkgtests in upstream `distro/tests` as well, I'll try syncing those as well in the future.

rfc1036: debhelper requires specifying a compatibility level either in the deprecated `debian/compat` file or with a build-dependency on `debhelper-compat`, you cannot avoid this. The package had "10" in `debian/compat`, which is still way too hold.

You should use a different git branch for each Debian/Ubuntu release that you are targeting, or else the results will be sub optimal.

`-DENABLE_TESTS=ON` was not enabled in the package I worked on.

PR #1967, opened 2023-01-06
Updated Debian packaging for libyang2

rfc1036:
I have updated the Debian packaging for libyang2 and uploaded it to the Debian archive.

Cc @oerdnj


michalvasko:
Thanks for the effort but we plan to update and release the DEB package ourselves soon. So I would keep this as a reference for @jruzicka-nic.

rfc1036:
Well. This is what is now in the Debian archive, so it is released. Do you plan to maintain your own private repository?

jruzicka-nic:
@rfc1036 Oh you've updated `libyang2` in the Debian archive - awesome, thanks!

Upstream Debian packaging is now maintained in `distro/pkg/deb` with the help of [apkg](https://apkg.readthedocs.io/en/latest/) so you'd need to update the files there.

I planned to take care of Debian <-> upstream sync as soon as my Debian Developer process finishes.

I also plan to maintain libyang2/libnetconf2/sysrepo/netopeer2 Debian packages, current maintainer Ondřej has already agreed to that.

rfc1036:
I am not familiar with this apkg, but:
- it does not appear to be packaged for Debian
- the Debian packaging of official Debian package must be in the /debian/ directory of the branch from which the package is built from

This means that if you want to maintain an upstream debianization for libyang2 in the upstream repository then the official Debian packaging will have to be maintained in some other repository (which is actually how Debian packages are usually maintained).

jruzicka-nic:
Right, `apkg` is upstream CI oriented and not stable/packaged (yet) so I'd prefer to maintain official Debian packaging the standard way in Debian Salsa with `gbp` as I do with all other packages - Salsa CI is nice too.

There's a [netconf-team](https://salsa.debian.org/netconf-team) on Salsa which contains (out of date) sources for depending `libnetconf2`, `sysrepo`, and `netopeer2` packages and I think `libyang2` belongs there as well.

However, my request to join `netconf-team` resulted in no response (I pinged ondrej directly as well) so I don't know howto proceed there other than creating a duplicate `sysrepo-team` or such.

```
Maintainer: libyang2 packagers <libyang2@packages.debian.org>
Homepage: https://github.com/CESNET/libyang
Vcs-Browser: https://salsa.debian.org/netconf-team/libyang2
Vcs-Git: https://salsa.debian.org/netconf-team/libyang2.git
```

What do you think, @rfc1036 ?

oerdnj:
> I pinged ondrej directly as well

`EAGAIN`

I'm trying this new thing called INBOX 20k :)

> 9520 unread emails
> 11637 unread emails

I've added you to the group...

jruzicka-nic:
> > I pinged ondrej directly as well
> 
> `EAGAIN`
> 
> I'm trying this new thing called INBOX 20k :)

:joy: 

> > 9520 unread emails
> > 11637 unread emails

That's seriously scary 🙈

> I've added you to the group...

Yes, I was able to create libyang2 repo now, thanks!

I see @rfc1036 added to the team as well so feel free to init the repo with your codes and then I can take over :)

Cheers!

oerdnj:
> That's seriously scary 🙈

Yeah, email is more UDP-like than TCP-like for me. If you don't get any response, retry. I am usually not ignoring people on purpose.

rfc1036:
I have uploaded 2.1.4-2 with the endianity fix and pushed it to salsa, but I forgot to update the Vcs-* fields.

Please note that unless regressions will pop up before the release I do not expect to do further work on this package.

Issue #1983, opened 2023-02-17
rpm packaging fails: In-source build is not allowed.


harmic:
I am building libyang RPMs on a Rocky Linux 8 machine.

The build fails with the following error:

```
CMake Error at CMakeLists.txt:9 (message):
  In-source build is not allowed.  Please make a standalone build directory
  and run CMake from there.  You may need to remove CMakeCache.txt.
```

I was able to work around it by editing the `distro/pkg/rpm/libyang.spec` like this:

```diff
--- a/distro/pkg/rpm/libyang.spec
+++ b/distro/pkg/rpm/libyang.spec
@@ -1,3 +1,4 @@
+%undefine __cmake_in_source_build
 Name: libyang
 Version: {{ version }}
 Release: {{ release }}%{?dist}
@@ -49,13 +50,13 @@ written (and providing API) in C.
 %cmake_build

 %if "x%{?suse_version}" == "x"
-cd redhat-linux-build
+cd %{__cmake_builddir}
 %endif
 make doc

 %check
 %if "x%{?suse_version}" == "x"
-cd redhat-linux-build
+cd %{__cmake_builddir}
 %endif
 ctest --output-on-failure -V %{?_smp_mflags}
```

The default for `__cmake_in_source_build` seems to be `1` but the CMakeLists file does not like that.
Also the build dir is no longer 'redhat-linux-build` but something like `x86_64-redhat-linux-gnu`.

michalvasko:
Please use [apkg](https://github.com/CESNET/libyang/tree/master/distro) for building the packages, everything works fine with it.

harmic:
I *did* use apkg.

jruzicka-nic:
Hey @harmic, thanks for the patch, it seems to work in my local testing :partying_face: 

I wasn't willing to debug and workaround broken `%cmake` macros on EL 8, as they're working in later distros (EL >= 9, all current Fedora and SUSE 15.4) but this seems to do the trick!

I've [added](https://gitlab.nic.cz/packaging/sysrepo-pkgtest/-/merge_requests/5) Rocky 8 and AlmaLinux 9 jobs to sysrepo packaging CI, so this should stay functional.

I'm also glad `apkg` worked for you, yay \o/

harmic:
Excellent @jruzicka-nic thanks for the quick turnaround!

PR #1985, opened 2023-02-17
rpm: CMake issues workaround for Rocky 8

jruzicka-nic:
Fixes: https://github.com/CESNET/libyang/issues/1983


jruzicka-nic:
I've tested this on rocky-8, alma-9, fedora-36, fedora-37 and suse-15, I think it should be safe to add and with latest AlmaLinux 9 finally working, we have a working packaging support for EL :partying_face:

PR #1999, opened 2023-03-27
Replaced pthread_key_create() by a memory pointer because the number …

tieliao:
…of pthread keys is limited by PTHREAD_KEY_MAX which is 1024. As such, the number of ly_ctx is also limited to 1024. This value can only be changed by re-building libc. This change works for my case but I'm not sure that it is safe or not for general use.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -327,7 +327,7 @@ struct ly_ctx {
 
     ly_ext_data_clb ext_clb;          /**< optional callback for providing extension-specific run-time data for extensions */
     void *ext_clb_data;               /**< optional private data for ::ly_ctx.ext_clb */
-    pthread_key_t errlist_key;        /**< key for the thread-specific list of errors related to the context */
+    struct ly_err_item *errlist;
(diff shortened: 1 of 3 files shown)


michalvasko:
> This change works for my case but I'm not sure that it is safe or not for general use.

Well, your solution is the first simplest idea anyone would have implementing this so you seem not to have understood why it is implemented differently.

There is a use-case that would not work with your change is actually the reason it is implemented the way it is. You can having a single context and use it in 2 threads simultaneously. There are many thread-safe functions using the context but accessing its error needs to be handled specially for it to work correctly. So, we will not merge your PR as it is removing features.

There are complex solutions possible that would not limit the number of contexts but it does not seems they are worth it as you are the first to encounter the problem and only in a test scenario, basically.

michalvasko:
> I'd argue that using a pthread key does not change anything

It allows reading 2 different errors from 2 different threads in case the same context is used, which is what @jktjkt tried to illustrate.

Issue #544, opened 2018-07-16
deref() function usage in path statement


nani82096:
I have an use case where I need to use deref function inside the path statement.
I encountered the following error (while loading it to sysrepo using the sysrepoctl)
(srctl_ly_log_cb:500) libyang: Unexpected character(s) 'd' (deref(../interf).

I read RFC and libyang code as well it says deref supported only inside must and when statements. Would be happy if I can use deref inside the path statement as well.

michalvasko:
Hi,
which `path` statement do you have in mind? But if you read in the RFC that it is not supported where you would like it to be then there is nothing we can do.

Regards,
Michal

pitekantropus:
Hi,

I have the same problem as the author of this topic. However, I couldn't find any information in RFC 7950 that deref function is not supported in path statement. If in fact it is not forbidden, is it possible that you implement it in libyang?

Reference: `https://tools.ietf.org/html/rfc7950#section-10.3.1`

michalvasko:
Hi,
unfortunately it really is not supported, just look at [path, second paragraph](https://tools.ietf.org/html/rfc7950#section-9.9.2) and then the [path-arg rule](https://tools.ietf.org/html/rfc7950#page-205), which is exact about what is allowed and what not.

Regards,
Michal

pitekantropus:
Hi,
thank you for quick answer. However, it's still not crystal clear for me, especially that function current() is allowed in path, it's in example in RFC. The only clue that may show deref restriction is this sentence "Predicates are used only for constraining the values for the key nodes for list entries".
Could you confirm or point me the proper explanation?

Regards,
Amadeusz

michalvasko:
Hi,
the last sentence of that second paragraph says `The syntax is formally defined by the rule "path-arg"`. I then referenced the place where you can find `path-arg` defined as a token in the YANG ABNF grammar. That is as an exact definition as it gets. And yes, `current()` function is allowed (actually mandatory in the predicate) as the only XPath function.

Regards,
Michal

pitekantropus:
Hi,
Ok, now I see that. Thank you again for clear explanation.

PR #2030, opened 2023-05-19
Feature/xpath in leafref

steweg:
This feature introduces allowing usage of XPath functions within leafref paths. It also fixes minor compilation errors

### DIFF
--- a/src/context.h
+++ b/src/context.h
@@ -197,7 +197,7 @@ struct ly_ctx;
 #define LY_CTX_ENABLE_IMP_FEATURES 0x0100 /**< By default, all features of newly implemented imported modules of
                                         a module that is being loaded are disabled. With this flag they all become
                                         enabled. */
-#define LY_CTX_ENABLE_XPATH_IN_LEAFREF 0x0200 /**< By default, path attribute of leafref accepts only path as defined in RFC 7950.
+#define LY_CTX_LEAFREF_EXTENDED 0x0200 /**< By default, path attribute of leafref accepts only path as defined in RFC 7950.
                                         By using this option, the path attribute will also allow using XPath functions as deref() */
 
 /** @} contextoptions */
--- a/src/context.h
+++ b/src/context.h
@@ -197,6 +197,8 @@ struct ly_ctx;
 #define LY_CTX_ENABLE_IMP_FEATURES 0x0100 /**< By default, all features of newly implemented imported modules of
                                         a module that is being loaded are disabled. With this flag they all become
                                         enabled. */
+#define LY_CTX_ENABLE_XPATH_IN_LEAFREF 0x0200 /**< By default, path attribute of leafref accepts only path as defined in RFC 7950.
+                                        By using this option, the path attribute will also allow using XPath functions as deref() */
 
 /** @} contextoptions */
 
--- a/src/path.c
+++ b/src/path.c
@@ -274,54 +274,48 @@ token_error:
  * @return LY_ERR value.
  */
 static LY_ERR
-ly_path_parse_deref(const struct ly_ctx *ctx, const struct lysc_node *ctx_node, const struct lyxp_expr *exp, uint32_t *tok_idx)
+ly_path_parse_deref(const struct ly_ctx *ctx, const struct lysc_node *ctx_node, const struct lyxp_expr *exp,
+        uint32_t *tok_idx)
 {
-    LY_ERR ret = LY_SUCCESS;
     size_t arg_len;
     uint32_t begin_token, end_token;
     struct lyxp_expr *arg_expr = NULL;
 
     /* mandatory FunctionName */
-    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_FUNCNAME), ret = LY_EVALID, error);
-    if (exp->used <= *tok_idx + 1) {
-        if (ctx) {
-            LOGVAL(ctx, LY_VCODE_XP_EOF);
-        }
-        return LY_EINCOMPLETE;
-    }
-
+    LY_CHECK_RET(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_FUNCNAME), LY_EVALID);
     if (!strncmp(&exp->expr[exp->tok_pos[*tok_idx]], "deref", 5)) {
-        LOGVAL(ctx, LYVE_XPATH, "Unexpected XPath function \"%.*s\" in path, expected \"deref(...)\"", exp->tok_len[*tok_idx], exp->tok_pos[*tok_idx]);
-        ret = LY_EVALID;
-        goto error;
+        LOGVAL(ctx, LYVE_XPATH, "Unexpected XPath function \"%.*s\" in path, expected \"deref(...)\"",
+                exp->tok_len[*tok_idx], exp->tok_pos[*tok_idx]);
+	return LY_EVALID;
     }
 
     /* mandatory '(' */
-    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_PAR1), ret = LY_EVALID, error);
+    LY_CHECK_RET(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_PAR1), LY_EVALID);
     begin_token = *tok_idx;
 
     /* count tokens till ')' */
-    while (lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_PAR2) && lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_FUNCNAME)) {
+    while (lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_PAR2) && *tok_idx < exp->used) {
+        /* emebedded functions are not allowed */
+        if (!lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_FUNCNAME)) {
+            LOGVAL(ctx, LYVE_XPATH, "Embedded function XPath function inside deref function within the path"
+                    "is not allowed");
+	    return LY_EVALID;
+        }
+
         (*tok_idx)++;
     }
 
-    /* embedded function inside function ')' */
-    if (!lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_FUNCNAME)) {
-        LOGVAL(ctx, LYVE_XPATH, "Embedded function XPath function inside deref function within the path is not allowed");
-        ret = LY_EVALID;
-        goto error;
-    }
-
     /* mandatory ')' */
-    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_PAR2), ret = LY_EVALID, error);
+    LY_CHECK_RET(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_PAR2), LY_EVALID);
     end_token = *tok_idx - 1;
 
-    /* verify the path of deref argument */
+    /* parse the path of deref argument */
     arg_len = exp->tok_pos[end_token] - exp->tok_pos[begin_token];
-    LY_CHECK_ERR_GOTO(ly_path_parse(ctx, ctx_node, &exp->expr[exp->tok_pos[begin_token]], arg_len, 1, LY_PATH_BEGIN_EITHER, LY_PATH_PREFIX_OPTIONAL, LY_PATH_PRED_LEAFREF, &arg_expr), ret = LY_EVALID, error);
+    LY_CHECK_RET(ly_path_parse(ctx, ctx_node, &exp->expr[exp->tok_pos[begin_token]], arg_len, 1,
+            LY_PATH_BEGIN_EITHER, LY_PATH_PREFIX_OPTIONAL, LY_PATH_PRED_LEAFREF, &arg_expr), LY_EVALID);
     lyxp_expr_free(ctx, arg_expr);
-error:
-    return ret;
+
+    return LY_SUCCESS;
 }
 
 LY_ERR
@@ -355,11 +349,13 @@ ly_path_parse(const struct ly_ctx *ctx, const struct lysc_node *ctx_node, const 
             /* relative path check specific to leafref */
             if (lref) {
                 /* optional function 'deref..' */
-                if ((ly_ctx_get_options(ctx) & LY_CTX_LEAFREF_EXTENDED) && !lyxp_check_token(NULL, exp, tok_idx, LYXP_TOKEN_FUNCNAME)) {
+                if ((ly_ctx_get_options(ctx) & LY_CTX_LEAFREF_EXTENDED) &&
+                        !lyxp_check_token(NULL, exp, tok_idx, LYXP_TOKEN_FUNCNAME)) {
                     LY_CHECK_ERR_GOTO(ly_path_parse_deref(ctx, ctx_node, exp, &tok_idx), ret = LY_EVALID, error);
 
                     /* '/' */
-                    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, &tok_idx, LYXP_TOKEN_OPER_PATH), ret = LY_EVALID, error);
+                    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, &tok_idx, LYXP_TOKEN_OPER_PATH), ret = LY_EVALID,
+                            error);
                 }
 
                 /* mandatory '..' */
@@ -367,7 +363,8 @@ ly_path_parse(const struct ly_ctx *ctx, const struct lysc_node *ctx_node, const 
 
                 do {
                     /* '/' */
-                    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, &tok_idx, LYXP_TOKEN_OPER_PATH), ret = LY_EVALID, error);
+                    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, &tok_idx, LYXP_TOKEN_OPER_PATH), ret = LY_EVALID,
+                            error);
 
                     /* optional '..' */
                 } while (!lyxp_next_token(NULL, exp, &tok_idx, LYXP_TOKEN_DDOT));
@@ -934,7 +931,52 @@ cleanup:
 }
 
 /**
- * @brief Appeds path elemets from source to destination array
+ * @brief Duplicate ly_path_predicate structure.
+ *
+ * @param[in] ctx libyang context.
+ * @param[in] pred The array of path predicates.
+ * @param[out] dup Duplicated predicates.
+ * @return LY_ERR value.
+ */
+LY_ERR
+ly_path_dup_predicates(const struct ly_ctx *ctx, const struct ly_path_predicate *pred, struct ly_path_predicate **dup)
+{
+    LY_ARRAY_COUNT_TYPE u;
+
+    if (!pred) {
+        return LY_SUCCESS;
+    }
+
+    LY_ARRAY_CREATE_RET(ctx, *dup, LY_ARRAY_COUNT(pred), LY_EMEM);
+    LY_ARRAY_FOR(pred, u) {
+        LY_ARRAY_INCREMENT(*dup);
+        (*dup)[u].type = pred->type;
+
+        switch (pred->type) {
+        case LY_PATH_PREDTYPE_POSITION:
+            /* position-predicate */
+            (*dup)[u].position = pred->position;
+            break;
+        case LY_PATH_PREDTYPE_LIST:
+        case LY_PATH_PREDTYPE_LEAFLIST:
+            /* key-predicate or leaf-list-predicate */
+            (*dup)[u].key = pred->key;
+            pred->value.realtype->plugin->duplicate(ctx, &pred->value, &(*dup)[u].value);
+            LY_ATOMIC_INC_BARRIER(((struct lysc_type *)pred->value.realtype)->refcount);
+            break;
+        case LY_PATH_PREDTYPE_LIST_VAR:
+            /* key-predicate with a variable */
+            (*dup)[u].key = pred->key;
+            (*dup)[u].variable = strdup(pred->variable);
+            break;
+        }
+    }
+
+    return LY_SUCCESS;
+}
+
+/**
+ * @brief Appends path elements from source to destination array
  *
  * @param[in] ctx libyang context.
  * @param[in] src The source path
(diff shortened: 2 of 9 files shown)


steweg:
I have looked little bit deeper and found the following:
https://www.rfc-editor.org/errata/eid5617

On top of that I have searched for other major implementations and the following supports it:
- pyang (https://www.rfc-editor.org/errata/eid5617)
- Cisco compiler (https://www.rfc-editor.org/errata/eid5617)
- OpenDaylight (https://www.javadoc.io/static/org.opendaylight.yangtools/yangtools-docs/5.0.6/org/opendaylight/yangtools/yang/model/api/PathExpression.DerefSteps.html)

Also other people were asking for this (https://github.com/CESNET/libyang/issues/544), so it is not only me but apparently wider audience. It seems that it is becoming pretty much standard.

Anyway up to you, if you wish to go on with this feature or not. If yes, please review the code. If not, then I will cancel this PR, and keep the patch within my own fork so if anyone would like to use it, then can clone it and use it from there. One-way or another, please let me know. Thanks

jktjkt:
Since this is a non-standard extension which is clearly something else than what the YANG 1.0 and 1.1 standards define today -- if this ever lands, please only activate it upon a feature flag/option/whatever. There's value in tools which implement the standard as-is, without custom extensions -- if only for compliance checking.

(I have no horse in this race. It\s just that I would like to check my models for standard conformance.)

steweg:
Good point. I have updated the code accordingly

src/path.c
```diff
@@ -249,6 +249,40 @@ ly_path_check_predicate(const struct ly_ctx *ctx, const struct lysc_node *cur_no
...
+    /* mandatory '(' */
+    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_PAR1), ret = LY_EVALID, error);
+
+    /* mandatory 'Expr' */
+    while (lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_PAR2) && lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_FUNCNAME)){
+         (*tok_idx)++;
+    }
+
+    /* embedded function inside function ')' */
+    if (!lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_FUNCNAME)) {
```
michalvasko, line 276: A more robust skip would be preferred such as counting the parentheses, for example.

michalvasko (changes requested):
Fine, I suppose I am willing to merge this if you solve all the issues. Except for the comments, change the base branch to `devel`, we do not accept PRs into `master`. Also, run the code through `uncrustify`, I am pretty sure there are formatting errors.

src/path.c
steweg: This depends how much you want to restrict this feature. Either you want to allow just "deref(relative-path-arg)" or generic XPath. Current check is validating just generic XPath, by relaying on generic xpath parser and validating just what is needed to get to the last part of relative path section. I am not sure how to make it better as there can be many complex scenarios of nested structures etc...

Do you have any exact proposal?

src/context.h
```diff
@@ -196,6 +196,8 @@ struct ly_ctx;
 #define LY_CTX_ENABLE_IMP_FEATURES 0x0100 /**< By default, all features of newly implemented imported modules of
                                         a module that is being loaded are disabled. With this flag they all become
                                         enabled. */
+#define LY_CTX_ENABLE_XPATH_IN_LEAFREF 0x0200 /**< By default, path attribute of leafref accepts only path as defined in RFC 7950.
```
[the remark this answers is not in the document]
steweg: done

src/path.c
```diff
@@ -275,6 +309,14 @@ ly_path_parse(const struct ly_ctx *ctx, const struct lysc_node *ctx_node, const
         if (lyxp_next_token(NULL, exp, &tok_idx, LYXP_TOKEN_OPER_PATH)) {
             /* relative path check specific to leafref */
             if (lref) {
+                /* optional function 'deref..' */
+                if (ctx && (ly_ctx_get_options(ctx) & LY_CTX_ENABLE_XPATH_IN_LEAFREF) && !lyxp_check_token(NULL, exp, tok_idx, LYXP_TOKEN_FUNCNAME)) {
```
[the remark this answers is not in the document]
steweg: done

src/path.c
michalvasko: Right, the argument is supposed to again be a leafref relative/absolute path, which is already being parsed so there are all the required functions for parsing it. And yes, you should also check just for `deref()` as only that function should be allowed.

tools/lint/main_ni.c
```diff
@@ -287,6 +287,9 @@ help(int shortout)
             "                create an exact YANG schema context. If specified, the '-F'\n"
             "                parameter (enabled features) is ignored.\n\n");
 
+    printf("  -X, --xpath-in-leafref\n"
+            "                Allow usage of XPath within leafref\n\n");
+
```
michalvasko, line 292: As I said previously, these names you are using are really inaccurate. Leafref is already a subset of XPath. What this option does is enable `deref()` in leafref, nothing else.

src/path.c
```diff
@@ -842,6 +884,75 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
...
+ * but also if the relative path starting in @p ctx_node reaches the document root via double dots.
+ * @param[in] expr Parsed path.
+ * @param[in] format Format of the path.
+ * @param[in] prefix_data Format-specific data for resolving any prefixes (see ::ly_resolve_prefix).
+ * @param[in] getnext_opts Options to be used for ::lys_getnext() calls.
+ * @param[out] path Compiled path.
+ * @return LY_ERR value.
+ */
+static LY_ERR
+_ly_path_compile_function(const struct ly_ctx *ctx, const struct lys_module *cur_mod, const struct lysc_node *ctx_node,
```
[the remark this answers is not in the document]
michalvasko: I do not understand the problem, you should only need `path.c` functions. Why are you even parsing `deref()` argument as XPath when it is just a relative/absolute path. And that can be parsed/compiled with path functions so you have everything you need.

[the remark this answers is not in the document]
steweg: Done

src/path.c
steweg: Ok, done

tools/lint/main_ni.c
steweg: Done

src/path.c
```diff
@@ -865,6 +933,118 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
...
+    strncpy(&new_path[pos], lref->path->expr, len);
+    pos += len;
+
+    /* add '[' */
+    new_path[pos] = '[';
+    ++pos;
+
+    /* add key predicate */
+    u = lref->path->used - 1;
+    strncpy(&new_path[pos], &lref->path->expr[lref->path->tok_pos[u]], lref->path->tok_len[u]);
```
michalvasko, line 1005: I am sorry, I guess I was not clear. What I suggested is to **compile** a leafref with deref by storing the actual dereferenced leafref nodes but I would keep the path expression as it is, with deref.

src/path.c
```diff
@@ -264,6 +264,66 @@ ly_path_check_predicate(const struct ly_ctx *ctx, const struct lysc_node *cur_no
...
+    }
+
+    /* mandatory '(' */
+    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_PAR1), ret = LY_EVALID, error);
+    begin_token = *tok_idx;
+
+    /* count tokens till ')' */
+    while (lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_PAR2) && lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_FUNCNAME)) {
+        (*tok_idx)++;
+    }
```
michalvasko, line 306: What we talked about in some other conversation was parsing the argument as a nested path because that is what is allowed there so I would expect to see `lyd_parse_path()` here. The compile function expects a valid path but this does not check anything, pretty much.

tools/lint/main_ni.c
```diff
@@ -287,6 +287,9 @@ help(int shortout)
             "                create an exact YANG schema context. If specified, the '-F'\n"
             "                parameter (enabled features) is ignored.\n\n");
 
+    printf("  -X, --enable-deref-within-leafref\n"
```
michalvasko, line 290: Again, sorry if I was not clear but this now seems too specific to me. To be future proof, I would reuse the flag name, something like `--extended-leafref`.

michalvasko (changes requested):
These may be the last needed changes. After that I will actually try to run and examine the code.

src/path.c
steweg: This is being already done, see lines 319-322. In your highlighted part I am just checking where the deref-arg path ends and that we don't have some nested XPath function within, therefore the closing ')' must be the one related to our deref-arg.

tools/lint/main_ni.c
steweg: Done

src/path.c
steweg: This whole exercise to temporarily create unwinded path is to get the "full path" to the output and not just the last part (as it was before). I didn't find any other way to do it.

To store this dereferenced leafref is also big question. I also thought about this before, but it seems to be useless (there is no real use for it right now). I am not even sure if you mean to store deref node itself or deref node leafref target? I assumed the second one. I also assume to store it as part of first output path element. If I assume wrongly, just let me know how exactly you wish to store this and I will adjust it.

src/path.c
michalvasko: Okay, will check properly after you address the other comment. Like I said, I have not executed the code yet so I may have read the code wrong.

src/path.c
steweg: Ok, that seems to be strange request. Lets consider this extended example:
```
list l1 {
    key t1;
    leaf t1 {
        type uint8;
    }
    list l2 {
        key t2;
        leaf t2 {
            type uint8;
        }
        list l3 {
            key t3;
            leaf t3 {
                type uint8;
            }
        }
    }
}
leaf r1 {type leafref {path "../l1/t1\";}}
leaf r2 {type leafref {path "deref(../r1)/../l2/t2\";}}
leaf r3 {type leafref {path "../l1[t1 = current()/../r1)/../l2/t2";}}
leaf r4 {type leafref {path "deref(../r2)/../l3/t3";}}
```
Let's compile `r2`. The resulted output path nodes:: `l1`, `l2`, `t2`. `l1` will also have predicate filled in with key set  to `t1`
Let's compile `r3`. The output path nodes: `l1`, `l2`, `t2`. `l1` will also have predicate filled in with key set  to `t1`

If we change it according to your proposal compilation of `r2` and `r3` will not have the same results. Is this really wanted behavior? I think that the output should be equal, shouldn't it?

One additional problem with your proposal is compilation of `r4`. What exact nodes should you expect to have in output path and in which order (I am not 100% sure if order matters or not)? Should it be:
 - `r1`, `l1`, `t1`, `r2`, `l2`, `t2`, `l3`, `t3`
 - `r2`, `l2`, `t2`, `r1`, `l1`, `t1`, `l3`, `t3`
 - `r2`, `r1`, `l1`, `t1`, `l2`, `t2`, `l3`, `t3`

If I compare it to output of `lyxp_atomize()`,  which seems to give similar outputs but in little bit different order:
 - `r4`, `t1`, `r1`, `l1`, `l2`, `t2`, `r2`, `l3`, `t3`

But even using the atoms output, there is still a needs to convert it to `ly_path` structures, which is not very easy way how to do it.

michalvasko: > If we change it according to your proposal compilation of r2 and r3 will not have the same results. Is this really wanted behavior? I think that the output should be equal, shouldn't it?

Based on my proposal the output (the compiled nodes) would be different yes. The reason is that when you evaluate each of the paths, different nodes would be traversed, which is what atoms are supposed to represent. And yes, the obvious observation is that you can write equal expressions in many ways with different efficiency. Part of the reason I do not like this `deref()` extension.

> One additional problem with your proposal is compilation of r4. What exact nodes should you expect to have in output path and in which order (I am not 100% sure if order matters or not)? Should it be:

For atoms, the order should not matter.

> But even using the atoms output, there is still a needs to convert it to ly_path structures, which is not very easy way how to do it.

Just another reason why I was not so happy about supporting this but it was all your idea. But you should not rely on `lyxp_atomize()`, use the existing path compilation which should already do everything, it just needs to be adjusted to be able to parse subexpressions (deref parameter) so you can collect all the paths and join them in a single path.

steweg: Ok, done

src/path.c
```diff
@@ -865,6 +933,116 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
...
+
+    /* prepare expr representing just deref arg */
+    expr2.tokens = &expr->tokens[2];
+    expr2.tok_pos = &expr->tok_pos[2];
+    expr2.tok_len = &expr->tok_len[2];
+    expr2.repeat = &expr->repeat[2];
+    expr2.used = *tok_idx - 2;
+    expr2.size = expr->size - 2;
+    expr2.expr = expr->expr;
```
michalvasko, line 1004: Assumes the whole expression starts with `deref(`, no reason for that especially since the current `tok_idx` is passed as a parameter.

src/path.c
```diff
@@ -865,6 +933,116 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
...
+
+    /* compile just deref arg */
```
michalvasko, line 1006: That is not all that is happening, so please update the comment. The dereferenced node is also found and the compiled path added to the final path.

src/path.c
```diff
@@ -865,6 +933,116 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
...
+    LY_CHECK_ERR_GOTO(ly_path_compile_leafref(ctx, ctx_node, top_ext, &expr2, oper, target, format, prefix_data, &path2), ret = LY_EVALID, cleanup);
+    node2 = path2[LY_ARRAY_COUNT(path2) - 1].node;
+    deref_leaf_node = (const struct lysc_node_leaf *)node2;
+    lref = (const struct lysc_type_leafref *)deref_leaf_node->type;
+    LY_CHECK_ERR_GOTO(ly_path_append(ctx, path2, path), ret = LY_EVALID, cleanup);
+    ly_path_free(ctx, path2);
+    path2 = NULL;
+
+    /* compile deref leafref */
```
michalvasko, line 1015: Again, please add more details.

src/path.c
```diff
@@ -865,6 +933,116 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
     return (ret == LY_ENOTFOUND) ? LY_EVALID : ret;
 }
 
+/**
+ * @brief Appeds path elemets from source to destination array
```
michalvasko, line 937: Typos.

src/path.c
```diff
@@ -865,6 +933,116 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
...
+ly_path_append(const struct ly_ctx *ctx, const struct ly_path *src, struct ly_path **dst)
+{
+    LY_ERR ret = LY_SUCCESS;
+    struct ly_path *p;
+    size_t u;
+
+    LY_ARRAY_FOR(src, u) {
+        LY_ARRAY_NEW_GOTO(ctx, *dst, p, ret, cleanup);
+        p->node = src[u].node;
+        p->ext = src[u].ext;
```
michalvasko, line 954: While most likely not needed for this specific use-case, but it may be required eventually and the function pretends to be generic, so please add predicate duplication.

src/path.c
```diff
@@ -865,6 +933,116 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
...
+    expr2.used = expr->used - *tok_idx;
+    expr2.size = expr->size - *tok_idx;
+    expr2.expr = expr->expr;
+
+    /* compile rest of the path */
+    LY_CHECK_ERR_GOTO(ly_path_compile_leafref(ctx, node2, top_ext, &expr2, oper, target, format, prefix_data, &path2), ret = LY_EVALID, cleanup);
+    LY_CHECK_ERR_GOTO(ly_path_append(ctx, path2, path), ret = LY_EVALID, cleanup);
+cleanup:
+    ly_path_free(ctx, path2);
+    path2 = NULL;
```
michalvasko, line 1038: Redundant line.

src/path.h
```diff
@@ -62,6 +62,7 @@ struct ly_path {
                                        - is inner node - path is relative */
     const struct lysc_ext_instance *ext;    /**< Extension instance of @p node, if any */
     struct ly_path_predicate *predicates;   /**< [Sized array](@ref sizedarrays) of the path segment's predicates */
+    const struct lysc_node *deref_node;     /**< Optional dereferenced node */
```
michalvasko, line 65: Not used anymore.

src/path.c
```diff
@@ -264,6 +264,66 @@ ly_path_check_predicate(const struct ly_ctx *ctx, const struct lysc_node *cur_no
...
+        goto error;
+    }
+
+    /* mandatory ')' */
+    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_PAR2), ret = LY_EVALID, error);
+    end_token = *tok_idx - 1;
+
+    /* verify the path of deref argument */
+    arg_len = exp->tok_pos[end_token] - exp->tok_pos[begin_token];
+    LY_CHECK_ERR_GOTO(ly_path_parse(ctx, ctx_node, &exp->expr[exp->tok_pos[begin_token]], arg_len, 1, LY_PATH_BEGIN_EITHER, LY_PATH_PREFIX_OPTIONAL, LY_PATH_PRED_LEAFREF, &arg_expr), ret = LY_EVALID, error);
```
michalvasko, line 321: Needs a manual line wrap for 120 characters.

src/path.c
```diff
@@ -264,6 +264,66 @@ ly_path_check_predicate(const struct ly_ctx *ctx, const struct lysc_node *cur_no
...
+    while (lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_PAR2) && lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_FUNCNAME)) {
+        (*tok_idx)++;
+    }
+
+    /* embedded function inside function ')' */
+    if (!lyxp_check_token(NULL, exp, *tok_idx, LYXP_TOKEN_FUNCNAME)) {
+        LOGVAL(ctx, LYVE_XPATH, "Embedded function XPath function inside deref function within the path is not allowed");
+        ret = LY_EVALID;
+        goto error;
+    }
```
michalvasko, line 313: I cannot really think of a use-case when this would not suffice but I would strongly prefer just one loop checking for `LYXP_TOKEN_PAR2` while counting any nested parentheses and checking for `LYXP_TOKEN_FUNCNAME` in the same loop body.

src/path.c
```diff
@@ -264,6 +264,66 @@ ly_path_check_predicate(const struct ly_ctx *ctx, const struct lysc_node *cur_no
...
+
+    /* mandatory ')' */
+    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_PAR2), ret = LY_EVALID, error);
+    end_token = *tok_idx - 1;
+
+    /* verify the path of deref argument */
+    arg_len = exp->tok_pos[end_token] - exp->tok_pos[begin_token];
+    LY_CHECK_ERR_GOTO(ly_path_parse(ctx, ctx_node, &exp->expr[exp->tok_pos[begin_token]], arg_len, 1, LY_PATH_BEGIN_EITHER, LY_PATH_PREFIX_OPTIONAL, LY_PATH_PRED_LEAFREF, &arg_expr), ret = LY_EVALID, error);
+    lyxp_expr_free(ctx, arg_expr);
+error:
```
michalvasko, line 323: Just some customs we have, if the label code is always performed, we call it `cleanup` and there should always be an empty line before the label (there are a few instances of this being broken).

src/path.c
```diff
@@ -264,6 +264,66 @@ ly_path_check_predicate(const struct ly_ctx *ctx, const struct lysc_node *cur_no
...
+    struct lyxp_expr *arg_expr = NULL;
+
+    /* mandatory FunctionName */
+    LY_CHECK_ERR_GOTO(lyxp_next_token(ctx, exp, tok_idx, LYXP_TOKEN_FUNCNAME), ret = LY_EVALID, error);
+    if (exp->used <= *tok_idx + 1) {
+        if (ctx) {
+            LOGVAL(ctx, LY_VCODE_XP_EOF);
+        }
+        return LY_EINCOMPLETE;
+    }
```
michalvasko, line 291: Seems this is a condition for calling this function so you can put this into an `assert`.

michalvasko (changes requested):
Thanks and I have executed the code now so there is some finishing touch needed but it seems to work fine, generally.

src/path.c
steweg: Done

src/path.c
steweg: Done

src/path.h
steweg: Done

src/path.c
steweg: Done

src/path.c
steweg: Done

src/path.c
steweg: Done

src/path.c
steweg: Done

src/path.c
steweg: Done

src/path.c
steweg: Done

src/path.c
```diff
@@ -865,6 +930,188 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
...
 
+/**
+ * @brief Duplicate ly_path_predicate structure.
+ *
+ * @param[in] ctx libyang context.
+ * @param[in] pred The array of path predicates.
+ * @param[out] dup Duplicated predicates.
+ * @return LY_ERR value.
+ */
+LY_ERR
```
michalvasko, line 941: Should be `static` I assume.

src/path.c
```diff
@@ -865,6 +930,188 @@ ly_path_compile_predicate_leafref(const struct lysc_node *ctx_node, const struct
...
+
+    /* compile rest of the path and append it to the path */
+    LY_CHECK_ERR_GOTO(ly_path_compile_leafref(ctx, node2, top_ext, &expr2, oper, target, format, prefix_data, &path2),
+            ret = LY_EVALID, cleanup);
+    LY_CHECK_ERR_GOTO(ly_path_append(ctx, path2, path), ret = LY_EVALID, cleanup);
+
+cleanup:
+    ly_path_free(ctx, path2);
+    if (ret) {
+        *path = NULL;
```
michalvasko, line 1109: Wouldn't this just leak `path`? I think it should be freed first, but only in the `if`.

src/path.c
```diff
@@ -1168,34 +1422,8 @@ ly_path_dup(const struct ly_ctx *ctx, const struct ly_path *path, struct ly_path
...
-                case LY_PATH_PREDTYPE_LIST_VAR:
-                    /* key-predicate with a variable */
-                    (*dup)[u].predicates[v].key = pred->key;
-                    (*dup)[u].predicates[v].variable = strdup(pred->variable);
-                    break;
-                }
-            }
-        }
+        (*dup)[u].ext = path[u].ext;
+        LY_CHECK_RET(ret = ly_path_dup_predicates(ctx, path[u].predicates, &(*dup)[u].predicates), ret);
```
michalvasko, line 1426: +1 for noticing this and not duplicating the code.

michalvasko (changes requested):
Some last comments.

src/path.c
steweg: Done

src/path.c
steweg: Done

michalvasko (approved):

michalvasko:
Fine, thanks but regarding the compilation warnings, they were all false positives and they had nothing to do with the core issue. I do not care that much so I merged it all but they should have been a separate PR, if anything.

PR #2038, opened 2023-06-02
yanglint tcl tests

lePici:
Added tcl tests that test options for yanglint. Parameters from non-interactive mode to interactive mode were also added, if they were missing and vice versa. Also found some bugs in libyang and fixed them.


CMakeModules/FindPCRE2.cmake
```diff
@@ -22,19 +22,23 @@ else()
...
-        /usr/lib
-        /usr/lib64
-        /usr/local/lib
-        /usr/local/lib64
-        /opt/local/lib
-        /sw/lib
-        ${CMAKE_LIBRARY_PATH}
-        ${CMAKE_INSTALL_PREFIX}/lib)
+    if (WIN32 AND "${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+        find_library(PCRE2_LIBRARY pcre2-8d)
```
michalvasko, line 26: Some explanation please. For debug build the `pcre2` library is called `pcre2-8d` and for release `pcre2-8`?

tools/lint/completion.c
```diff
@@ -75,6 +75,34 @@ get_cmd_completion(const char *hint, char ***matches, unsigned int *match_count)
...
+/**
+ * @brief Provides completion for arguments.
+ *
+ * @param[in] hint User input.
+ * @param[in] args Array of all possible arguments. The last element must be NULL.
+ * @param[out] matches Matches provided to the user as a completion hint.
+ * @param[out] match_count Number of matches.
+ */
+static void
+get_arg_completion(const char *hint, const char **args, char ***matches, unsigned int *match_count)
```
michalvasko, line 87: Nice addition.

CMakeModules/FindPCRE2.cmake
lePici: That's right, I added a comment.

michalvasko: Strange, but okay. But then you could just add the potential name to the list of NAMES and it should work fine. With a comment mentioning the reason and that it affect only Windows.

lePici: yes, it seems to me that it only applies to Windows. That list of NAMES works as sorted, so there was no way to add it there to correctly use the name for the Release and Debug version. At least that's how it appeared to me when I tried it.

michalvasko: I see, so both Release and Debug libraries are available, why is then the Debug one required to be linked? I do not think anyone would need to debug pcre2 usage in libyang so I think always linking the Release version should be fine.

michalvasko: Okay, then I suppose there is no way to improve/avoid this.

jktjkt: @lePici: my MSVC and VCPKG experience is [limited](https://github.com/Telecominfraproject/oopt-gnpy-libyang/blob/main/.github/workflows/ci.yaml), but I think that the difference between the `Release` and `Debug` targets mostly concerns just the project that is being built "right now". In other words, it should be possible to use a debug build of `libyang` while linking to a release `libpcre2`, or the other way round, or any combination of that. But maybe I'm wrong and there are different MSVCRT copies?

michalvasko: Feel free to improve this in another PR if you can, I cannot help as I have not even compiled it on Win.

lePici: @jktjkt: I think you have more Windows experience than I do. I just ran into a problem with the Debug configuration, so this is how I solved it. Maybe I did something wrong during the installation phase and cmake/vcpkg/msvc is confused.

PR #2050, opened 2023-06-16
Fix libyang.spec to support older rpmbuild with RedHat/CentOS 7

### LINKED ISSUE #2049: libyang doesn't build packages on RedHat/CentOS 7
Current (2.1.80 and `devel`) does not support building packages on CentOS 7. The rpm spec file uses newer cmake directives which are not implemented on the older rpmbuild as shipped with CentOS 7

mwinter-osr:
Fixes #2049

### DIFF
--- a/distro/pkg/rpm/libyang.spec
+++ b/distro/pkg/rpm/libyang.spec
@@ -47,27 +47,46 @@ written (and providing API) in C.
 
 %prep
 %autosetup -p1
+%if 0%{?rhel} && 0%{?rhel} < 8
+    mkdir build
+%endif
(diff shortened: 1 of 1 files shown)


mwinter-osr:
Looks like CI fails on Windows on a unrelated system issue

michalvasko (approved):

michalvasko:
Fine, I suppose, but we are generally moving away from supporting such outdated distributions (at least for the other projects such as *libnetconf2* where the code would just be too messy not to mention EOL of the provided OpenSSL version).

PR #2084, opened 2023-08-12
printer BUGFIX fix for ly_realloc() failure in ly_vprint_()

barbarost:
In case ly_realloc() fails in ly_vprint_(), out->method.mem.buf should not be set to NULL but *out->method.mem.buf should be


michalvasko (approved):

michalvasko:
Minor simplification, fine.

PR #2093, opened 2023-08-31
Windows: accept backslashes in file names

jktjkt:
Without this patch, a warning gets printed for any .yang files that are read using native path names:
```
libyang[1]: File name "D:\a\oopt-gnpy-libyang\oopt-gnpy-libyang\tests\yang\ietf-network@2018-02-26.yang" does not match module name "ietf-network".
```

### DIFF
--- a/src/tree_schema_common.c
+++ b/src/tree_schema_common.c
@@ -696,9 +696,8 @@ static LY_ERR
 lysp_load_module_check(const struct ly_ctx *ctx, struct lysp_module *mod, struct lysp_submodule *submod, void *data)
 {
     struct lysp_load_module_check_data *info = data;
-    const char *filename, *dot, *rev, *name;
+    const char *name;
     uint8_t latest_revision;
-    size_t len;
     struct lysp_revision *revs;
 
     name = mod ? mod->mod->name : submod->name;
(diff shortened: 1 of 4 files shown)


jktjkt:
> Paths on Windows can have either `/` or `\` as the separator?

Yes. Users can (and do) invoke this function either with paths which went through some canonicalization code (like `std::filesystem::path` in C++, or `pathlib.Path` on Python, etc), which would typically get you backslashes, but it's also very common to have these paths "hardcoded" to use forward slashes only, which Windows support just fine.

> That is what the code seems to indicate. In any case, I would prefer to not change `compat` and just put something like `FILENAME_SLASH` macro into `config.h.in` that would change based on the platform. Then, use it in a dedicated function for getting the file name from a path.

That would work if that thing was a regular expression, but because it's a simple character search and Windows can have both `/` and `\`, this approach won't work. I don't see any other way than accepting both, and for that a new function is needed.

michalvasko:
I see. What I would do no matter what is to write a dedicated function for getting all the path components, it seems most of the code is duplicated. And why not just put a single `#ifdef` there for optional backslash support?

jktjkt:
> I see. What I would do no matter what is to write a dedicated function for getting all the path components, it seems most of the code is duplicated. And why not just put a single `#ifdef` there for optional backslash support?

So you'd like me to change these three places which perform some checks on the filename. OK, later today.

jktjkt:
(looks like I need more coffee so as not to confuse these two buttons, yay)

src/tree_schema_common.c
```diff
@@ -2613,3 +2590,39 @@ lys_stmt_flags(enum ly_stmt stmt)
 
     return 0;
 }
+
+void ly_sanity_check_module_filename(const struct ly_ctx *ctx, const char *name, const char *revision, const char *filename)
```
michalvasko, line 2594: Uncrustify does not like the return value on the same line as the function declaration. Also, is it really a sanity check? We have no control over how the file is named so I would say it is a "regular" check.

jktjkt: Both done.

jktjkt:
The CI failure message is not helpful. Can it print a diff instead of saying this?
```
Checking format of the sources with /usr/local/bin/uncrustify ...
/home/runner/work/libyang/libyang/uncrustify.cfg:2852: unknown option 'align_nl_cont_spaces'
FAIL: src/tree_schema_common.c (File size changed from 82094 to 82097)
```

michalvasko:
I am guessing you are missing parentheses on line 2604 on the comparison. But printing diff seems like a good idea for an improvement.

src/tree_schema_internal.h
```diff
@@ -732,4 +732,6 @@ uint8_t lys_stmt_flags(enum ly_stmt stmt);
  */
 LY_ERR lyplg_ext_get_storage_p(const struct lysc_ext_instance *ext, int stmt, const void ***storage_p);
 
+void ly_check_module_filename(const struct ly_ctx *ctx, const char *name, const char *revision, const char *filename);
```
michalvasko, line 735: Brief doxygen would be nice.

jktjkt: Done

jktjkt:
> I am guessing you are missing parentheses on line 2604 on the comparison.

Fixed.

michalvasko (approved):

PR #2096, opened 2023-09-07
printer json BUGFIX nested union value print

Ilyesbenhamouda:
Fix an infinite loop in json_print_value() when
resolved type of a union member is also union.

Fixes: 183b911ec920 ("printer json BUGFIX union realtype value print")


michalvasko:
Please switch the target branch to `devel`, we do not accept PRs into `master`. Also, do you have a non-working use-case illustrating the need for this fix? I think you are right but a new test would still be appreciated.

michalvasko:
Thanks, test added.

PR #2101, opened 2023-09-14
Make tree_data static functions available in api

irfanHaslanded:
We need certain functions to be exposed via a public API to make it
 possible to use in other languages such as Rust.

 lyd_parent, lyd_child, and lyd_get_value are defined as static inline
 and is difficult for tools to wrap them.


michalvasko (commented):
I guess I am fine with this but I am not sure making the functions inline has any benefit, I would think the compilers are smart enough to do it themselves as needed. So you may simply change them into normal functions.

irfanHaslanded:
Thanks Michal, I will do that then.

src/tree_data.h
```diff
@@ -1004,15 +1004,8 @@ struct lyd_node_opaq {
...
-static inline struct lyd_node *
-lyd_parent(const struct lyd_node *node)
-{
-    if (!node || !node->parent) {
-        return NULL;
-    }
-
-    return &node->parent->node;
-}
+LIBYANG_API_DECL struct lyd_node *
```
michalvasko, line 1007: Needs to be on a single line, all the declarations.

michalvasko (changes requested):

src/tree_data.h
irfanHaslanded: Done

jktjkt:
> lyd_parent, lyd_child, and lyd_get_value are defined as static inline and is difficult for tools to wrap them.

How come? These functions are defined in a public header, so they should be usable just fine, AFAICT. What is the exact problem that you're solving with this patch, and what error message in which scenario are you getting without this patch?

michalvasko (approved):

irfanHaslanded:
> > lyd_parent, lyd_child, and lyd_get_value are defined as static inline and is difficult for tools to wrap them.
> 
> How come? These functions are defined in a public header, so they should be usable just fine, AFAICT. What is the exact problem that you're solving with this patch, and what error message in which scenario are you getting without this patch?

If we have to use these functions in another language such as Rust, tools such as rust-bindgen, do not easily generate wrappers for them.

jktjkt:
That's been supported since 2022, https://github.com/rust-lang/rust-bindgen/pull/2335

PR #2103, opened 2023-09-15
Add lyd_print_mem_len API to return printed size

irfanHaslanded:
It can be useful to know how many bytes were written, and avoid an expensive and redundant strlen by the application.


michalvasko:
You can write such a wrapper for your application, if you need it, just use `lyd_print_all()`, for example, instead of `lyd_print_()` so unless you have some arguments supporting adding this API, I will close this PR.

irfanHaslanded:
I see, similar to `sr_lyd_print_lyb`. Thanks Michal

Issue #2106, opened 2023-09-25
Creation of keyless list elements takes a lot of time


olivier-matz-6wind:
Hello,

This problem was observed in sysrepo, in sr_modinfo_generate_config_change_notif(), where a keyless list is built. When working with large configurations with notification replay enabled, a huge amount of time is spent there.

For reference, this is the output of "perf report":

```
-   79.40%  netopeer2-serve  libyang.so.2.34.1      [.] lyht_insert_with_resize_cb
   - lyht_insert_with_resize_cb
      - 44.34% lyd_insert_hash_add
	 - 44.33% lyd_insert_node
	    - 44.33% _lyd_new_list
		 lyd_new_list
		 sr_modinfo_generate_config_change_notif
		 sr_changes_notify_store
		 sr_apply_changes
		 np2srv_rpc_editconfig_cb
		 sr_shmsub_rpc_listen_call_callback
		 sr_shmsub_rpc_listen_process_rpc_events
		 sr_subscription_process_events
		 sr_shmsub_listen_thread
		 start_thread
		 __clone3 (inlined)
      - 35.00% lyht_resize
	 - lyht_insert_with_resize_cb
	    - 34.98% lyd_insert_hash_add
	       - lyd_insert_node
		  - 34.98% _lyd_new_list
		       lyd_new_list
		       sr_modinfo_generate_config_change_notif
		       sr_changes_notify_store
		       sr_apply_changes
		       np2srv_rpc_editconfig_cb
		       sr_shmsub_rpc_listen_call_callback
		       sr_shmsub_rpc_listen_process_rpc_events
		       sr_subscription_process_events
		       sr_shmsub_listen_thread
		       start_thread
		       __clone3 (inlined)
```

I reproduced this behavior with a simple test program:

```
#include <stddef.h>
#include <stdio.h>
#include <time.h>
#include <stdint.h>
#include <errno.h>

#include <libyang/libyang.h>

const char *keyless_list_yang = "module keyless-list {\n"
				"  yang-version 1.1;\n"
				"  namespace \"http://example.com/keyless-list\";\n"
				"  prefix keyless-list;\n"
				"\n"
				"  container root {\n"
				"    config false;\n"
				"    list list {\n"
				"      leaf value {\n"
				"        type uint32;\n"
				"      }\n"
				"    }\n"
				"  }\n"
				"}\n";

const char *key_list_yang = "module key-list {\n"
			   "  yang-version 1.1;\n"
[...]

PR #2107, opened 2023-09-25
Optimize hash table insertion

olivier-matz-6wind:
Hello,

This is a draft pull request that solves the issue #2106 

The idea is to allow to insert elements in the hash table in O(1), even in case of collisions. First, the hash table is reworked to avoid browsing records to find a free one. Then, a new API is added to allow an insertion without checking for duplicates.

Feedback is of course welcome!

Note: the first patch breaks the validation test, but it is fixed by the second. From what I understand, it is expected that the browsing order is the same than the insert order, but I think this was not always guaranteed by the previous implementation.

### DIFF
--- a/src/hash_table.c
+++ b/src/hash_table.c
@@ -73,8 +73,10 @@ lyht_init_hlists_and_records(struct ly_ht *ht)
 
     ht->hlists = malloc(sizeof(ht->hlists[0]) * ht->size);
     LY_CHECK_ERR_RET(!ht->hlists, free(ht->recs); LOGMEM(NULL), LY_EMEM);
-    for (i = 0; i < ht->size; i++)
-        ht->hlists[i] = LYHT_NO_RECORD;
+    for (i = 0; i < ht->size; i++) {
+        ht->hlists[i].first = LYHT_NO_RECORD;
+        ht->hlists[i].last = LYHT_NO_RECORD;
+    }
     ht->first_free_rec = 0;
 
     return LY_SUCCESS;
@@ -181,7 +183,7 @@ static LY_ERR
 lyht_resize(struct ly_ht *ht, int operation)
 {
     struct ly_ht_rec *rec;
-    uint32_t *old_hlists;
+    struct ly_ht_hlist *old_hlists;
     unsigned char *old_recs;
     uint32_t old_first_free_rec;
     uint32_t i, old_size;
@@ -213,7 +215,7 @@ lyht_resize(struct ly_ht *ht, int operation)
 
     /* add all the old records into the new records array */
     for (i = 0; i < old_size; i++) {
-        for (rec_idx = old_hlists[i], rec = lyht_get_rec(old_recs, ht->rec_size, rec_idx);
+        for (rec_idx = old_hlists[i].first, rec = lyht_get_rec(old_recs, ht->rec_size, rec_idx);
              rec_idx != LYHT_NO_RECORD;
              rec_idx = rec->next, rec = lyht_get_rec(old_recs, ht->rec_size, rec_idx)) {
             LY_ERR ret;
@@ -346,7 +348,7 @@ __lyht_insert_with_resize_cb(struct ly_ht *ht, void *val_p, uint32_t hash, lyht_
 {
     uint32_t hlist_idx = hash & (ht->size - 1);
     LY_ERR r, ret = LY_SUCCESS;
-    struct ly_ht_rec *rec;
+    struct ly_ht_rec *rec, *prev_rec;
     lyht_value_equal_cb old_val_equal = NULL;
     uint32_t rec_idx;
 
@@ -361,8 +363,15 @@ __lyht_insert_with_resize_cb(struct ly_ht *ht, void *val_p, uint32_t hash, lyht_
     assert(rec_idx < ht->size);
     rec = lyht_get_rec(ht->recs, ht->rec_size, rec_idx);
     ht->first_free_rec = rec->next;
-    rec->next = ht->hlists[hlist_idx];
-    ht->hlists[hlist_idx] = rec_idx;
+
+    if (ht->hlists[hlist_idx].first == LYHT_NO_RECORD) {
+        ht->hlists[hlist_idx].first = rec_idx;
+    } else {
+        prev_rec = lyht_get_rec(ht->recs, ht->rec_size, ht->hlists[hlist_idx].last);
+        prev_rec->next = rec_idx;
+    }
+    rec->next = LYHT_NO_RECORD;
+    ht->hlists[hlist_idx].last = rec_idx;
 
     rec->hash = hash;
     memcpy(&rec->val, val_p, ht->rec_size - SIZEOF_LY_HT_REC);
@@ -432,10 +441,14 @@ lyht_remove_with_resize_cb(struct ly_ht *ht, void *val_p, uint32_t hash, lyht_va
     }
 
     if (prev_rec_idx == LYHT_NO_RECORD) {
-        ht->hlists[hlist_idx] = rec->next;
+        ht->hlists[hlist_idx].first = rec->next;
+        if (rec->next == LYHT_NO_RECORD)
+            ht->hlists[hlist_idx].last = LYHT_NO_RECORD;
     } else {
         prev_rec = lyht_get_rec(ht->recs, ht->rec_size, prev_rec_idx);
         prev_rec->next = rec->next;
+        if (rec->next == LYHT_NO_RECORD)
+            ht->hlists[hlist_idx].last = prev_rec_idx;
     }
 
     rec->next = ht->first_free_rec;
--- a/src/hash_table.c
+++ b/src/hash_table.c
@@ -180,7 +180,7 @@ lyht_free(struct ly_ht *ht, void (*val_free)(void *val_p))
  * @return LY_ERR value.
  */
 static LY_ERR
-lyht_resize(struct ly_ht *ht, int operation)
+lyht_resize(struct ly_ht *ht, int operation, int check)
 {
     struct ly_ht_rec *rec;
     struct ly_ht_hlist *old_hlists;
@@ -220,7 +220,10 @@ lyht_resize(struct ly_ht *ht, int operation)
              rec_idx = rec->next, rec = lyht_get_rec(old_recs, ht->rec_size, rec_idx)) {
             LY_ERR ret;
 
-            ret = lyht_insert(ht, rec->val, rec->hash, NULL);
+            if (check)
+                ret = lyht_insert(ht, rec->val, rec->hash, NULL);
+            else
+                ret = lyht_insert_no_check(ht, rec->val, rec->hash, NULL);
 
             assert(!ret);
             (void)ret;
@@ -344,7 +347,7 @@ lyht_find_next(struct ly_ht *ht, void *val_p, uint32_t hash, void **match_p)
 
 static LY_ERR
 __lyht_insert_with_resize_cb(struct ly_ht *ht, void *val_p, uint32_t hash, lyht_value_equal_cb resize_val_equal,
-        void **match_p)
+        void **match_p, int check)
 {
     uint32_t hlist_idx = hash & (ht->size - 1);
     LY_ERR r, ret = LY_SUCCESS;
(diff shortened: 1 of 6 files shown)


olivier-matz-6wind:
I noticed the test failures, I'll fix them if we move forward with this draft.

src/hash_table.h
```diff
@@ -171,6 +171,19 @@ LIBYANG_API_DECL LY_ERR lyht_find_next_with_collision_cb(struct ly_ht *ht, void
  */
 LIBYANG_API_DECL LY_ERR lyht_insert(struct ly_ht *ht, void *val_p, uint32_t hash, void **match_p);
 
+/**
+ * @brief Insert a value into a hash table, without checking for collision.
```
michalvasko, line 175: If I understand correctly, this inserts a value without checking whether the value has already been inserted, not without checking for hash collisions. That is a huge difference.

src/hash_table_internal.h
```diff
@@ -46,32 +43,67 @@
  */
 struct ly_ht_rec {
     uint32_t hash;        /* hash of the value */
-    int32_t hits;         /* collision/overflow value count - 1 (a filled entry has 1 hit,
-                           * special value -1 means a deleted record) */
-    unsigned char val[1]; /* arbitrary-size value */
-} _PACKED;
+    uint32_t next;        /* index of next collision */
+    unsigned char val[]; /* arbitrary-size value */
```
michalvasko, line 47: When writing this, I researched it quite extensively but I do not remember the details. Bu I think that, strictly according to the standard, this is not a valid declaration and some compilers may have a problem with it.

src/hash_table.c
```diff
@@ -622,23 +492,11 @@ lyht_remove(struct ly_ht *ht, void *val_p, uint32_t hash)
...
+    unsigned int bits = sizeof(uint32_t) * 8;
 
-    /* ... and by using it to shift 1 to the left we get the closest sufficient hash table size */
-    size = 1 << i;
+    if (item_count == 1)
+        return 1;
 
-    return size;
+    /* return next power of 2 (greater or equal) */
+    return 1 << (bits - __builtin_clz(item_count - 1));
```
michalvasko, line 501: You can optimize this, I have never attempted to, but using compiler built-ins is just asking for compatibility problems so please avoid it.

src/hash_table_internal.h
```diff
@@ -46,32 +43,67 @@
  */
 struct ly_ht_rec {
     uint32_t hash;        /* hash of the value */
-    int32_t hits;         /* collision/overflow value count - 1 (a filled entry has 1 hit,
-                           * special value -1 means a deleted record) */
-    unsigned char val[1]; /* arbitrary-size value */
-} _PACKED;
```
michalvasko, line 52: What is the problem with having the records packed? You have encountered some alignment issues? Because it can save a lot of memory.

src/hash_table_internal.h
```diff
@@ -46,32 +43,67 @@
...
  * @brief (Very) generic hash table.
  *
- * Hash table with open addressing collision resolution and
- * linear probing of interval 1 (next free record is used).
- * Removal is lazy (removed records are only marked), but
- * if possible, they are fully emptied.
+ * The hash table is composed of a table of indexes that references the
+ * first record. The records contain a next index that references the
+ * next record in case of collision.
+ * The free records are chained starting from first_free_rec.
```
michalvasko, line 61: The original description was meant to formally characterize the hash table implementation so that it can be understood and changed easier. It may be my fault, but I understand almost nothing from this, seems very generic. It is fine if you do not use any common algorithm names but then I would expect a brief description of the functionality instead.

michalvasko (changes requested):
Thanks for the effort, in general it seems fine but I have made some comments. I am also not completely sure I understand what exact inefficiencies you solved so please correct me if I have it wrong.
1. Skip the check that a value is already in the hash table for all data manipulation, is completely redundant (interestingly, it seems that way and I could not think of a use-case where it could lead to problems).
2. Avoid linear search for an empty record in case the hash-based record is taken. Implemented using a "bucket" for each record that holds the information about the first/last filled record for the hash.

Would be great if something like this (if I am right) is mentioned in the hash table description. Also, from what I remember, buckets were used as an overflow space for collisions to not fill additional hash table records in some hash table collision algorithm so I am not sure the best name was chosen for these structures. Lastly, I believe the size of hash tables has increased by `size * 8 B`, which is not ideal but should probably be fine.

olivier-matz-6wind:
Thanks @michalvasko for the comments.

> Thanks for the effort, in general it seems fine but I have made some comments. I am also not completely sure I understand what exact inefficiencies you solved so please correct me if I have it wrong.
> 
>     1. Skip the check that a value is already in the hash table for all data manipulation, is completely redundant (interestingly, it seems that way and I could not think of a use-case where it could lead to problems).

Yes, this is achieved by the new API lyht_insert_no_check().

>     2. Avoid linear search for an empty record in case the hash-based record is taken. Implemented using a "bucket" for each record that holds the information about the first/last filled record for the hash.

Yes.

> Would be great if something like this (if I am right) is mentioned in the hash table description.

OK, I'll update the hash table description with more details.

> Also, from what I remember, buckets were used as an overflow space for collisions to not fill additional hash table records in some hash table collision algorithm so I am not sure the best name was chosen for these structures.

I can rename it as "hlist" if you feel it's more relevant.

> Lastly, I believe the size of hash tables has increased by `size * 8 B`, which is not ideal but should probably be fine.

Yes. It is +(size * 4) after the first commit and +(size * 8) after the second commit.

While there, do you have an idea why the second commit would be needed to fix the validation unit test?

src/hash_table.h
olivier-matz-6wind: Agree, I'll fix the description.

src/hash_table_internal.h
olivier-matz-6wind: This syntax is conform to C99. See "6.7.2.1 Structure and union specifiers" in https://www.open-std.org/JTC1/SC22/WG14/www/docs/n1256.pdf

I know that val[0] is indeed not portable (gcc-ism), but I never had issues with val[]. However I have no experience outside Unix.

Also, I wonder if val[1] couldn't be interpreted by the compiler as a fixed array, which could trigger an out-of-bound access.

src/hash_table.c
olivier-matz-6wind: I didn't think that it was not available under Windows. I'll fallback to something more standard.

src/hash_table_internal.h
olivier-matz-6wind: I did not encounter any alignment issue.

I think it was correct before, but changing the size of ly_ht_rec makes it incorrect. Let me try to explain the idea.

To me, the packed attribute imply 2 things:
1/ no padding is added between fields. In this case, we only have u32, so there is no padding even without the attribute.
2/ the compiler will generate a (slower) code to access the structure fields that will work whatever the alignment of the structure. This will impact access to ly_ht_rec structure, but not structure located right after (which have probably no packed attribute).

Before my patches, sizeof(struct ly_ht_rec) is 8, so we don't really change the alignment constraint of the structure located right after. So, the accesses to the record always match the default alignment, and accesses to struct ly_ht_rec can be anywhere thanks to the packed attribute.

But the 2nd patch of my patchset changes the size of struct ly_ht_rec from 8 to 12. If alignment constraint of the record is 8, it could trigger unaligned accesses on some architectures.

So the idea was to ensure that size(ly_ht_rec) + record_size is always aligned.

I think the best solution would be to understand how to remove the 2nd patch without breaking the tests, to keep the packed attribute as before, and to add a compile-time check to ensure that size(struct ly_ht_rec) == 8.

What do you think?

src/hash_table_internal.h
olivier-matz-6wind: You are right, I'll try to enhance the description.

src/hash_table_internal.h
michalvasko: I think it was recommended as the safe choice that should always work. Also, what is and out-of-bound access? Never heard about such checks in C, it should not care. I am not strictly against it but `[1]` worked for quite a lot of time and no issues were reported whatsoever so I do not think there is any reason to change it.

src/hash_table_internal.h
michalvasko: To me, it mainly implies that the records will not waste (possibly a lot of) space, for huge hash tables. But it is true that it is used only for aligned values (usually pointers) so it makes no difference. However, I seem to remember that the actual reason I have added that attribute was to fix endianity problems. Because if the value is padded to be larger, assigning it may use the padded bytes and when reading it back, they may be ignored, something along these lines. But again the same argument, it worked until now, why change it?

michalvasko:
> I can rename it as "hlist" if you feel it's more relevant.

Sure, just to avoid people assuming they are buckets from the other algorithm.

> While there, do you have an idea why the second commit would be needed to fix the validation unit test?

Sorry, I am working on something else and would prefer not to leave it in the middle of things unless something else that is critical appears. Just do what you can and incorporate the feedback I give you, I will then take a look later if you will not be able to solve some issues.

src/hash_table_internal.h
olivier-matz-6wind: > But again the same argument, it worked until now, why change it?

The patchset introduces a change: before, if a record had an alignment constraint of 8, it was ensured that the start of the record structure was aligned to 8. If I keep the packed attribute as is, the record will be aligned to 4.

It is probably not a problem on modern archs, but I had the feeling that we should avoid this (see https://www.kernel.org/doc/html/next/core-api/unaligned-memory-access.html).

Anyway, I'll remove this change in v2, we can still come back on it if needed.

Thanks

michalvasko: Maybe I am missing something but I do not think it was (is currently) aligned to anything. If a hash table is created with value size 1 B, there will be no alignment, every record would have 9 B (the real size since it is packed) and the records are one after the other in an array. So technically, the alignment is the value size. But you have not changed the size of the record so like I said, it should work just as before.

olivier-matz-6wind: Yes, for a structure of size 1B has an alignment constraint of 1B, if the sum has a size of 9B, it is still ok.

But if we take a structure of which needs to be aligned to 8B:
struct foo {
  uint64_t x;
  char y;
};

Its size is 16B because the compiler adds padding, to keep all elements aligned when doing an array of struct foo.

The sum size(ly_ht_rec) + record_size was 24B before my patches (so, still ok), but it is 28B after, which breaks the alignment constraint.

olivier-matz-6wind: > Its size is 16B because the compiler adds padding, to keep all elements aligned when doing an array of struct foo.

Well, to be precise, this is not always true, it depends on ABI.

michalvasko: Okay, I have never put structures into the hash table but yes, if the structure is not packed, `sizeof` should return its padded size. But we are not doing this internally, like I said, so I would prefer to keep the records packed to save some memory.

olivier-matz-6wind: Got it, will do like this in v2, thanks!

olivier-matz-6wind:
changelog between v1 and v2:

- rename bucket as hlist
- revert to val[1] instead of val[] in record structure
- revert to packed record structure
- better document ly_ht structure (how the hash table works)
- move lyht_insert_no_check() declaration in the proper patch
- fix API documentation of lyht_insert_no_check()
- do not use __builtin_clz() which is not available under windows
- call ly_ht_insert_no_check() on resize
- use lyht_get_fixed_size(u) in patch "optimize creation of children htable"
- fix minor format and cosmetic issues

olivier-matz-6wind:
> call ly_ht_insert_no_check() on resize

Thinking a bit more about it, I will remove this change, and revert to what I had in v1.

There is certainly something I don't understand about the need of specifying a val_eq_cb when resizing. My naive idea was that we don't need to check anything at resize, since the elements were already there. But there is an explicit API to do set the callback at resize, so my idea is probably wrong...

olivier-matz-6wind:
changelog between v2 and v3:
- call ly_ht_insert_no_check() on resize only if the resize is triggered by a ly_ht_insert_no_check(), as it was done in v1

michalvasko (approved):

michalvasko:
Thanks, the ABI problem should be fine.

PR #2119, opened 2023-10-23
log Remove enclosing '#ifndef NDEBUG' from ly_log_dbg_groups

spellingmistake:
Not having the prototype for ly_log_dbg_groups breaks my release builds (frrouting) due to an implicit function declaration. Judging from the implementation of ly_log_dbg_groups, it was intended to be present in release builds and just do nothing. So the removal of the #ifndef shouldn't be problematic.


michalvasko (approved):

michalvasko:
No problem but merged only into `devel`, will be merged into `master` later.

PR #2154, opened 2024-01-15
Support for printing and parsing bare json values for sysrepo-gnmi

irfanHaslanded:
- Support for parsing JSON  bare leaf values

    Add support for parsing bare values for top-level (from the
    parser's point of view) leaf nodes, using a new LYD_OPT_BARETOPLEAF
    flag. The parsing of leaf nodes is too tied in with the allocation of
    a new node and parsing of the node name and namespace to easily
    trigger it from that case, so instead the case is check and handled
    explicitly in the top-level parser loop.

- Support for printing JSON bare values

    Add support for printing top-level leaf nodes as bare values, which is
    triggered through the use of a new LYD_BARETOPLEAF flag.

### DIFF
--- a/src/parser_data.h
+++ b/src/parser_data.h
@@ -172,6 +172,9 @@ struct ly_in;
                                                  when parsing validated data to skip some validation tasks and modify
                                                  some validation behavior (auto-deletion of cases). */
 
+#define LYD_PARSE_BARETOPLEAF 0x2000000     /**< Fragments for top-level leaf values are expected to be bare and not
+                                                 have a top-level object or node name. */
+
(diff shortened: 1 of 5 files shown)


michalvasko (changes requested):
The parser is quite convoluted as it is so I am against adding new flags and special cases into it. Why not write a new function(s) for the functionality you want? It seems quite simple and the functions would then likely be as well.

jktjkt:
Neither the commit message nor the PR description explains why this is needed. How is it related to YANG or any of its serialization formats? What is the use case, i.e., the problem that this PR solves for you? What you're doing and why you need that?

Also, the parsing code has no documentation, and the one added test works with a simple string. IMHO that doesn't really show why this is needed.

PR #2155, opened 2024-01-16
Adding leafref references and API to lyd_node_term

steweg:
This patch introduces ability to store references between lyd_node_terms, when it comes to leafrefs. This can be used by higher layers to prevent removals of lyd_nodes, which are being referenced or to say, from which place the given data node is being referenced from.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -373,6 +373,7 @@ struct ly_ctx {
     void *ext_clb_data;               /**< optional private data for ::ly_ctx.ext_clb */
     struct ly_ht *err_ht;             /**< hash table of thread-specific list of errors related to the context */
     pthread_mutex_t lyb_hash_lock;    /**< lock for storing LYB schema hashes in schema nodes */
+    struct ly_ht *term_nodes_ext_ht;  /**< hash table of term data nodes extension */
 };
 
 /**
--- a/src/common.h
+++ b/src/common.h
@@ -373,7 +373,7 @@ struct ly_ctx {
     void *ext_clb_data;               /**< optional private data for ::ly_ctx.ext_clb */
     struct ly_ht *err_ht;             /**< hash table of thread-specific list of errors related to the context */
     pthread_mutex_t lyb_hash_lock;    /**< lock for storing LYB schema hashes in schema nodes */
-    struct ly_ht *term_nodes_ext_ht;  /**< hash table of term data nodes extension */
+    struct ly_ht *leafref_links_ht;   /**< hash table of leafref links between term data nodes */
 };
 
 /**
--- a/src/context.c
+++ b/src/context.c
@@ -240,7 +240,7 @@ ly_ctx_ht_err_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UN
 }
 
 /**
- * @brief Hash table value-equal callback for comparing term data node extension hash table record.
+ * @brief Hash table value-equal callback for comparing leafref links hash table record.
  */
 static ly_bool
 ly_ctx_ht_leafref_links_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UNUSED(cb_data))
@@ -251,9 +251,9 @@ ly_ctx_ht_leafref_links_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod)
 }
 
 /**
- * @brief Callback for freeing term data node extension hash table values.
+ * @brief Callback for freeing leafref links hash table values.
  *
- * @param[in] val_p Pointer to a pointer to a term data node extension item to free with all the siblings.
+ * @param[in] val_p Pointer to a pointer to a leafref links item to free with all the siblings.
  */
 static void
 ly_ctx_ht_leafref_links_rec_free(void *val_p)
--- a/src/context.c
+++ b/src/context.c
@@ -620,8 +620,6 @@ ly_ctx_set_options(struct ly_ctx *ctx, uint16_t option)
     if (!(ctx->flags & LY_CTX_LEAFREF_LINKING) && (option & LY_CTX_LEAFREF_LINKING)) {
         ctx->term_nodes_ext_ht = lyht_new(1, sizeof(struct lyd_term_nodes_ext_rec), ly_ctx_ht_term_node_ext_equal_cb, NULL, 1);
         LY_CHECK_ERR_RET(!ctx->term_nodes_ext_ht, LOGARG(ctx, option), LY_EMEM);
-    } else if ((ctx->flags & LY_CTX_LEAFREF_LINKING) && !(option & LY_CTX_LEAFREF_LINKING)) {
-        lyht_free(ctx->term_nodes_ext_ht, ly_ctx_ht_term_node_ext_rec_free);
     }
 
     if (!(ctx->flags & LY_CTX_SET_PRIV_PARSED) && (option & LY_CTX_SET_PRIV_PARSED)) {
@@ -663,6 +661,11 @@ ly_ctx_unset_options(struct ly_ctx *ctx, uint16_t option)
 
     LY_CHECK_ARG_RET(ctx, ctx, LY_EINVAL);
     LY_CHECK_ERR_RET(option & LY_CTX_NO_YANGLIBRARY, LOGARG(ctx, option), LY_EINVAL);
+
+    if ((ctx->flags & LY_CTX_LEAFREF_LINKING) && (option & LY_CTX_LEAFREF_LINKING)) {
+        lyht_free(ctx->term_nodes_ext_ht, ly_ctx_ht_term_node_ext_rec_free);
+	ctx->term_nodes_ext_ht = NULL;
+    }
 
     if ((ctx->flags & LY_CTX_SET_PRIV_PARSED) && (option & LY_CTX_SET_PRIV_PARSED)) {
         struct lys_module *mod;
--- a/src/context.c
+++ b/src/context.c
@@ -239,6 +239,30 @@ ly_ctx_ht_err_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UN
     return !memcmp(&err1->tid, &err2->tid, sizeof err1->tid);
 }
 
+/**
+ * @brief Hash table value-equal callback for comparing term data node extension hash table record.
+ */
+static ly_bool
+ly_ctx_ht_term_node_ext_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UNUSED(cb_data))
+{
+    struct lyd_term_nodes_ext_rec *rec1 = val1_p, *rec2 = val2_p;
+
+    return rec1->node == rec2->node;
+}
+
+/**
+ * @brief Callback for freeing term data node extension hash table values.
+ *
+ * @param[in] val_p Pointer to a pointer to a term data node extension item to free with all the siblings.
+ */
+static void
+ly_ctx_ht_term_node_ext_rec_free(void *val_p)
+{
+    struct lyd_term_nodes_ext_rec *rec = val_p;
+
+    lyd_free_term_node_ext_rec(rec);
+}
+
 LIBYANG_API_DEF LY_ERR
 ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
 {
@@ -261,6 +285,11 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
 
     /* plugins */
     LY_CHECK_ERR_GOTO(lyplg_init(), LOGINT(NULL); rc = LY_EINT, cleanup);
+
+    if (options & LY_CTX_LEAFREF_LINKING) {
+        ctx->term_nodes_ext_ht = lyht_new(1, sizeof(struct lyd_term_nodes_ext_rec), ly_ctx_ht_term_node_ext_equal_cb, NULL, 1);
+        LY_CHECK_ERR_GOTO(!ctx->term_nodes_ext_ht, rc = LY_EMEM, cleanup);
+    }
 
     /* initialize thread-specific error hash table */
     ctx->err_ht = lyht_new(1, sizeof(struct ly_ctx_err_rec), ly_ctx_ht_err_equal_cb, NULL, 1);
@@ -587,6 +616,13 @@ ly_ctx_set_options(struct ly_ctx *ctx, uint16_t option)
     LY_CHECK_ARG_RET(ctx, ctx, LY_EINVAL);
     LY_CHECK_ERR_RET((option & LY_CTX_NO_YANGLIBRARY) && !(ctx->flags & LY_CTX_NO_YANGLIBRARY),
             LOGARG(ctx, option), LY_EINVAL);
+
+    if (!(ctx->flags & LY_CTX_LEAFREF_LINKING) && (option & LY_CTX_LEAFREF_LINKING)) {
+        ctx->term_nodes_ext_ht = lyht_new(1, sizeof(struct lyd_term_nodes_ext_rec), ly_ctx_ht_term_node_ext_equal_cb, NULL, 1);
+        LY_CHECK_ERR_RET(!ctx->term_nodes_ext_ht, LOGARG(ctx, option), LY_EMEM);
+    } else if ((ctx->flags & LY_CTX_LEAFREF_LINKING) && !(option & LY_CTX_LEAFREF_LINKING)) {
+        lyht_free(ctx->term_nodes_ext_ht, ly_ctx_ht_term_node_ext_rec_free);
+    }
 
     if (!(ctx->flags & LY_CTX_SET_PRIV_PARSED) && (option & LY_CTX_SET_PRIV_PARSED)) {
         ctx->flags |= LY_CTX_SET_PRIV_PARSED;
@@ -1317,6 +1353,11 @@ ly_ctx_destroy(struct ly_ctx *ctx)
 
     /* leftover unres */
     lys_unres_glob_erase(&ctx->unres);
+
+    /* clean the term data node extension hash table */
+    if (ctx->term_nodes_ext_ht) {
+        lyht_free(ctx->term_nodes_ext_ht, ly_ctx_ht_term_node_ext_rec_free);
+    }
 
     /* clean the error hash table */
     lyht_free(ctx->err_ht, ly_ctx_ht_err_rec_free);
--- a/src/context.c
+++ b/src/context.c
@@ -243,9 +243,9 @@ ly_ctx_ht_err_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UN
  * @brief Hash table value-equal callback for comparing term data node extension hash table record.
  */
 static ly_bool
-ly_ctx_ht_term_node_ext_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UNUSED(cb_data))
+ly_ctx_ht_leafref_links_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UNUSED(cb_data))
 {
-    struct lyd_term_nodes_ext_rec *rec1 = val1_p, *rec2 = val2_p;
+    struct lyd_leafref_links_rec *rec1 = val1_p, *rec2 = val2_p;
 
     return rec1->node == rec2->node;
 }
@@ -256,11 +256,11 @@ ly_ctx_ht_term_node_ext_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod)
  * @param[in] val_p Pointer to a pointer to a term data node extension item to free with all the siblings.
  */
 static void
-ly_ctx_ht_term_node_ext_rec_free(void *val_p)
+ly_ctx_ht_leafref_links_rec_free(void *val_p)
 {
-    struct lyd_term_nodes_ext_rec *rec = val_p;
+    struct lyd_leafref_links_rec *rec = val_p;
 
-    lyd_free_term_node_ext_rec(rec);
+    lyd_free_leafref_links_rec(rec);
 }
 
 LIBYANG_API_DEF LY_ERR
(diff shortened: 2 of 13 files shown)


michalvasko:
I appreciate the effort but you could have asked before implementing it. We had a similar principle used in *libyang* v1 and decided to abandon it in v2 because 1) it required a lot of additional code when manipulating a data tree (relinking) and 2) caused freeing data to be slow. But finding a leafref target is pretty much *O(1)* (technically *O(n)* where *n* is the depth of a node) so this is really not necessary.

steweg:
Ok, couple of points:
1. The main problem is not to find leafref target, but rather the reverse. Meaning having ability to say if target node is being referenced by some other nodes. Without reverse reference storage, this job is like going through complete data model, finding all leafrefs, finding all their targets and cross-check if given node is in those targets.....
2. How come this be so complex and slow? Finding leafref target is done anyway inside the validation of leafref, so majority of work is already done anyway. Storing the result in form of direct pointers should be very easy. And also storing reverse pointer back to leafref node should not be that hard. The only think left out are cases, when the value of the node is being changed on the fly:
2.1 Node duplication - the references are not copied
2.2 Node deletion (target or leafref node) - the references are automatically cleaned-up on both sides
2.3 Node value change (target or leafref node) - the references are not automatically changed (gap in my implementation)
3. The manual linking can be expensive operation, but that is up to user to consider whether it is really needed or not to be called. But it might be needed if the model is using "required-instance false;" as in that case the validation of leafref is skipped, and as such no automatic linking is done.

Saying all of this, do you have some better proposal how to solve my original use-case to find out if the data node can be safely deleted and not breaking any sort of leafrefs?

From my PoV, I would rather have this ability in the libyang due to automation of linking etc.... And to support that I can offer help to fix also 2.3 point above. And we can also even separate the logic as a separate feature, so if users are concerned about performance and not really need this feature, then they will not be affected at all. If user will need it, they can explicitly enabled by using some new context FLAG.

Would you agree on such approach?

michalvasko:
Okay, if made into a fully optional feature (such as passing a flag to validation), I may be willing to accept it in some form (it was mandatory in *libyang* v1).

But at the same time, it may not be difficult to implement this completely on the side. I can imagine simply creating a hash table (now that it is public) that can link a node to all the leafrefs pointing to it. Yes, creating it would duplicate some work of the validation but then work efficiently for what you need.

steweg:
Ok, I have added context flag for this feature. The validation flag cannot be used as it would be needed to adjust plugin type API for all plugins (seem to be drastic change).
I have also fixed value modification scenario and add tests for it as well

src/context.h
```diff
@@ -199,6 +199,9 @@ struct ly_ctx;
                                         enabled. */
 #define LY_CTX_LEAFREF_EXTENDED 0x0200 /**< By default, path attribute of leafref accepts only path as defined in RFC 7950.
                                         By using this option, the path attribute will also allow using XPath functions as deref() */
+#define LY_CTX_LEAFREF_LINKING_ENABLED 0x0400 /**< Link leafref nodes with its target during validation. It also enables usage of
```
michalvasko, line 202: `ENABLED` seems superfluous, just `LY_CTX_LEAFREF_LINKS` should be fine.

src/tree_data.c
```diff
@@ -3267,3 +3269,82 @@ lyd_get_value(const struct lyd_node *node)
 
     return NULL;
 }
+
+LIBYANG_API_DEF LY_ERR
+lyd_link_leafref_node(struct lyd_node_term *node, struct lyd_node_term *leafref_node)
```
michalvasko, line 3274: What is the use of all these new API functions, why not perform all those tasks automatically?

src/tree_edit.h
```diff
@@ -206,7 +206,7 @@ void *ly_realloc(void *ptr, size_t size);
  * @param[in] ARRAY Pointer to the array to affect.
  */
 #define LY_ARRAY_DECREMENT(ARRAY) \
-        --(*((LY_ARRAY_COUNT_TYPE*)(ARRAY) - 1))
+        --(*((LY_ARRAY_COUNT_TYPE*)(ARRAY) - 1)); \
```
michalvasko, line 209: Redundant change.

src/tree_edit.h
```diff
@@ -231,6 +231,28 @@ void *ly_realloc(void *ptr, size_t size);
 #define LY_ARRAY_FREE(ARRAY) \
         if (ARRAY){free((LY_ARRAY_COUNT_TYPE*)(ARRAY) - 1);}
 
+/**
+ * @brief Remove item from array based on value
+ *
+ * @param[in, out] ARRAY A ([sized array](@ref sizedarrays)) to be modified.
+ * @param[in] INDEX The item position to be removed.
+ */
+#define LY_ARRAY_REMOVE_VALUE(ARRAY, VALUE) \
```
michalvasko, line 240: Seems it keeps the original order, is that necessary, why not just replace the removed value with the last value? If you think it is, wouldn't it be better to call `memmove()` once instead?

michalvasko (changes requested):
Additional question, what about unlinking data nodes, are the leafref links kept intentionally? I am not sure that is a good idea but in any case, the exact behavior should be documented. But I would probably prefer the links to be used only within a single data tree, less prone to errors.

src/tree_data.h
```diff
@@ -854,6 +854,14 @@ struct lyd_node_term {
...
 
     struct lyd_value value;          /**< node's value representation */
+    struct lyd_node_term **leafref_nodes;   /**< list of the leafref data nodes [sized array](@ref sizedarrays)).
+                                                 By default it is empty. It is filled automatically by validation
+                                                 function of leafref nodes. It can also be manipulated manually by
+                                                 using [link api](@ref lyd_link_leafref_node),
+                                                 [unlink](@ref lyd_unlink_leafref_node). Freeing of the resources is
+                                                 automatic. */
+    struct lyd_node_term *target_node;      /**< pointer to leafref target data node, by default is NULL. The logic
+                                                 is the same as for [leafref_nodes](@ref leafref_nodes). */
```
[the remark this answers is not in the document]
steweg: Shifting this to lyd_value doesn't make sense as it will still expand size of 2 pointers to all possible value, so somehow it will not help from memory perspective.
What is needed is the following:
- For all leafref data term nodes it should be stored the `target_node`
- For all data term nodes (including leafrefs) it is needed to store list of `leafref_nodes`
I was originally thinking to use `priv` for this purpose, but as I need to store both sections it would be needed to create some special structure for this + memory mgmt. operations for this. If it is better idea, then I rewrite it like this using such approach.

src/tree_data.c
steweg: There are  the following use-cases:
1. When model is using leafref with `required-instance false;`. In that case the validation of leafref is not happening and as such not even adding links. Therefore it has to be somehow triggered manually. I would say that calling validation is kind of non-intuitive for user and is much more complex procedure to achieve needed steps.
2. When user don't want to use the validation of libyang at all (e.g. using LYD_PARSE_ONLY), but still wish to execute checks of leafrefs and store the links.

In reality I think that we can make `lyd_link_leafref_node` and `lyd_unlink_leafref_node` as internal and keeping just `lyd_link_leafref_node_all` as public API for this use-case. What do you think about that?

src/tree_edit.h
steweg: Done

src/tree_edit.h
steweg: Done

src/tree_data.h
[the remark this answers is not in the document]
steweg: Ok, not sure what memory magic you are referring to, but all the utest were successful including those using valgrind, so I think that memory magic still works. Anyway, if we really want to put this outside `lyd_node_term` then the only thing which can help is the proposed solution using hash_table. If I rewrite it using such approach, will it be acceptable from your PoV?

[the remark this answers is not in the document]
michalvasko: Where would you put a pointer to this hash table? I am also thinking about another solution:

1. Add support for a new optional context callback that would be called for every validated data node in this context.
2. Write one such a public callback that would create a structure with all the leafref backlinks (and maybe also the target for a leafref) and set it to the `priv` pointer.

You said you have some other data in the `priv` pointer but then you could simply write your own callback that would first call this public one and then adjust the `priv` pointer in any way you need. Also, the second point is just so that the functionality is in *libyang* but I do not really care for that, up to you. What do you think?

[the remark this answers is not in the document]
steweg: I have rewritten this using hash table stored within the context, I am not sure if this is better or not, up to you to judge. I personally don't like it as it seems quite strange solution. Anyway it works.

Yes I can make this extension using `priv` + callbacks on my side, but that rather complicates it my side, my project is not in C, but rather in python and I am using [libyang-python c](https://github.com/CESNET/libyang-python/), which doesn't really goes easily with additional C code. On top of that I would need callback not just for validation, but also for node deletion, unlink, value change & duplication.... so changing this like that seems to be even more complicated then having it directly under one C project...

src/tree_data.c
steweg: Rewritten, to have just `lyd_link_leafref_node_tree` as public API

src/common.h
```diff
@@ -373,6 +373,7 @@ struct ly_ctx {
     void *ext_clb_data;               /**< optional private data for ::ly_ctx.ext_clb */
     struct ly_ht *err_ht;             /**< hash table of thread-specific list of errors related to the context */
     pthread_mutex_t lyb_hash_lock;    /**< lock for storing LYB schema hashes in schema nodes */
+    struct ly_ht *term_nodes_ext_ht;  /**< hash table of term data nodes extension */
```
michalvasko, line 376: Please rename this and all the other functions/structures using it. Using `ext` for extensions should be avoided, it is already used for YANG extensions and there is no need for that. I would use `leafref_links_ht`, for example.

src/tree_data.h
```diff
@@ -997,6 +997,20 @@ struct lyd_node_opaq {
     const struct ly_ctx *ctx;       /**< libyang context */
 };
 
+/**
+ * @brief Context term data node record.
+ */
+struct lyd_term_nodes_ext_rec {
```
michalvasko, line 1003: Regarding the renaming, this could then be `lyd_leafref_links_rec`.

src/tree_data.h
```diff
@@ -997,6 +997,20 @@ struct lyd_node_opaq {
...
+    const struct lyd_node_term *node;           /** pointer to the node itself */
+    const struct lyd_node_term **leafref_nodes; /** list of the leafref data nodes [sized array](@ref sizedarrays)).
+                                                    By default it is empty. It is filled automatically by validation
+                                                    function of leafref nodes. It can also be populated based on manual request
+                                                    using [link api](@ref lyd_link_leafref_node_tree). Freeing of the resources
+                                                    is automatic. */
+    const struct lyd_node_term *target_node;    /** pointer to leafref target data node, by default is NULL. The logic
```
michalvasko, line 1010: Mention that it is filled only for leafrefs.

src/tree_data.h
```diff
@@ -997,6 +997,20 @@ struct lyd_node_opaq {
     const struct ly_ctx *ctx;       /**< libyang context */
 };
 
+/**
+ * @brief Context term data node record.
+ */
+struct lyd_term_nodes_ext_rec {
+    const struct lyd_node_term *node;           /** pointer to the node itself */
+    const struct lyd_node_term **leafref_nodes; /** list of the leafref data nodes [sized array](@ref sizedarrays)).
```
michalvasko, line 1005: Explain what `leafref data nodes` are. Specifically, this is the list of leafrefs pointing to `node`, I believe.

src/tree_data.c
```diff
@@ -3267,3 +3272,155 @@ lyd_get_value(const struct lyd_node *node)
...
+ * @brief Gets the term data node extension record for given node
+ *
+ * @param[in] node The term data node.
+ * @param[out] record The term data node extension record
+ * @param[in] create Whether to create record if not exists
+ * @return LY_SUCCESS on success.
+ * @return LY_ERR value on error.
+ */
+LY_ERR
+lyd_get_or_create_term_nodes_ext_record(const struct lyd_node_term *node, struct lyd_term_nodes_ext_rec **record, ly_bool create)
```
michalvasko, line 3286: I believe this function can be `static`.

michalvasko (changes requested):
Alright, I think it can be accepted with this design. I have provided just the first batch of comments, will look at it in more detail after you fix these.

src/tree_data.h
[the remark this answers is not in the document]
michalvasko: The solution seems fine to me.

src/common.h
steweg: Done

src/tree_data.h
steweg: Done

src/tree_data.h
steweg: Done

src/tree_data.c
steweg: Done

src/common.h
michalvasko: Seems you forgot to push the changes, there are none.

src/tree_data.h
steweg: Adjusted

src/context.c
```diff
@@ -239,6 +239,30 @@ ly_ctx_ht_err_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UN
...
+{
+    struct lyd_leafref_links_rec *rec1 = val1_p, *rec2 = val2_p;
+
+    return rec1->node == rec2->node;
+}
+
+/**
+ * @brief Callback for freeing leafref links hash table values.
+ *
+ * @param[in] val_p Pointer to a pointer to a leafref links item to free with all the siblings.
```
michalvasko, line 256: Does not seem accurate and the code proves it, it is just a pointer to the record (which should be correct, you are storing actual records in the HT, not pointers to them).

src/parser_data.h
```diff
@@ -196,7 +196,6 @@ struct ly_in;
  * already a case and another one was added, the previous one is silently auto-deleted. Otherwise (if data from 2 or
  * more cases were created) a validation error is raised,
  * - default values are added.
- *
```
michalvasko, line 199: Redundant, but just a minor issue.

src/tree_data.h
```diff
@@ -997,6 +997,21 @@ struct lyd_node_opaq {
     const struct ly_ctx *ctx;       /**< libyang context */
 };
 
+/**
+ * @brief Structure of leafref links record.
+ */
+struct lyd_leafref_links_rec {
+    const struct lyd_node_term *node;           /** pointer to the data node itself */
+    const struct lyd_node_term **leafref_nodes; /** list of the leafref pointing to this data node [sized array](@ref sizedarrays)),
+                                                    By default it is empty. It is filled automatically by validation
```
michalvasko, line 1006: Probably good to mention that this is true only for `require-instance true` leafrefs.

src/tree_edit.h
```diff
@@ -231,6 +231,29 @@ void *ly_realloc(void *ptr, size_t size);
...
+    { \
+        LY_ARRAY_COUNT_TYPE index__; \
+        ly_bool remove__ = 0; \
+        LY_ARRAY_FOR(ARRAY, index__) { \
+            if (ARRAY[index__] == VALUE) { \
+                if (index__ != LY_ARRAY_COUNT(ARRAY) - 1) { \
+                    memmove(&(ARRAY[index__]), &(ARRAY[LY_ARRAY_COUNT(ARRAY) - 1]), sizeof *(ARRAY)); \
+                } \
+                remove__ = 1; \
```
michalvasko, line 249: Missing `break` after this line?

src/tree_data.h
```diff
@@ -2702,6 +2717,29 @@ LIBYANG_API_DECL LY_ERR ly_time_str2ts(const char *value, struct timespec *ts);
...
+ * @brief Gets the leafref links record for given node
+ *
+ * This API requires usage of LY_CTX_LEAFREF_LINKING context flag.
+ *
+ * @param[in] node The term data node.
+ * @param[out] record The leafref links record
+ * @return LY_SUCCESS on success.
+ * @return LY_ERR value on error.
+ */
+LIBYANG_API_DECL LY_ERR lyd_get_leafref_links(const struct lyd_node_term *node, struct lyd_leafref_links_rec **record);
```
michalvasko, line 2730: To be a bit safer, the out parameter can be `const` I think.

src/tree_data.h
```diff
@@ -2702,6 +2717,29 @@ LIBYANG_API_DECL LY_ERR ly_time_str2ts(const char *value, struct timespec *ts);
...
+ * @brief Traverse through data tree including root node siblings and adds leafrefs links to the given nodes
+ *
+ * This API requires usage of LY_CTX_LEAFREF_LINKING context flag.
+ *
+ * @param[in] tree The data tree root node.
+ * @return LY_SUCCESS on success.
+ * @return LY_ERR value on error.
+ */
+LIBYANG_API_DECL LY_ERR lyd_link_leafref_node_tree(const struct lyd_node *tree);
+
```
michalvasko, line 2742: Just some minor refactoring but I like to use "prefixes" of specialized connected functions so in this case I would rename these to `lyd_leafref_get_links` and `lyd_leafref_link_node_tree`.

michalvasko (changes requested):
Possibly my last comments.

src/tree_data.h
steweg: Done

src/tree_data.h
steweg: Done

src/tree_edit.h
steweg: Done

src/tree_data.h
steweg: Done

src/parser_data.h
steweg: Done

src/context.c
steweg: Reworded

src/context.h
```diff
@@ -199,6 +199,9 @@ struct ly_ctx;
                                         enabled. */
 #define LY_CTX_LEAFREF_EXTENDED 0x0200 /**< By default, path attribute of leafref accepts only path as defined in RFC 7950.
                                         By using this option, the path attribute will also allow using XPath functions as deref() */
+#define LY_CTX_LEAFREF_LINKING 0x0400 /**< Link leafref nodes with its target during validation. It also enables usage of
```
michalvasko, line 202: Please mention the `required-instance true` here as well, not to confuse the users.

src/tree_data.c
```diff
@@ -3267,3 +3272,153 @@ lyd_get_value(const struct lyd_node *node)
...
+        }
+    }
+
+    return LY_SUCCESS;
+}
+
+LIBYANG_API_DEF LY_ERR
+lyd_leafref_get_links(const struct lyd_node_term *node, const struct lyd_leafref_links_rec **record)
+{
+    struct lyd_leafref_links_rec *record2 = NULL;
```
michalvasko, line 3312: Probably not needed, just use `record` directly.

src/tree_data.c
```diff
@@ -3267,3 +3272,153 @@ lyd_get_value(const struct lyd_node *node)
...
+    LY_ERR ret;
+
+    assert(record);
```
michalvasko, line 3315: Do not use `assert` for checking API function parameters, use `LY_CHECK_ARG_RET`, for both `node` and `record`.

src/tree_data.c
```diff
@@ -3267,3 +3272,153 @@ lyd_get_value(const struct lyd_node *node)
...
+{
+    const struct lyd_node *sibling, *elem;
+    struct lyd_node *target;
+    char *errmsg;
+    struct lyd_node_term *leafref_node;
+    struct lysc_node_leaf *leaf_schema;
+    struct lysc_type_leafref *lref;
+    LY_ERR ret;
+
+    assert(tree);
```
michalvasko, line 3361: Param check.

src/tree_edit.h
```diff
@@ -231,6 +231,30 @@ void *ly_realloc(void *ptr, size_t size);
...
```
michalvasko, line 249: Seems `remove__` is completely redundant now.

michalvasko (changes requested):
Seems I have missed a few things before.

src/tree_data.c
steweg: Done

src/tree_data.c
steweg: Done

src/tree_data.c
steweg: Done

src/context.h
steweg: Done

src/tree_edit.h
steweg: Done

src/tree_data.c
```diff
@@ -3309,14 +3309,9 @@ lyd_get_or_create_leafref_links_record(const struct lyd_node_term *node, struct
...
-    struct lyd_leafref_links_rec *record2 = NULL;
-    LY_ERR ret;
-
-    assert(record);
-
-    ret = lyd_get_or_create_leafref_links_record(node, &record2, 0);
-    *record = record2;
-    return ret;
+    LY_CHECK_ARG_RET(NULL, node, LY_EINVAL);
+    LY_CHECK_ARG_RET(NULL, record, LY_EINVAL);
```
michalvasko, line 3313: This macro supports up to 6 arguments to check so you can join these lines. And please leave a blank line after it.

michalvasko (changes requested):
Okay, I think only a last one change is needed and it should be ready.

src/tree_data.c
steweg: Done

michalvasko (approved):

michalvasko:
Fine, thanks.

idryzhov:
Hi! If I understand correctly, this code assumes that a leafref always references a single node in the data-tree (as there's a single `target_node` pointer in `struct lyd_leafref_links_rec`). But in reality leafref can reference a set of nodes, for example, if we reference a list key and the list has multiple keys:
```
list example {
  key "name surname";
  leaf name { type string; }
  leaf surname { type string; }
}

typedef name-ref {
  type leafref {
    path "/example/name";
  }
}
```

How is this code supposed to work with such model?

michalvasko:
Yes, leafref can also reference leaf-lists, which can also have the same values in the state data. It would be good to test it but I believe one of the targets would then be referenced (not sure which, probably the first or last found), which should be fine.

steweg:
I have looked into the code and it all falls to this function `lyplg_type_resolve_leafref`. This function returns one (the first one) or none target node right now. Also in my project we are not using duplicates in leaf-list neither the setup as mentioned by @idryzhov. That's why I have assumed that response can be just one. I have no problems to adjust the structure to multiple ones, but my question to @michalvasko is whether this should really be fixed first by replacing `lyplg_type_resolve_leafref` with some version that supports finding all targets and making this one deprecated. Otherwise it can be misleading for user (as it was for myself etc..)

michalvasko:
Whatever you do, I think you can simply change it, this PR has been merged very recently and has not been released so it should be fine. But I have no opinion regarding what to change, it is up to you who have a use for this, I cannot decide whether there is any value in having all the possibly-referenced targets stored or only a single one is enough.

steweg:
Hi, I have created patch and send it under #2161

PR #2159, opened 2024-01-21
Fix compilation warning: uninitialized ctx

vjardin:
Cosmetic fix


vjardin:
already fixed into the devel branch

PR #2160, opened 2024-01-21
Avoid compilation warning - empty C file

vjardin:
Fix:
  warning: ISO C forbids an empty translation unit

Because of NDEBUG, cmd_debug.c is an empty C file so let's set at least 1 function for this C file that shall be used.


vjardin:
already fixed into the devel branch

PR #2161, opened 2024-01-24
Feature/multiple leafref targets

steweg:
Adding support for multiple leafref target nodes

### DIFF
--- a/src/plugins_types.c
+++ b/src/plugins_types.c
@@ -986,9 +986,21 @@ cleanup:
     return rc;
 }
 
-LIBYANG_API_DEF LY_ERR
-lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct lyd_node *node, struct lyd_value *value,
-        const struct lyd_node *tree, struct lyd_node **target, char **errmsg)
+/**
+ * @brief Find leafref target in data.
+ *
+ * @param[in] lref Leafref type.
+ * @param[in] node Context node.
+ * @param[in] value Target value.
+ * @param[in] tree Full data tree to search in.
+ * @param[out] target Optional first found target.
+ * @param[out] targets Optional array of found targets.
+ * @param[out] errmsg Error message in case of error.
+ * @return LY_ERR value.
+ */
+static LY_ERR
+_lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct lyd_node *node, struct lyd_value *value,
+        const struct lyd_node *tree, struct lyd_node **target, struct lyd_node ***targets, char **errmsg)
 {
     LY_ERR rc = LY_SUCCESS;
     struct lyxp_expr *target_path = NULL;
@@ -996,6 +1008,7 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
     const char *val_str, *xp_err_msg;
     uint32_t i;
     int r;
+    struct lyd_node **item = NULL;
 
     LY_CHECK_ARG_RET(NULL, lref, node, value, errmsg, LY_EINVAL);
 
@@ -1003,6 +1016,11 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
         *target = NULL;
     }
     *errmsg = NULL;
+
+    if (targets) {
+        LY_ARRAY_FREE(*targets);
+        *targets = NULL;
+    }
 
     /* get the canonical value */
     val_str = lyd_value_get_canonical(LYD_CTX(node), value);
@@ -1073,9 +1091,38 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
     if (target) {
         *target = set.val.nodes[i].node;
     }
+    if (targets) {
+        for (i = 0; i < set.used; ++i) {
+            if (set.val.nodes[i].type != LYXP_NODE_ELEM) {
+                continue;
+            }
+            if (((struct lyd_node_term *)set.val.nodes[i].node)->value.realtype != value->realtype) {
+                continue;
+            }
+
+            if (!lref->plugin->compare(&((struct lyd_node_term *)set.val.nodes[i].node)->value, value)) {
+                LY_ARRAY_NEW_GOTO(LYD_CTX(node), *targets, item, rc, cleanup);
+                *item = set.val.nodes[i].node;
+            }
+        }
+    }
 
 cleanup:
     lyxp_expr_free(LYD_CTX(node), target_path);
     lyxp_set_free_content(&set);
     return rc;
+}
+
+LIBYANG_API_DEF LY_ERR
+lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct lyd_node *node, struct lyd_value *value,
+        const struct lyd_node *tree, struct lyd_node **target, char **errmsg)
+{
+    return _lyplg_type_resolve_leafref(lref, node, value, tree, target, NULL, errmsg);
+}
+
+LIBYANG_API_DEF LY_ERR
+lyplg_type_resolve_leafref2(const struct lysc_type_leafref *lref, const struct lyd_node *node, struct lyd_value *value,
+        const struct lyd_node *tree, struct lyd_node ***targets, char **errmsg)
+{
+    return _lyplg_type_resolve_leafref(lref, node, value, tree, NULL, targets, errmsg);
 }
--- a/src/plugins_types.c
+++ b/src/plugins_types.c
@@ -1001,7 +1001,9 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
 
     *errmsg = NULL;
     if (targets) {
-        ly_set_clean(*targets, NULL);
+        /* make sure that the previous set is freed */
+        ly_set_free(*targets, NULL);
+        *targets = NULL;
     }
 
     /* get the canonical value */
@@ -1071,6 +1073,7 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
         goto cleanup;
     }
     if (targets) {
+        ly_set_new(targets);
         for (i = 0; i < set.used; ++i) {
             if (set.val.nodes[i].type != LYXP_NODE_ELEM) {
                 continue;
@@ -1080,7 +1083,8 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
             }
 
             if (!lref->plugin->compare(&((struct lyd_node_term *)set.val.nodes[i].node)->value, value)) {
-                ly_set_add(*targets, set.val.nodes[i].node, 0, NULL);
+                rc = ly_set_add(*targets, set.val.nodes[i].node, 0, NULL);
+                LY_CHECK_GOTO(rc, cleanup);
             }
         }
     }
--- a/src/plugins_types.c
+++ b/src/plugins_types.c
@@ -1001,8 +1001,6 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
 
     *errmsg = NULL;
     if (targets) {
-        /* make sure that the previous set is freed */
-        ly_set_free(*targets, NULL);
         *targets = NULL;
     }
 
@@ -1073,7 +1071,7 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
         goto cleanup;
     }
     if (targets) {
-        ly_set_new(targets);
+        LY_CHECK_GOTO(ly_set_new(targets), cleanup);
         for (i = 0; i < set.used; ++i) {
             if (set.val.nodes[i].type != LYXP_NODE_ELEM) {
                 continue;
--- a/src/plugins_types.c
+++ b/src/plugins_types.c
@@ -1071,7 +1071,7 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
         goto cleanup;
     }
     if (targets) {
-        LY_CHECK_GOTO(ly_set_new(targets), cleanup);
+        LY_CHECK_GOTO(rc = ly_set_new(targets), cleanup);
         for (i = 0; i < set.used; ++i) {
             if (set.val.nodes[i].type != LYXP_NODE_ELEM) {
                 continue;
--- a/src/plugins_types.c
+++ b/src/plugins_types.c
@@ -986,21 +986,9 @@ cleanup:
     return rc;
 }
 
-/**
- * @brief Find leafref target in data.
- *
- * @param[in] lref Leafref type.
- * @param[in] node Context node.
- * @param[in] value Target value.
- * @param[in] tree Full data tree to search in.
- * @param[out] target Optional first found target.
- * @param[out] targets Optional array of found targets.
- * @param[out] errmsg Error message in case of error.
- * @return LY_ERR value.
- */
-static LY_ERR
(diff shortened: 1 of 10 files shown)


src/tree_edit.h
```diff
@@ -241,16 +241,33 @@ void *ly_realloc(void *ptr, size_t size);
...
     }
 
+/**
+ * @brief Checks if the value exists in array. If yes returns RET code
+ *
+ * @param[in] ARRAY ([sized array](@ref sizedarrays)) to be checked.
+ * @param[in] VALUE The item value to be found in ARRAY.
+ * @param[in] RET The return code returned if VALUE exists inside ARRAYA
+ */
+#define LY_CHECK_ARRAY_VALUE_EXISTS_RET(ARRAY, VALUE, RET) \
```
michalvasko, line 261: No point in defining a macro for a single use.

src/tree_data.c
```diff
@@ -3365,54 +3380,74 @@ lyd_leafref_link_node_tree(const struct lyd_node *tree)
...
+ * @brief Removes links between node and node2.
+ *
+ * @param[in] node Data node to which, the leafref is pointing to.
+ * @param[in] node2 Data node to which, the leafref is pointing to.
+ * @param[in] Whether node2 is target data node (1), or leafref data node (0).
+ * @return LY_SUCCESS on success.
+ * @return LY_ERR value on error.
+ */
+static LY_ERR
+_lyd_unlink_leafref_node(const struct lyd_node_term *node, const struct lyd_node_term *node2, ly_bool target)
```
michalvasko, line 3418: Confusing parameters of the function and since it is used only twice with them switched, why not put the whole functionality into `lyd_unlink_leafref_node()`.

src/plugins_types.c
```diff
@@ -1073,9 +1091,38 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
...
+    }
 
 cleanup:
     lyxp_expr_free(LYD_CTX(node), target_path);
     lyxp_set_free_content(&set);
     return rc;
 }
+
+LIBYANG_API_DEF LY_ERR
+lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct lyd_node *node, struct lyd_value *value,
```
michalvasko, line 1117: Is this function even needed? Ignore compatibility, it is a completely new feature so it can be changed without problems.

michalvasko (changes requested):
Please refactor the code to get less functions and less code, it makes it difficult to maintain.

src/tree_edit.h
steweg: Done

src/tree_data.c
steweg: The whole point was to avoid code duplication. Do you prefer to have it duplicated within `lyd_unlink_leafref_node`?

michalvasko: Yes, I would consider it more readable and ultimately requiring less code.

src/plugins_types.c
steweg: Well, it is called just from two places internally, which can be changed to use the second variant allowing multiple targets. Do you want me to do that?

steweg:
@michalvasko One missing test is for your use-case with leaf-list using duplicate values. I am not 100% sure how such test can be done as such thing according to RFC is supported only when leaf-list is using `config false;`. But when I am trying to do that, I am getting strange errors. Here is proposed code snippet.
```
+++ b/tests/utests/data/test_tree_data.c
@@ -720,6 +720,16 @@ test_data_leafref_nodes2(void **state)
             "      path \"../l1/l1\";"
             "    }"
             "  }"
+            "  leaf-list ll1 {"
+            "    type string;"
+            "    config false;"
+            "  }"
+            "  leaf ref2 {"
+            "    type leafref {"
+            "      path \"../ll1\";"
+            "    }"
+            "    config false;"
+            "  }"
             "}";

     UTEST_ADD_MODULE(schema, LYS_IN_YANG, NULL, NULL);
@@ -730,7 +740,9 @@ test_data_leafref_nodes2(void **state)
             "    {\"l1\": \"A\", \"l2\": \"B\"},"
             "    {\"l1\": \"A\", \"l2\": \"C\"}"
             "  ],"
-            "  \"test-data-hash:ref1\": \"A\""
+            "  \"test-data-hash:ref1\": \"A\","
+            "  \"test-data-hash:lll\": [\"asd\", \"qwe\", \"asd\"],"
+            "  \"test-data-hash:ref2\": \"asd\""
             "}";
```

And the error output:
ERROR: LY_SUCCESS != 0x7; MSG: Invalid leafref value "asd" - no target instance "../ll1" with the same value.
Could not run test: [   LINE   ] --- /root/libyang/tests/utests/data/test_tree_data.c:749: error: Failure!ly_err_last(current_utest_context->ctx)

I believe that the my code will work also for this use-case, but it seems like that there is a bug somewhere in original code beyond this leafref linking feature.

src/plugins_types.c
michalvasko: Yes, why not, just simplify the code and in general try to reduce the functionality instead of adding to it. Is there any point in supporting finding only a single target? I would prefer all targets are supported or only one, you have chosen the former with this PR so stick to it please.

src/plugins_types.c
```diff
@@ -1073,9 +1091,38 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
...
+        const struct lyd_node *tree, struct lyd_node **target, char **errmsg)
+{
+    return _lyplg_type_resolve_leafref(lref, node, value, tree, target, NULL, errmsg);
+}
+
+LIBYANG_API_DEF LY_ERR
+lyplg_type_resolve_leafref2(const struct lysc_type_leafref *lref, const struct lyd_node *node, struct lyd_value *value,
```
michalvasko, line 1124: I realized another thing, you are returning a LY_ARRAY from an API function, I do not think we ever do that and return `ly_set` instead, so please do that. I suppose using LY_ARRAY in the hash table records makes sense because it is a bit smaller (but maybe not worth it) so it is up to you to design it sensibly.

michalvasko:
Regarding the code snippet, it seems you are creating `lll` leaf-lists instead of `ll1`.

src/plugins_types.c
steweg: Done

steweg:
> Regarding the code snippet, it seems you are creating `lll` leaf-lists instead of `ll1`.

Thanks, problem solved

src/plugins_types.c
steweg: Done

steweg:
@michalvasko Ok, now I am getting error of ABI check. I guess that this is exactly due to backward compatibility. So how should we proceed? Should I revert it and add `lyplg_type_resolve_leafref2` back again?

michalvasko:
Ignore it, I will move the base version against to check to the updated one after this PR is merged. It makes little difference anyway because the next released version (the first one with these functions) will be SO version 3.

src/xpath.c
```diff
@@ -4062,24 +4063,35 @@ xpath_deref(struct lyxp_set **args, uint32_t UNUSED(arg_count), struct lyxp_set
...
+                    r = LY_EVALID;
+                } else {
+                    /* insert nodes into set */
+                    for (i = 0; i < targets->count; ++i) {
+                        set_insert_node(set, targets->dnodes[i], 0, LYXP_NODE_ELEM, 0);
+                    }
+                    r = LY_SUCCESS;
                 }
+                ly_set_free(targets, NULL);
+                LY_CHECK_RET(r);
```
michalvasko, line 4083: Do not like the magic with `r` here. On error the set does not have to be freed (at least that is how every function is supposed to behave).

src/plugins_types.h
```diff
@@ -146,6 +147,7 @@ struct lysc_type_leafref;
  * - ::lyplg_type_parse_int()
  * - ::lyplg_type_parse_uint()
  * - ::lyplg_type_resolve_leafref()
+ * - ::lyplg_type_resolve_leafref2()
```
michalvasko, line 150: Redundant.

michalvasko (changes requested):
Now it looks much better, with some minor changes it should be good to go.

src/plugins_types.h
```diff
@@ -1199,12 +1201,12 @@ LIBYANG_API_DECL LY_ERR lyplg_type_validate_patterns(struct lysc_pattern **patte
  * @param[in] node Context node.
  * @param[in] value Target value.
  * @param[in] tree Full data tree to search in.
- * @param[out] target Optional found target.
+ * @param[out] target Optional set of found targets.
```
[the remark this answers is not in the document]
steweg: I was trying to keep backward compatibility logic in the same/similar way as was the single pointer before. So the set is filled only if caller is needing it. The need is expressed by providing empty set as out parameter when it have non-NULL value.

Do you want to completely change this and rather always provide results to caller, even if caller is not interested in it?

src/plugins_types/leafref.c
```diff
@@ -79,19 +80,30 @@ lyplg_type_validate_leafref(const struct ly_ctx *ctx, const struct lysc_type *ty
...
+        for (i = 0; i < targets->count; ++i) {
+            ret = lyd_link_leafref_node((struct lyd_node_term *)targets->dnodes[i], (struct lyd_node_term *)ctx_node);
+            LY_CHECK_GOTO(ret, cleanup);
+        }
     }
 
-    return LY_SUCCESS;
+cleanup:
+    if (targets) {
+        ly_set_free(targets, NULL);
```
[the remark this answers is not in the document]
steweg: No, it can't be. The set is created on row 84, if not created then targets is NULL. The line 103 exactly prevents it freeing of NULL pointer....

src/xpath.c
steweg: It should be, it is created within the same block of code on row 4068. So if the function returns without calling ly_set_free, there will be memory leak....normally in other function you are using goto pattern to skip part of code and perform cleanup..... here it is bit hard. I can introduce cleanup goto label, but it will be in the middle of code. Would you prefer such solution?

src/plugins_types.h
steweg: Done

src/plugins_types.h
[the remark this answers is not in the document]
michalvasko: Currently, you can either pass `targets` `NULL` when the targets should not be returned or pass a pointer to an existing set to fill it with the targets. To follow all the other functions, I requested you change the latter functionality so that you pass pointer to a `NULL` set if you want the targets.

src/plugins_types/leafref.c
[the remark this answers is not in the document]
michalvasko: I meant that it is safe to call `ly_set_free()` on `NULL` so the `if` condition is redundant.

src/xpath.c
michalvasko: After you change the behavior, the cleanup should not be needed anymore because the set will not be allocated.

src/plugins_types/leafref.c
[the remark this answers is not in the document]
steweg: Done

src/plugins_types.h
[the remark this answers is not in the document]
steweg: Done

src/xpath.c
steweg: I still think it is necessary to free the set

steweg: Anyway I have refactored it to be more standardized

src/xpath.c
```diff
@@ -4076,14 +4085,16 @@ xpath_deref(struct lyxp_set **args, uint32_t UNUSED(arg_count), struct lyxp_set
                             lyd_get_value(&leaf->node));
                     return LY_EVALID;
```
michalvasko, line 4086: Once you define `cleanup`, it is best to always go there even when not needed (because it might later and the expectation is that cleanup code is always executed).

michalvasko (changes requested):

src/plugins_types.c
```diff
@@ -999,10 +999,12 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
 
     LY_CHECK_ARG_RET(NULL, lref, node, value, errmsg, LY_EINVAL);
 
-    if (target) {
-        *target = NULL;
-    }
     *errmsg = NULL;
+    if (targets) {
+        /* make sure that the previous set is freed */
+        ly_set_free(*targets, NULL);
```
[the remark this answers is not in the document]
steweg: Done

src/plugins_types.c
```diff
@@ -1070,8 +1072,21 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
         }
         goto cleanup;
     }
-    if (target) {
-        *target = set.val.nodes[i].node;
+    if (targets) {
+        ly_set_new(targets);
```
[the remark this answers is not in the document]
steweg: Done

src/tree_data.c
```diff
@@ -3365,20 +3378,27 @@ lyd_leafref_link_node_tree(const struct lyd_node *tree)
 
                 if (leaf_schema->type->basetype == LY_TYPE_LEAFREF) {
                     lref = (struct lysc_type_leafref *)leaf_schema->type;
-                    if (lyplg_type_resolve_leafref(lref, elem, &leafref_node->value, tree, &target, &errmsg)) {
+                    if (lyplg_type_resolve_leafref(lref, elem, &leafref_node->value, tree, &targets, &errmsg)) {
```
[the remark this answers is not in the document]
steweg: Done

src/xpath.c
steweg: Done

src/plugins_types.c
```diff
@@ -1070,8 +1070,21 @@ lyplg_type_resolve_leafref(const struct lysc_type_leafref *lref, const struct ly
         }
         goto cleanup;
     }
-    if (target) {
-        *target = set.val.nodes[i].node;
+    if (targets) {
+        LY_CHECK_GOTO(ly_set_new(targets), cleanup);
```
michalvasko, line 1074: Should be `LY_CHECK_GOTO(rc = ly_set_new(targets), cleanup);` to return an error.

src/xpath.c
```diff
@@ -4037,9 +4039,9 @@ xpath_deref(struct lyxp_set **args, uint32_t UNUSED(arg_count), struct lyxp_set
             oper = (sleaf->flags & LYS_IS_OUTPUT) ? LY_PATH_OPER_OUTPUT : LY_PATH_OPER_INPUT;
 
             /* it was already evaluated on schema, it must succeed */
-            r = ly_path_compile_leafref(set->ctx, &sleaf->node, NULL, lref->path, oper, LY_PATH_TARGET_MANY,
+            ret = ly_path_compile_leafref(set->ctx, &sleaf->node, NULL, lref->path, oper, LY_PATH_TARGET_MANY,
                     LY_VALUE_SCHEMA_RESOLVED, lref->prefixes, &p);
-            if (!r) {
+            if (!ret) {
```
michalvasko, line 4044: Please use another (`r`) variable for non-final return values, avoids the code on line 4053 (and potential future issues).

michalvasko (changes requested):
Last 2 minor problems, should be ready then.

src/plugins_types.c
steweg: Done

src/xpath.c
steweg: Done

michalvasko (approved):

michalvasko:
Alright, should be fine now.

PR #2162, opened 2024-01-25
A couple of fixes for JSON printer

idryzhov:


idryzhov:
I don't understand the CI fail:
```
FAIL: src/printer_json.c (File size changed from 31887 to 31891)
```
Of course the size changed, I modified this file :)

src/printer_json.c
```diff
@@ -477,8 +477,9 @@ json_print_attributes(struct jsonpr_ctx *pctx, const struct lyd_node *node, ly_b
 {
     const struct lys_module *wdmod = NULL;
 
-    if (node->schema && (node->schema->nodetype != LYS_CONTAINER) && (node->flags & LYD_DEFAULT) &&
-            (pctx->options & (LYD_PRINT_WD_ALL_TAG | LYD_PRINT_WD_IMPL_TAG))) {
+    if (node->schema && (node->schema->nodetype != LYS_CONTAINER) && (((node->flags & LYD_DEFAULT) &&
+            (pctx->options & (LYD_PRINT_WD_ALL_TAG | LYD_PRINT_WD_IMPL_TAG))) ||
+            ((pctx->options & LYD_PRINT_WD_ALL_TAG) && lyd_is_default(node)))) {
```
michalvasko, line 482: Can you explain this change? I am looking at the RFC and reading `a data node that is considered by the server to contain default data will include an XML attribute` for the *report-all-tagged* mode. I believe only nodes with the `LYD_DEFAULT` flag are considered default, explicitly set nodes equal to the default value are NOT default.

michalvasko:
The problem is with `uncrustify` - code formatting.

src/printer_json.c
idryzhov: I think the RFC definition is vague, it doesn't really explain what `considered by the server to contain default data` means - implicit defaults only, or values equal to defaults, no matter implicit or explicit. Anyway, that doesn't even matter here. This is basically the difference between `WD_ALL_TAG` and `WD_IMPL_TAG`, from the `WD_IMPL_TAG` docs:
```
Same as LYD_PRINT_WD_ALL_TAG but the attributes are added only to the nodes that
are not explicitly present in the original data tree despite their
value is equal to their default value.
```
This works as described in XML printer, but not in JSON - there's no difference between the two in JSON and this commit fixes it.

I agree with you though that `WD_IMPL_TAG` makes more sense and probably that's what RFC describes.

michalvasko: Right, the description of the values supports your change so I suppose that is how it is intended to work. What exactly was the intention of the RFC is unknown.

src/printer_json.c
```diff
@@ -793,7 +794,8 @@ json_print_leaf_list(struct jsonpr_ctx *pctx, const struct lyd_node *node)
         LY_CHECK_RET(json_print_value(pctx, LYD_CTX(node), &((const struct lyd_node_term *)node)->value, node->schema->module));
 
         if (!pctx->first_leaflist) {
-            if ((node->flags & LYD_DEFAULT) && (pctx->options & (LYD_PRINT_WD_ALL_TAG | LYD_PRINT_WD_IMPL_TAG))) {
+            if (((node->flags & LYD_DEFAULT) && (pctx->options & (LYD_PRINT_WD_ALL_TAG | LYD_PRINT_WD_IMPL_TAG))) ||
+                ((pctx->options & LYD_PRINT_WD_ALL_TAG) && lyd_is_default(node))) {
```
michalvasko, line 798: One more indent is expected.

michalvasko (changes requested):
Should be fine then except the formatting and I have found one error. If fixing that does not help, you should use `uncrustify` to format the sources automatically (if you have it, cmake will inform you, just run `make format`).

idryzhov:
Thanks, I used uncrustify and pushed the updated version.

src/printer_json.c
idryzhov: Just for the record.

I actually looked more into the RFC and it explains everything in the [example](https://datatracker.ietf.org/doc/html/rfc6243#appendix-A.3.2):
```
   In this example, the server's basic mode is equal to 'trim', so all
   data nodes that would contain the schema default value are tagged.
   If the server's basic mode is 'explicit', then only data nodes that
   are not explicitly set data are tagged.  If the server's basic mode
   is 'report-all', then no data nodes are tagged.
```

So, the result of `report-all-tagged` is not always the same and depends on the server's basic mode of operation. If its basic mode is `trim`, then one should use `WG_ALL_TAG` to put a tag on all values equal to the default, even if they are set explicitly. If its basic mode is `explicit`, then one should use `WD_IMPL_TAG` to put a tag only on implicit default values.

michalvasko (approved):

PR #2168, opened 2024-02-08
Bugfix in printer yin

NextLitete:
when a yin file is printed,"include" close label should be '>' instead of '}'


michalvasko:
Quite an obvious bug due to lack of testing, thanks.

Issue #2169, opened 2024-02-08
Way to dissable plugin_types built-in valiadations when using PARSE_ONLY


steweg:
Hi,

is there some easy way to disable advanced leaf/leaf-list type validations when using PARSE_ONLY flag?

Background: I need to somehow evaluate staff like "when" conditions prior I will get valid value from end-user. This is needed for UI layer to effectively hide the field if "when" condition is false. Assumption is that when conditions never needs the value of node itself to evaluate given expression. Of course I can generate random values based on basic types as int8 or string. But the problem will start when I have definitions like this:
```
    typedef multicast-ipv4-prefix {
        type inet:ipv4-prefix {
            pattern '2(2[4-9]|3[0-9])\..*';
        }
        description "The definition of IPv4 multicast prefix";
    }
```
For this one, I need to generate value that is not only matching original `inet:ipv4-prefix`, but also other custom patterns. So generating value which matches all the patterns is very problematic.

So far I have looked into the code and it seems that the main problem is that validations are not separated from storing operations. I have tried to rewrite it and split it. That worked for basic types as string. But then I find out that you have special plugins for ietf based types as ip-address, date-and-time... And for these ones, it seems that I cannot really do that easily as almost each of them has other special methods beyond within the `store` function, which are expecting that the basic string length/pattern validation was already completed correctly. So I can move those parts to `validate` function, but it would mean that if `store` function will end with `LY_EINCOMPLETE`, user must first call `validate` function prior functions like `compare` will work, which I don't really think is a good idea... 

The other way around this is to introduce totally new API to create dummy data nodes by temporarily override lysc_type structure in a way that it will replace the ietf types with basic types and set all advanced validations as string patterns to NULL. Once data node is created, put everything back.

Do you have any better idea, how to achieve that?

michalvasko:
You have correctly observed that you cannot separate value validation and storing, the storage is often tailored to the exact value and so must be valid. Maybe describe the use-case with more details, I understand the general idea but to show/hide nodes with `when` you must have some input data. But you are trying to generate these data so I do not understand what is the condition exactly for showing/hiding the nodes.

steweg:
Ok, in my case I am trying to develop dynamic UI which is driven by YANG schema. Let's consider simple scenario like this:
```
{
  leaf L1 { type string; }
  leaf L2 { type string; when "../leaf1 = 'asd'"; }
}
```
[...]

PR #2171, opened 2024-02-13
Introducing LY_CTX_BASIC_PLUGINS_ONLY & LYPLG_TYPE_STORE_ONLY

steweg:
This patch separates the build-in type validations from store operations to allow storing a node even if the value doesn't pass the validations. To do that a new context option was introduced:
  LY_CTX_STORE_INVALID_DATA

This option also prevent loading of advanced plugins, which cannot really separate store and validation operations

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -283,6 +283,7 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
     struct ly_in *in = NULL;
     LY_ERR rc = LY_SUCCESS;
     struct lys_glob_unres unres = {0};
+    ly_bool builtin_plugins_only;
 
     LY_CHECK_ARG_RET(NULL, new_ctx, LY_EINVAL);
 
@@ -293,7 +294,8 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
     lydict_init(&ctx->dict);
 
     /* plugins */
-    LY_CHECK_ERR_GOTO(lyplg_init(), LOGINT(NULL); rc = LY_EINT, cleanup);
+    builtin_plugins_only = (options & LY_CTX_BUILTIN_PLUGINS_ONLY) ? 1 : 0;
+    LY_CHECK_ERR_GOTO(lyplg_init(builtin_plugins_only), LOGINT(NULL); rc = LY_EINT, cleanup);
 
     if (options & LY_CTX_LEAFREF_LINKING) {
         ctx->leafref_links_ht = lyht_new(1, sizeof(struct lyd_leafref_links_rec), ly_ctx_ht_leafref_links_equal_cb, NULL, 1);
@@ -1234,8 +1236,7 @@ ly_ctx_get_yanglib_data(const struct ly_ctx *ctx, struct lyd_node **root_p, cons
         LY_CHECK_GOTO(ret = ylib_deviation(cont, mod, 0), error);
 
         /* conformance-type */
-        LY_CHECK_GOTO(ret = lyd_new_term(cont, NULL, "conformance-type", mod->implemented ? "implement" : "import", 0,
-                NULL), error);
+        LY_CHECK_GOTO(ret = lyd_new_term(cont, NULL, "conformance-type", mod->implemented ? "implement" : "import", 0, NULL), error);
 
         /* submodule list */
         LY_CHECK_GOTO(ret = ylib_submodules(cont, mod->parsed, 0), error);
--- a/src/context.h
+++ b/src/context.h
@@ -202,6 +202,13 @@ struct ly_ctx;
                                         'require-instance false;'. It also enables usage of
                                         [lyd_leafref_get_links](@ref lyd_leafref_get_links) and
                                         [lyd_leafref_link_node_tree](@ref lyd_leafref_link_node_tree) APIs. */
+#define LY_CTX_BUILTIN_PLUGINS_ONLY 0x0800 /**< By default, context loads all available built-in plugins and extensions. This
+                                        options prohibits loading of built-in non-YANG plugin types and extensions (ipv4-address,
+                                        ipv6-address, etc.). Instead the value is processed by built-in YANG plugins (usually string)
+                                        with all its behavioral implications as hex-string comparison becoming case-sensitive.
+                                        Change of this flag during the lifetime of context is not fully supported. Once the non-YANG
+                                        plugins or extensions are loaded during context creation, they will be used regardless of this
+                                        flag. Therefore it is recommended to set this flag during the first context creation. */
 
 /** @} contextoptions */
 
--- a/src/tree_data.h
+++ b/src/tree_data.h
@@ -282,11 +282,10 @@ struct rb_node;
  * - ::lyd_new_inner()
  * - ::lyd_new_term()
  * - ::lyd_new_term_bin()
- * - ::lyd_new_term_canon()
  * - ::lyd_new_list()
- * - ::lyd_new_list_bin()
- * - ::lyd_new_list_canon()
  * - ::lyd_new_list2()
+ * - ::lyd_new_list3()
+ * - ::lyd_new_list3_bin()
  * - ::lyd_new_any()
  * - ::lyd_new_opaq()
  * - ::lyd_new_opaq2()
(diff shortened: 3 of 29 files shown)


steweg:
Some background:
We have briefly discussed this in #2169. It turned out that opaque node was usable, but only for one use-case to evaluate the "when" condition, without having and lyd_node instance. Unfortunately, I realized that I need it also for cases:
- to evaluate possible leafref targets
- to be able to add implicit default values
- to be able to create whole data tree of values, which comply with just type itself (string is string, int8 is int8...) with no advanced validations during creation/storing
- to be able to identify associated schema from any of the data node.
- to be able to create data nodes from dictionary using libyang-python `dict_to_dnode`, which effectively uses `lyd_new_term`, `lyd_new_list`, `lyd_new_inner` calls. This one is not really a problem of `libyang` but it's a convenient
For these cases usage of an opaque node is very problematic as it requires to store `schema` in `priv` and perform complete tree traversal with temporarily putting a schema value, evaluating, and removing it back. It also can have side effects, as during the evaluation opaque node is wrongly considered as real node, so if anything in the code tries to e.g. access the value of opaq node, it will cause segfaults/invalid reads

src/context.h
```diff
@@ -203,6 +203,15 @@ struct ly_ctx;
                                         'require-instance false;'. It also enables usage of
                                         [lyd_leafref_get_links](@ref lyd_leafref_get_links) and
                                         [lyd_leafref_link_node_tree](@ref lyd_leafref_link_node_tree) APIs. */
+#define LY_CTX_STORE_INVALID_DATA 0x0800 /**< By default, storing of value of given data node is subject to build-in validations
+                                        as patterns, length, ranges etc.. This option bypass these validations, and allow user
```
jktjkt, line 207: nit: typo, two periods

src/context.h
```diff
@@ -203,6 +203,15 @@ struct ly_ctx;
...
+                                        to validate the values manually by additional calls as lyd_validate.
+                                        It also supresses usage of advanced built-in plugin types (ipv4-address, ipv6-address etc..),
+                                        which expects that value is already valid during storage operation. Instead the value is
+                                        processed by basic plungin (usually string).
+                                        Change of this flag during the lifetime of context is not fully supported, once the advanced
+                                        plugins are loaded during context creation, they will be used regardless of this flag.
+                                        Therefore it is recommended to set this flag during the context creation */
```
jktjkt, line 214: nit: missing trailing period

jktjkt, line 214: Unless I'm very wrong, these plugins implement not just data storage, but also conversion of values to their canonical forms. Many types are defined to accept a wide range of textual inputs where many "different" inputs will result in a single canonical value. Examples are uppercase vs. lowercase values in `ietf-yang-types:hex-string`. If you disable these plugins, values will not be converted to their canonical form. I suspect that this will in turn break any reasonable `when` and `must` statements because these are defined to use comparisons against the canonical form.

If this patch gets accepted, at least the documentation should be clear about these implications.

src/context.h
steweg: Done

src/context.h
steweg: True, but in my case, it is acceptable. Description was extended with such consequence.

steweg: Done

michalvasko (changes requested):
Okay, I think your use-case can be supported but using a different design that may even improve things overall. It can be divided into 2 independent parts:

1) New context flag for it to avoid using (loading) any type plugins except for the basic ones for the built-in YANG types. I believe this should also include external plugins (which you have not disabled in this PR).

2) New internal type plugin option `LYPLG_TYPE_STORE_ONLY` which would be set for every store callback called when parsing data with `LYD_PARSE_ONLY` and would mean that the callback can skip all validation tasks that would not prevent the value from being stored successfully.

    You have mostly implemented this already leading to more `validate` callbacks, which would not be called in `store` if the new option is set (and never return `LY_EINCOMPLETE`), called otherwise. This would lead to another important change in the background, some validate callbacks would need the data tree (current ones) and some not (the new ones). I am not certain this is relevant and everything will probably be fine if this difference is ignored and the data are always provided to `validate` (because it would normally be called as part of `store`).

    This will also lead to more efficient `LYD_PARSE_ONLY` that would skip redundant type validation. Also, as you mentioned, you will need this new functionality not only as part of `lyd_parse()` so you can add functions such as `lyd_new_term_store_only()`, not sure if you need anything else.

steweg:
On 1st point:
- I have kept the external plugins on purpose. The reason is that user have a control about this. So if user provide his own plugin, he knows it's behavior and must take it into account. The problematic part is only advanced plugins within libyang, which you currently cannot really control from user perspective.
- I was also thinking about possibility to have different set of plugins per context, but this can be done as a separate patch in future (if needed), as well as some more granular filter of which plugins to load and which not (kind of deny-list or allow-list...)

On 2nd point:
- creation of separate parse flag on plugin level is kind of ok. The only drawback will be that no all plugins will support it, which could be misleading to user. Or do you want me to perform check in all advanced plugin, so that if user will try to use that flag with advanced plugin, it will fail with LY_EDENIED?
- I have put the validations calls into the store functions to avoid need to have duplicate code. If you prefer duplicate code, then I can rewrite it as well, just let me know
- Return of `LY_EINCOMPLETE` was already used by leafref plugin, that's why I have followed the same logic, as it seem to be the most generic way. If we don't return this code, caller will not be aware if the validation call is needed or not after the storing. To do that, it will be needed to combine information of what exact type of plugin was called + whether new flag was used + return code (something similar what I have done in `schema_compile.c`). Or simply always call validation regardless of everything. The cleanest alternative would be to totally separate storing and validations, but in that case we would introduce backward compatibility issue. And again introduce problem of what to do with advanced plugin that cannot really separate those operations. So I still think, that current solution with using of LY_EINCOMPLETE is the correct to avoid all side-effects. Please confirm how to proceed from your PoV

michalvasko (changes requested):
Without looking at this in much detail, I meant it when I said that you can add new functions for the store-only flag, please do not modify the current ones. This is not about backwards-compatibility, they already have one annoying parameter (`output`), they do not need another. Also, this makes sense only for a few functions such as `lyd_new_term()` and okay, even `lyd_new_list()` or `lyd_new_meta()`.

steweg:
Well adding a new variant of public calls seem to be quite odd..... how about converting `ly_bool output` in `lyd_new_` to `uint32_t options` or `uint8_t options`? This way we can convert output to `LYD_NEW_LIST_OUTPUT = 1` and introduce `LYD_NEW_LIST_STORE_ONLY = 2` and kind of maintain backward compatibility.

Or do it totally different and say that all `lyd_new_*` API will always imply `LYPLG_TYPE_STORE_ONLY` usage

michalvasko:
> Or do it totally different and say that all lyd_new_* API will always imply LYPLG_TYPE_STORE_ONLY usage

Was thinking about that but I do not think it is really desired, in general, to not get an error on an invalid value.

But the `options` parameter is a good idea. It would allow for removal of `lyd_new_(term|list)_(bin|canon)()` as well. So we end up with options: `LYD_NEW_OUTPUT`, `LYD_NEW_VAL_STORE_ONLY`, `LYD_NEW_VAL_BIN`, and `LYD_NEW_VAL_CANON`.

One just needs to sanitize the options because some should not be combined (or an option implies another). Specifically:
- `LYD_NEW_VAL_STORE_ONLY` and `LYD_NEW_VAL_CANON` - for store_only the canonical value should not be generated because it may not be possible on invalid values
- `LYD_NEW_VAL_STORE_ONLY` and `LYD_NEW_VAL_BIN` - binary value must always be canonical and valid
- `LYD_NEW_VAL_CANON` and `LYD_NEW_VAL_BIN` - value is either binary or canonical string

Also, the `LYD_NEW_VAL_*` options will make no sense for `lyd_new_inner()`. I think this will simplify the API and make it easily extensible (make the options `uint32_t` to be safe), thanks for your effort.

steweg:
I have accidentally push incomplete patch, no review is needed yet. I will let you know once finished on my side

steweg:
Ok, now it should be the ready for review

michalvasko (changes requested):
So again, thanks for the effort, a lot of changes were required. Unfortunately, I have not realized that adding the binary value option to the `new` functions would add another parameter (value length) and make the value `void *` instead of `char *`. I would really prefer to avoid changes requiring rewriting of all these functions usage in existing code (as long as no options are needed) so I have to ask you to keep this single variant (`_bin`) in a separate function. However, the documentation still needs to be improved significantly and I see no point in adding all those comments so I will do that separately after merging this PR and I could also make the binary value change, up to you.

Another thing, please change the target branch to `new_so`, these changes cannot simply be merged into `devel`, the other projects needs to be synced first with the changes.

Additional note, `lyplg_type_check_hints()` should probably be a part of `validate` callbacks as well, it does not affect the storing process.

src/tree_data_new.c
```diff
@@ -736,13 +715,14 @@ lyd_new_ext_list(const struct lysc_ext_instance *ext, const char *name, struct l
...
 lyd_new_list2(struct lyd_node *parent, const struct lys_module *module, const char *name, const char *keys,
-        ly_bool output, struct lyd_node **node)
+        uint32_t options, struct lyd_node **node)
 {
     LY_ERR r;
     struct lyd_node *ret = NULL;
     const struct lysc_node *schema;
     struct lysc_ext_instance *ext = NULL;
     const struct ly_ctx *ctx = parent ? LYD_CTX(parent) : (module ? module->ctx : NULL);
+    uint32_t lys_opts = (options & LYD_NEW_VAL_OUTPUT) ? LYS_GETNEXT_OUTPUT : 0;
```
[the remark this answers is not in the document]
steweg: Done

src/tree_data_new.c
```diff
@@ -846,66 +818,36 @@ _lyd_new_list3(struct lyd_node *parent, const struct lys_module *module, const c
...
-        size_t value_len, LY_VALUE_FORMAT format, ly_bool output, struct lyd_node **node)
+lyd_new_term(struct lyd_node *parent, const struct lys_module *module, const char *name, const void *value,
+        size_t value_len, uint32_t options, struct lyd_node **node)
 {
     LY_ERR r;
     struct lyd_node *ret = NULL;
     const struct lysc_node *schema;
     struct lysc_ext_instance *ext = NULL;
     const struct ly_ctx *ctx = parent ? LYD_CTX(parent) : (module ? module->ctx : NULL);
+    uint32_t lys_opts = (options & LYD_NEW_VAL_OUTPUT) ? LYS_GETNEXT_OUTPUT : 0;
```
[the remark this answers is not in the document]
steweg: Done

src/tree_data_new.c
```diff
@@ -545,25 +568,25 @@ lyd_new_ext_inner(const struct lysc_ext_instance *ext, const char *name, struct
...
  */
 static LY_ERR
 _lyd_new_list_node(const struct ly_ctx *ctx, const struct lyd_node *parent, const struct lys_module *module,
-        const char *name, ly_bool output, struct lyd_node **node)
+        const char *name, uint32_t options, struct lyd_node **node)
 {
     struct lyd_node *ret = NULL;
     const struct lysc_node *schema;
     struct lysc_ext_instance *ext = NULL;
+    uint32_t lys_opts = (options & LYD_NEW_VAL_OUTPUT) ? LYS_GETNEXT_OUTPUT : 0;
```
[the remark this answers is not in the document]
steweg: Done

src/plugins_exts/schema_mount.c
```diff
@@ -1007,7 +1007,7 @@ schema_mount_validate(struct lysc_ext_instance *ext, struct lyd_node *sibling, c
         while (lyd_parent(diff_parent)) {
             diff_parent = lyd_parent(diff_parent);
         }
-        if ((ret = lyd_new_meta(LYD_CTX(diff_parent), diff_parent, NULL, "yang:operation", "none", 0, NULL))) {
+        if ((ret = lyd_new_meta(LYD_CTX(diff_parent), diff_parent, NULL, "yang:operation", "none", LYD_NEW_VAL_STORE_ONLY, NULL))) {
```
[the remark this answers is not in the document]
steweg: I though that it was used only for some diff part, for which validation is not needed. Anyway changed to 0.

steweg:
I have put back the `lyd_new_term_bin()`, with just minor update to use options. And updated restrictions of `lyd_new_term()` to be not working with binary value. Rebased on new_so and squashed my patches into one.

I deliberately kept `lyplg_type_check_hints` within the `store`, just to force user to put value, which really is of given type (so if I put value of `"asd"` to node of type `int8`, I get an error).

src/context.c
```diff
@@ -283,6 +283,7 @@ ly_ctx_new(const char *search_dir, uint16_t options, struct ly_ctx **new_ctx)
     struct ly_in *in = NULL;
     LY_ERR rc = LY_SUCCESS;
     struct lys_glob_unres unres = {0};
+    ly_bool no_advanced_builtin;
```
michalvasko, line 286: Please rename this to be accurate. I would probably use a negation, `builtin_plugins_only`.

src/context.h
```diff
@@ -202,6 +202,13 @@ struct ly_ctx;
                                         'require-instance false;'. It also enables usage of
                                         [lyd_leafref_get_links](@ref lyd_leafref_get_links) and
                                         [lyd_leafref_link_node_tree](@ref lyd_leafref_link_node_tree) APIs. */
+#define LY_CTX_BASIC_PLUGINS_ONLY 0x0800 /**< By default, context loads all available built-in plugins. This options prohibits
+                                        loading of advanced built-in plugin types (ipv4-address, ipv6-address, etc.). Instead the value is
+                                        processed by basic built-in pluings (usually string) with all its behavioral implications as
+                                        hex-string comparison becoming case-sensitive.
+                                        Change of this flag during the lifetime of context is not fully supported. Once the advanced
+                                        plugins are loaded during context creation, they will be used regardless of this flag.
+                                        Therefore it is recommended to set this flag during the first context creation. */
```
michalvasko, line 211: Clarify this a bit and please use built-in as in "built-in YANG types" and not "built-in libyang plugins". Then the option can be renamed to `LY_CTX_BUILTIN_TYPE_PLUGINS_ONLY` which should give a better idea about the loaded plugins.

src/tree_data.h
```diff
@@ -1248,53 +1245,60 @@ LIBYANG_API_DECL LY_ERR lyd_new_inner(struct lyd_node *parent, const struct lys_
...
- * @param[in] ... Ordered canonical key values of the new list instance, all must be set. No keys are expected for
- * key-less lists.
+ * @param[in] ... Ordered key values of the new list instance, all must be set. In case of an instance-identifier
+ * or identityref value, the JSON format is expected (module names instead of prefixes). No keys are expected for key-less lists.
+ * In case of format set to LY_FORMAT_LYB, every key value must be followed by its length.
  * @return LY_ERR value.
  */
-LIBYANG_API_DECL LY_ERR lyd_new_list_canon(struct lyd_node *parent, const struct lys_module *module, const char *name,
-        ly_bool output, struct lyd_node **node, ...);
+LIBYANG_API_DECL LY_ERR lyd_new_list(struct lyd_node *parent, const struct lys_module *module, const char *name,
```
michalvasko, line 1300: Please keep the separate `lyd_new_list3_bin()` to avoid needing `LYD_NEW_VAL_BIN_VALUE` (so please remove it), which is accepted only for this function. It is true that internally it would be useful but that is not a reason to keep it public so I leave up to you how to communicate a binary value internally.

michalvasko (changes requested):
Almost there, hopefully some final adjustments needed.

michalvasko:
Regarding `lyplg_type_check_hints()`, it should not cause problems to skip it when storing the value. For `int`, for example, the value is always parsed (to check it can be stored) and that will fail for any invalid value.

I am talking mainly about encoding problems that if a number is encoded as a string, it can still be stored and the validation should fail. However, hints are not available in the validate callback and it is not worth it to add them there so leave it the way it is then.

src/context.c
steweg: done

src/context.h
steweg: Done, but used `LY_CTX_BUILTIN_PLUGINS_ONLY` as it disables also built-in extensions

michalvasko (approved):
Seems fine now, thanks, I will now improve the documentation myself.

PR #2174, opened 2024-02-14
This commit fixes absent clearing memory, allocated by vasprintf for …

optimden:
This commit fixes absent clearing memory, allocated by vasprintf for "char *msg".

In case ly_realloc for *out->method.mem.buf fail, we don't free msg pointer now.
So, it can produces a memory leak.


michalvasko (approved):

michalvasko (changes requested):
You have used TABs instead of spaces for indent.

michalvasko (approved):

PR #2180, opened 2024-02-21
schema printer BUGFIX extension instances

zhangtaog:
when a yin file is printed, LY_STMT_EXTENSION_INSTANCE statements need to achieve


michalvasko (approved):
I believe the comment can also now be removed.

PR #2188, opened 2024-02-27
add option to use 16bits hash in lyb

billykinggym:
lyb hash only has 8bits, which not sufficient in some cases
this PR add a compile option that the devolper could set the hash size to 16bit if needed. by default , the value is 8bit as before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,6 +251,8 @@ else()
     set(INTERNAL_DOCS NO)
 endif()
 
+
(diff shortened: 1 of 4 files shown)


michalvasko:
Yes, the hash size is actually in public API which makes libraries with 8b vs 16b hashes incompatible not just when communicating using LYB but even their ABIs. And there are only a few changes needed, is it necessary to be merged into the repo? To keep things simple, I would prefer you keep it as a local patch. Like I said, the libraries are not compatible when the hash size changes.

PR #2194, opened 2024-03-04
Crating a new opaq leaflist node the value isn't populated.

lucoliva:
Crating a new opaq leaflist node the value isn't populated. The value is passed to lyd_create_opaq as prefix parameter.


michalvasko (approved):

michalvasko:
Right, thanks.

PR #2199, opened 2024-03-08
[libyang]Add a void null pointer dereference.

13935000463:
The eval_name_test_with_predicate_get_scnode function is modified. The non-null judgment of node->schema is added to avoid null pointer dereference.

### DIFF
--- a/src/xpath.c
+++ b/src/xpath.c
@@ -7985,7 +7985,7 @@ continue_search:
             scnode = lys_find_child(NULL, moveto_mod, name, name_len, 0, 0);
         }
     } else if (!*found || (lysc_data_parent(*found) != node->schema)) {
-        if ((format == LY_VALUE_JSON) && !moveto_mod) {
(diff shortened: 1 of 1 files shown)


michalvasko (changes requested):
Actually, `lyd_node_schema()` function should be used at all the places instead of direct dereference to avoid similar issues. Also, we only accept PRs to the `devel` branch. If you want, I can fix this issue properly but if you update it according to what I said, I will merge it.

13935000463:
> Actually, `lyd_node_schema()` function should be used at all the places instead of direct dereference to avoid similar issues. Also, we only accept PRs to the `devel` branch. If you want, I can fix this issue properly but if you update it according to what I said, I will merge it.

The node input parameter is used in the lyd_opaq_node scenario.I have a possibility that the schema here will be empty in the actual usage scenario, causing a reset. I hope to merge it, thank you very much.

michalvasko:
My main point was that your fix was not a proper one, there were still other cases when an opaque node would cause a crash. I have committed what should be a full fix.

PR #2203, opened 2024-03-19
libyang checks: start with FRRouting

vjardin:
Hi,

the main purpose of this pull request is to be able to have a framework to test libyang (including the current pull requests, some tags for references) along with the libyang in order to easy the shared integration.

It starts with FRRouting, but it can be extended with other projects.

Currently, it is a test based on compile and link. Once it'll be merged, I can extend it for some few smoke tests from the projects that consume libyang so they can report a proper execution of their smoke tests.

### DIFF
--- a/.github/workflows/frr-checks.yml
+++ b/.github/workflows/frr-checks.yml
@@ -17,7 +17,6 @@ jobs:
         frr-versions:
           - master
         libyang-versions:
-          - v2.1.128
           - ${{ github.ref_name }}
     steps:
       - name: add missing packages per building-frr-for-ubuntu2204
--- a/.github/workflows/frr-checks.yml
+++ b/.github/workflows/frr-checks.yml
@@ -1,0 +1,111 @@
+name: libyang+FRR HEAD CI
+run-name: libyang CI FRR ${{ github.actor }}  ⚗️
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - '**'
+jobs:
+  build-frr:
+    name: Build FRR
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        compiler: [ gcc ]
+        frr-versions: [ master ]
+        libyang-versions:
+          - v2.1.128
+          - ${{ github.ref_name }}
+    steps:
+      - name: add missing packages per building-frr-for-ubuntu2204
+        uses: ConorMacBride/install-package@v1
+        with:
+          apt:
+            git
+            autoconf
+            libtool
+            make
+            libreadline-dev
+            texinfo
+            pkg-config
+            libelf-dev
+            libpam0g-dev
+            libjson-c-dev
+            bison
+            flex
+            libc-ares-dev
+            python3-dev
+            python3-sphinx
+            install-info
+            build-essential
+            libsnmp-dev
+            perl
+            libcap-dev
+            libelf-dev
+            libunwind-dev
+            protobuf-c-compiler
+            libprotobuf-c-dev
+            libgrpc++-dev
+            protobuf-compiler-grpc
+            libsqlite3-dev
+            libzmq5
+            libzmq3-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/frr-checks.yml
```diff
@@ -0,0 +1,111 @@
...
+jobs:
+  build-frr:
+    name: Build FRR
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        compiler: [ gcc ]
+        frr-versions: [ master ]
+        libyang-versions:
+          - v2.1.128
```
michalvasko, line 19: Why use a specific version? What sense does it make to use the same version every time? It will not use the latest changes that should be tested I assumed.

michalvasko (changes requested):
Seems the scripts do not work, look at the CI output.

.github/workflows/frr-checks.yml
vjardin: It is to have a reference and start this question to be transparent.

There are 2 sides:
  - on FRR side, you are right, it should compile with many libyang versions ; I'll work on it.
  - on libyang side, you are right, 1 libyang should compile with many FRR versions ; for the sake of this merge request, I am just using "master" of FRR.

I could drop v2.1.128 from the pull request checks, but I feel that it should run on a scheduled Action : what's about once a day for instance ?

michalvasko: Once a day is fine but we have used triggers for now so it can also be on every push (to `devel` ideally because `master` is changed only on release).

vjardin: @michalvasko , see https://github.com/CESNET/libyang/pull/2203/commits/8d644cacd88224c4e6dc2a38320e3c1d65a4402e which shows how such matrix of cases could be managed from FRRouting side too.

michalvasko: I am still not sure I understand. For libyang, I would think you always want to test the latest version with latest changes whether there were no incompatibilities introduced. As for FRR, I guess there could be several versions to check that none of the supported FRR versions were broken.

vjardin: @michalvasko > For libyang, I would think you always want to test the latest version with latest changes 
@vjardin > yes, so I did update the pull request dropping the specific version of libyang for the time being. A full set of matrix run should be for a scheduled job.

@michalvasko > As for FRR, I guess there could be several versions to check
@vjardin > yes, so the array:
```
        frr-versions:
          - master
```
will be extended accordingly after this pull request.

michalvasko: Okay, so please fix the initial _current libyang_ - _frr master_ build so that it finishes successfully.

vjardin: Hmmm 🤔 currently I cannot since the evolutions of libyang breaks the applications using it like FRR. So I'd suggest the following :
  - step1: merge it asis so this pull request is focuses on this template for such Action
  - step2: fix current libyang - frr master to comply and be successful (as you suggest)
  - step3: add scheduled build with tags / versions matrix for a broader nightly check

Is it ok for you ?

michalvasko: > Is it ok for you ?

Not really because I do not understand.

1. Why merge an action that is going to fail? Why not wait until it actually succeeds?
2. Yes, it is up to you to do this (after libyang is released) and then a successful action can be added.
3. This is an option although I prefer actions triggered by some changes (push), similarly to our `devel-push` action.

vjardin: OK, I get your point related to 1. and 2. I'll follow your suggestion "fix first", or least I'll explore it deeper.

Regarding 3. ; I'll submit a specific pull request that leverages "scheduled build" only using v2.1.x and then later, once issues will be fixed, it can be extended (which I do prefer like you) with an action on every push.

vjardin:
I am closing this pull request and I'll (re)open a new one once issues will be fixed.

PR #2204, opened 2024-03-20
[cosmetic] CodeQL actions and checkout

vjardin:
Let's upgrade to the latest versions in order to avoid some warnings.


michalvasko (changes requested):
Right, I have updated this in `ci.yml` but not the rest. However, I have used `*@main` to avoid this problem in future, please change it similarly.

PR #2209, opened 2024-03-27
distro UPDATE separate package for YANG modules

idryzhov:
Currently, YANG modules are installed as part of the main libyang deb/rpm package. This makes it impossible to install both libyang2 and libyang3 at the same time, because package managers don't allow to simultaneously install packages that provide the same files.

As the YANG modules are already embedded into libyang binaries and not needed for libyang operation, we can move them to a separate libyang-data optional package. It allows to install both libyang2 and libyang3 at the same time.

P.S. Another option is to install YANG modules as part of `libyang-dev` package. Please, let me know if you think it's a better option and I'll update my PR accordingly.

### DIFF
--- a/distro/pkg/deb/control
+++ b/distro/pkg/deb/control
@@ -63,3 +63,11 @@ Description: parser toolkit for IETF YANG data modeling - executable tools
  during the creation of IETF YANG schemas.  The tools are not generally
  useful for normal operation where libyang primarily processes configuration
  data, not schemas.
+
+Package: libyang-modules
+Section: devel
+Architecture: all
+Description: parser toolkit for IETF YANG data modeling - YANG modules
+ This package provides the IETF YANG modules used by libyang. These modules
+ are embedded into libyang so this package is not necessary for libyang
+ operation.
--- a/distro/pkg/deb/libyang-modules.install
+++ b/distro/pkg/deb/libyang-modules.install
@@ -1,0 +1,1 @@
+usr/share/yang/modules/libyang
--- a/distro/pkg/deb/libyang3.install
+++ b/distro/pkg/deb/libyang3.install
@@ -1,2 +1,1 @@
 usr/lib/*/*.so.*
-usr/share/yang/modules/libyang
--- a/distro/pkg/rpm/libyang.spec
+++ b/distro/pkg/rpm/libyang.spec
@@ -17,6 +17,9 @@ BuildRequires:  cmake(cmocka) >= 1.0.1
 BuildRequires:  make
 BuildRequires:  pkgconfig(libpcre2-8) >= 10.21
 
+%package modules
+Summary:    YANG modules for libyang
+
 %package devel
 Summary:    Development files for libyang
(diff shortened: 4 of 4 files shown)


michalvasko:
I do not like the idea of another separate package much. Moving it to the `libyang-dev` package is one option, acceptable for me. Another solution would be to rename the YANG module directory simply to `libyang3`. What do you think about that? I do not have a strong preference for this compared to your suggestion.

idryzhov:
Actually I realized that putting the models into the `dev` package will not work. It will conflict with `libyang2`, as `libyang2` already provides the models and we can't change that retroactively.

Your idea of putting them into another directory for `libyang3` should work. But the reason I decided to use separate packages is because the models are not really required for libyang operation. I'm not even sure why does libyang install the models at all, maybe you can clarify that?

idryzhov:
Hi @michalvasko, do you have any comment? We need to sort this out before the first release of `libyang3`, otherwise we'll have huge problems. Basically, we won't be able to install it on any system that already has a package that requires `libyang2`.

jktjkt:
While I recognize that some people are using libyang without sysrepo, I think that in the majority of deployments it will be libyang+sysrepo+netopeer2. Unless I'm mistaken, it is not possible to mix two different sysrpeo versions, so the odds of someone having libyang v2 and v3 at the same time and their systems still working are rather slim. Perhaps a simple package-level blocker might be enough?

idryzhov:
https://github.com/FRRouting/frr is using libyang and it's a pretty popular package. People may be using older versions of FRR which uses libyang2 on the same system with newer sysrepo which uses libyang3. It will work properly as libyang2/3 have different major SO versions. I don't see any reason to prohibit that.

If you're ok with accepting the solution of putting the model files into a different directory then I'll update my PR. Removing this conflict between the packages is more important for me than having the models in a separate package. I just thought that a separate package is a more clear solution.

distro/pkg/deb/control
```diff
@@ -63,3 +63,11 @@ Description: parser toolkit for IETF YANG data modeling - executable tools
  during the creation of IETF YANG schemas.  The tools are not generally
  useful for normal operation where libyang primarily processes configuration
  data, not schemas.
+
+Package: libyang-data
```
michalvasko, line 67: Maybe call the package `libyang-modules`?

michalvasko (changes requested):

michalvasko:
Okay, after some consideration I am fine with having a separate package. You are right that they are not needed for libyang, they are being installed mostly to follow all the other projects that install their YANG modules and can be useful for other applications based on YANG.

distro/pkg/deb/control
idryzhov: Sure, I'll update the PR a bit later today. Thanks!

idryzhov:
Updated the package name to `libyang-modules`.

michalvasko (approved):

michalvasko:
Alright, merging.

PR #2213, opened 2024-04-01
Adding context based extensions plugins support

steweg:
This patch adds ability to load plugins directly from memory without need to create shared library by using lyplg_add_plugin() API.

It also allows to associate plugin directly with context, so given plugin will not affect all contexts, just given context

### DIFF
--- a/src/plugins.h
+++ b/src/plugins.h
@@ -85,6 +85,24 @@ enum LYPLG {
  */
 LIBYANG_API_DECL LY_ERR lyplg_add(const char *pathname);
 
+/**
+ * @brief Manually load a plugins from memory
+ *
+ * Note, that a plugin can be loaded only if there is at least one context. The loaded plugins are connected with the
+ * existence of a context. When all the contexts are destroyed, all the plugins are unloaded.
+ *
+ * @param[in] ctx The context to which the plugin should be associated with. If NULL, the plugin is considered to be shared
+ * between all existing contexts.
+ * @param[in] version The version of plugin records.
+ * @param[in] type The type of plugins records.
+ * @param[in] recs An array of plugin records provided by the plugin implementation. The array must be terminated by a zeroed
+ * record.
+ *
+ * @return LY_SUCCESS if the plugins with compatible version were successfully loaded.
+ * @return LY_EDENIED in case there is no context and the plugin cannot be loaded.
+ * @return LY_EINVAL when recs is NULL or the plugin contains invalid content for this libyang version.
+ */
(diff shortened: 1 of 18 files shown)


src/plugins.h
```diff
@@ -85,6 +85,24 @@ enum LYPLG {
...
+ * @param[in] version The version of plugin records.
+ * @param[in] type The type of plugins records.
+ * @param[in] recs An array of plugin records provided by the plugin implementation. The array must be terminated by a zeroed
+ * record.
+ *
+ * @return LY_SUCCESS if the plugins with compatible version were successfully loaded.
+ * @return LY_EDENIED in case there is no context and the plugin cannot be loaded.
+ * @return LY_EINVAL when recs is NULL or the plugin contains invalid content for this libyang version.
+ */
+LIBYANG_API_DECL LY_ERR lyplg_add_plugin(struct ly_ctx *ctx, uint32_t version, enum LYPLG type, const void *recs);
```
michalvasko, line 105: I would prefer having 2 functions for each plugin type so that `recs` can be a specific type instead of `void *`.

michalvasko (changes requested):
Despite this being a fairly significant change, the code seems clean and I have just one idea for a minor improvement.

src/plugins.h
steweg: Ok, no problem to do that. With that approach, I guess that I can remove even the `version` from the API as this will be implicitly defined by the plugin record structure itself. Can you confirm?

michalvasko: It seems in the current design the API version is a separate symbol in the SO hidden by one macro that the plugins should use, not in the `record` structure. So the version will have to remain a separate argument.

steweg: But wasn't that introduced exactly because you need to be sure that you can cast safely from `void *` to specific plugin record structure? I guess that only if you are changing plugin record structure itself, you are increasing the expected version number.

If the new API will expect exact plugin structure on the input, so no casting is needed, then the `version` seems to be quite point-less. Or am I missing something?

steweg: Done. Kept the `version`

But frankly speaking it seems to be pointless from API perspective to have 2 separate APIs as anyway the code needs to be converted anyway to `void *` if we would like to reuse existing code as much as possible. Anyway it is up to you to judge.

michalvasko (approved):

src/plugins.h
michalvasko: Internally converting to `void *` is fine, I just wanted the type to be in the API function prototype so that the users know it immediately without having to look for it and to get warnings in case they get it wrong.

michalvasko:
Okay, thanks.

PR #2216, opened 2024-04-04
Adding ability to use JSON 'null' values

steweg:
This patch introduces parsing flag, which allows user to use JSON 'null' value with leaf, leaf-list and anydata


michalvasko (approved):

michalvasko:
Okay, looks good.

PR #2218, opened 2024-04-05
Allowing JSON 'null' value for inner data nodes

steweg:


steweg:
Sorry, I forgot to allow this for containers/lists in my previous patch

michalvasko (approved):

PR #2244, opened 2024-05-23
tree schema BUGFIX correct predicate order in lysc_path

snar:
Assuming simplified schema
```
list m { key "a b c" ... leaf data {}}
```
`lyd_path` of data node gives us `/m[a='..'][b='..'][c='..']`, and `lysc_path(..., LYSC_PATH_DATA_PATTERN, ...)` returns predicates in reverse order: `/m[c='%s'][b='%s'][a='%s']`.

### DIFF
--- a/src/tree_schema.c
+++ b/src/tree_schema.c
@@ -716,6 +716,7 @@ lysc_path_until(const struct lysc_node *node, const struct lysc_node *parent, LY
 
         if ((pathtype == LYSC_PATH_DATA_PATTERN) && (iter->nodetype == LYS_LIST)) {
             char *predicates = NULL;
+
             key = NULL;
             while ((key = lys_getnext(key, iter, NULL, 0)) && lysc_is_key(key)) {
                 s = predicates;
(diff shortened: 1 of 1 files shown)


michalvasko:
You are right, but a simpler fix was possible.

snar:
Your fix is indeed simpler, but in some scenarios it produces incorrect results.
For example, with scheme 
```
list l {
    key "k l m";
    leaf k {}
    leaf l {}
    leaf m {}
    leaf n {}
}
```
path to `/l/n` is computed as `/l/n[k='%s'][l='%s'][m='%s']` instead of `/l[k='%s'][l='%s'][m='%s']/n`. Root cause: path can be already non-empty when adding predicates (with `/n` in this example), so you can't just add predicate to the end of it.

michalvasko:
Okay, so please add such a test, you can modify the one I added.

michalvasko:
Please resolve the conflicts so basically make the change again on top of the current `devel`.

michalvasko:
Right, sorry, not being able to rebase confused me. Thanks.

Issue #2220, opened 2024-04-05
Versions don't match anymore


choppsv1:
A new tagged release was made where the SO version was move from 2 to 3; however, the tagged project version appears to remain at 2.

Having the versions diverge like this just makes things really complex and hard for projects that are trying to use libyang. Why shouldn't the project version number also increment to indicate the non-backward compat changes as well? Packaging should also change from libyang2 to libyang3, I would think, if it hasn't.

michalvasko:
Yes, SO version has increased to 3 (because it had to) but project version only to 2.2 (from 2.1) because the changes were not that significant. Why is that a problem? The packages should use the major SO version as suffix and the included scripts do, I also know about other similar packages, which have a completely unrelated version and SO version, never had any issues with that.

choppsv1:
Who is the consumer of this 2.2 version string (i.e., who is using or cares about the somewhat arbitrary "not that significant" information)? Users and project owners care about whether things work or not, and the major version update informs on that. Also, if you look at pkg-config tests in autoconf I think they normally use the version string you are not changing the major of.

Jafaral:
SO version is now 3, that is a significant change, even if it were a a single code line change. A major version number pump is cheap. Why not just do that and avoid the confusion? It is not like you are going to run out of digits to use for major numbers any time soon.

Jafaral:
from https://semver.org/ :

Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes
MINOR version when you add functionality in a backward compatible manner
PATCH version when you make backward compatible bug fixes

Jafaral:
Of course, Libyang as a project can just ignore all of that and leave it to downstream projects, package maintainers, and system integrator to sort out the dependencies and give libyang whatever version number they want. Or they can from the start adhere to common versioning practices and make everyone's life easier.

michalvasko:
The motivation to use project and SO version is to have a version for both use-cases. To learn the details of API/ABI compatibility between versions and to learn the general significance of changes, respectively.

I have looked at the libraries installed on my system and all except one (`libxml2`) had a completely different SO version (in the name of the library file) than the version of the package they were in. So I am confused why we are even discussing this and why it should be a problem.

vjardin:
[...]

PR #2259, opened 2024-06-26
src/version.h.in: Define project version

davischw:
Make project version string and numbers easily accessible to downstream projects during buid by defining them in the version header file.

This is an attempt to solve issue #2220 by brute force.


michalvasko:
No problem, the project version can be in the header, too.

PR #2260, opened 2024-06-26
src/version.[ch]: Add version information API calls

davischw:
Add shared object library and project version information API calls. This will enable downstream projects to compare the library versions of the one installed on the system (via API calls) and the one linked against (via defines in 'version.h' during build) at runtime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,6 +159,7 @@ set(libsrc
     src/xml.c
     src/xpath.c
     src/validation.c
+    src/version.c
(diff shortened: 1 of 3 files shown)


michalvasko (changes requested):
Okay, I suppose there is no issue with having the version in the API as well. But I am not sure so many functions and a structure is needed and would prefer to reduce the number of public symbols. Maybe just make the 2 structures public?

michalvasko:
I want to make a new release so I adjusted it manually.

PR #2266, opened 2024-07-06
fix parsing of numbers within XML-encoded anyxml

jktjkt:
This was caught by the libyang-cpp test suite. On Windows/MSVC, an anyxml snippet that was parsed from this XML:
```xml
<ax xmlns="http://example.com/coze"><x>1</x><x>2</x><x>3</x></ax>
```
...would get printed like this:
```json
{"example-schema:ax":{"x":["1","2","3"]}}
```
...while the correct form is:
```json
{"example-schema:ax":{"x":[1,2,3]}}
```
This was caused by 21eaa39f9a, a change in `lydxml_get_hints_opaq` which probably tried to silence a compiler warning. The code wrongly assumed that the C `long` will be able to hold `UINT32_MAX`, which is not the case on Windows where `long` has the same range as `int32_t`. As a result, when the XML is parsed, these numeric values are compared against -1 on Windows, and against 4294967295 on many other platforms, and therefore any positive number was tagged with the hint of "this needs a 64bit serialization" on Windows, which in turn triggers printing these values as quoted string in JSON.

The fix simply makes sure that we never perform a narrowing conversion from a constant. I was looking for a way of using explicitly sized types (i.e., int64_t in this case), but there's only `strtoll` in C. The alternative would be `sscanf(... SCNd64 ...)`, but the semantics is subtly different when it comes to whitespace.

The bisecting process was non-trivial for me because the workflow which bypasses CI introduced a test regression (fixed by 7ebd607ee) and a temporary Windows breakage (fixed by a1ecc1695). Instead of a simple `git bisect`, this required some manual patch backporting.

### DIFF
--- a/src/parser_xml.c
+++ b/src/parser_xml.c
@@ -441,7 +441,9 @@ lydxml_get_hints_opaq(const char *name, size_t name_len, const char *value, size
 {
     struct lyd_node_opaq *opaq;
     char *ptr;
-    long num;
+    /* this needs to be at least 64bit, and it "should not" be an explicit int64_t
+     * because the code calls strtoll later on, which "might" return a bigger type */
+    long long num;
 
     *hints = 0;
     *anchor = NULL;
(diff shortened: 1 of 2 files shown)


michalvasko:
Last `devel` CI run passed on WIN but this PR does not for some reason, any ideas?

jktjkt:
That's breakage due to commit e55443f5ce6c3fc0a4cbb0b00bf3930a5802c501; I see that 420979e2f9cf7d3fa677ba0b51b52102e3e22f26 fixes that. Let me rerun the build.

jktjkt:
...and I do not see a button to do that :(, maybe you as the project owner can either rebase this or somehow make the job unstuck? It's been listed as "queued" for me for 18 hours.

michalvasko:
Never mind that, macOS kept failing and now won't even start. I can restart the other checks, but I though I did already yesterday.

jktjkt:
macos 11 won't build ever again as the images were removed, https://github.com/actions/runner-images/issues/10097

michalvasko:
Right... not sure how the upgrade will go. As for the WIN checks, they still fail, no idea why.

michalvasko (approved):

PR #2268, opened 2024-07-16
CI: fix Mac OS build

jktjkt:
Some of these tests use `expect`, and the native version that's available in the CI images is old/different/whatever. Let's use "the usual one" which is available through Homebrew.


michalvasko:
Nice, thanks.

michalvasko (approved):

PR #2297, opened 2024-09-07
tree data new BUGFIX correctly mark lyd_new_term_bin definition

gotthardp:
The `lyd_new_term_bin` is declared as an API function using `LIBYANG_API_DECL`.
Its definition however also uses `LIBYANG_API_DECL`, whereas I believe it should use `LIBYANG_API_DEF` like all the other functions.

```
/** Exporting symbols to a shared library and importing back afterwards
 *
 * - use LIBYANG_API_DECL to mark a declaration in the public header
 * - use LIBYANG_API_DEF to mark a definition (in the source code for the actual implementaiton)
 * */
```


michalvasko (approved):

michalvasko:
Right, thanks.

PR #2303, opened 2024-09-27
diff BUGFIX free diff if error occured

Humblesaw:


michalvasko (changes requested):
Formatting, tabs used.

PR #2344, opened 2025-02-06
data: option to allow json int/bool as strings

bradh352:
Prior to v1.0.212 the default behavior was to allow numbers
and boolean values to be in quotes, which is technically a
violation of the spec.

This adds a new `LYD_PARSE_JSON_STRING_DATATYPES` parse option
which will restore the prior behavior when enabled.

SONiC is using v1.0.73 currently and has a large installed base which
may be in violation of the new behavior, so adding such a flag is
required for this usecase.

Signed-off-by: Brad House (@bradh352)

### DIFF
--- a/src/parser_data.h
+++ b/src/parser_data.h
@@ -179,7 +179,10 @@ struct ly_in;
 #define LYD_PARSE_JSON_NULL 0x4000000       /**< Allow using JSON empty value 'null' within JSON input, such nodes are
                                                  silently skipped and treated as non-existent. By default, such values
                                                  are invalid. */
-
+#define LYD_PARSE_JSON_STRING_DATATYPES 0x8000000 /*!**< By default, JSON data values are validated to be in the proper format.
+                                                        For instance numbers are expected to not be enclosed in quotes, nor
+                                                        are boolean values.  Setting this option will disable this
+                                                        validation.  Prior to v1.0.212 this was the default behavior. */
 #define LYD_PARSE_OPTS_MASK 0xFFFF0000      /**< Mask for all the LYD_PARSE_ options. */
 
 /** @} dataparseroptions */
--- a/src/parser_data.h
+++ b/src/parser_data.h
@@ -179,10 +179,10 @@ struct ly_in;
 #define LYD_PARSE_JSON_NULL 0x4000000       /**< Allow using JSON empty value 'null' within JSON input, such nodes are
                                                  silently skipped and treated as non-existent. By default, such values
                                                  are invalid. */
-#define LYD_PARSE_JSON_STRING_DATATYPES 0x8000000 /*!**< By default, JSON data values are validated to be in the proper format.
-                                                        For instance numbers are expected to not be enclosed in quotes, nor
-                                                        are boolean values.  Setting this option will disable this
-                                                        validation.  Prior to v1.0.212 this was the default behavior. */
+#define LYD_PARSE_JSON_STRING_DATATYPES 0x8000000 /**< By default, JSON data values are expected to be in the correct
+                                                       format according to RFC 7951 based on their type. Using this
+                                                       option the validation can be softened to accept boolean and
+                                                       number type values enclosed in quotes. */
 #define LYD_PARSE_OPTS_MASK 0xFFFF0000      /**< Mask for all the LYD_PARSE_ options. */
 
 /** @} dataparseroptions */
--- a/src/parser_json.c
+++ b/src/parser_json.c
@@ -340,7 +340,7 @@ cleanup:
 /**
  * @brief Get the hint for the data type parsers according to the current JSON parser context.
  *
- * @param[in] jsonctx JSON parser context. The context is supposed to be on a value.
+ * @param[in] lydctx JSON data parser context.
  * @param[in,out] status Pointer to the current context status,
  * in some circumstances the function manipulates with the context so the status is updated.
  * @param[out] type_hint_p Pointer to the variable to store the result.
@@ -348,8 +348,9 @@ cleanup:
  * @return LY_EINVAL in case of invalid context status not referring to a value.
  */
 static LY_ERR
-lydjson_value_type_hint(struct lyjson_ctx *jsonctx, enum LYJSON_PARSER_STATUS *status_p, uint32_t *type_hint_p)
+lydjson_value_type_hint(struct lyd_json_ctx *lydctx, enum LYJSON_PARSER_STATUS *status_p, uint32_t *type_hint_p)
 {
+    struct lyjson_ctx *jsonctx = lydctx->jsonctx;
     *type_hint_p = 0;
 
     if (*status_p == LYJSON_ARRAY) {
@@ -383,6 +384,10 @@ lydjson_value_type_hint(struct lyjson_ctx *jsonctx, enum LYJSON_PARSER_STATUS *s
         return LY_EINVAL;
     }
 
+    if (lydctx->parse_opts & LYD_PARSE_JSON_STRING_DATATYPES) {
+        *type_hint_p |= LYD_VALHINT_STRING_DATATYPES;
+    }
+
     return LY_SUCCESS;
 }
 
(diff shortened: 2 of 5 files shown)


michalvasko:
Not sure if a project not working correctly justifies supporting such invalid use-cases but I suppose adding a flag does not break anything. It just seems it should be a JSON parsing flag instead of a context flag.

bradh352:
> Not sure if a project not working correctly justifies supporting such invalid use-cases but I suppose adding a flag does not break anything. It just seems it should be a JSON parsing flag instead of a context flag.

Yeah, I wasn't around when the initial implementation was done, but since libyang wasn't validating in v1 I guess assumptions were made.

Regarding it being a parser option, Ok, sure, it wasn't obvious to me how the code path was setting flags, but I see now that the code is casting between lyd_ctx and lyjson_ctx (which wasn't immediately obvious), so I would be able to pass down the flags there.  

So would this work? :
 * Create a new `LYD_PARSE_JSON_LOOSE_DATATYPES` value
 * Create a new `LYD_VALHINT_JSON_LOOSE_DATATYPES`
 * Modify `lydjson_value_type_hint()` to also take `lyd_json_ctx`
 * In `lydjson_value_type_hint()` if `lyd_son_ctx->parse_opts & LYD_PARSE_JSON_LOOSE_DATATYPES` set additional hint of `LYD_VALHINT_JSON_LOOSE_DATATYPES`
 * In `lyplg_type_check_hints()` use the new hint to determine if strict type enforcement is on

If that would be acceptable to you then I'll rework this patch.

I'd really like to solidify the plan that will be accepted here since we're working on upstreaming libyang3 in SONiC:
https://github.com/sonic-net/sonic-buildimage/pull/21679

michalvasko:
I do not think you really need `LYD_VALHINT_JSON_LOOSE_DATATYPES` but rather something like `LYPLG_TYPE_STORE_IGNORE_VALHINT` to pass this information to plugin types. `LYD_PARSE_JSON_LOOSE_DATATYPES` should be fine.

bradh352:
> I do not think you really need `LYD_VALHINT_JSON_LOOSE_DATATYPES` but rather something like `LYPLG_TYPE_STORE_IGNORE_VALHINT` to pass this information to plugin types. `LYD_PARSE_JSON_LOOSE_DATATYPES` should be fine.

I don't know if completely ignoring the valhint is good, I was leaving in place validation of if a string value wasn't enclosed in quotes since I haven't seen an instance of that, and truthfully I didn't check if v1 allowed that or not.  But I could go either way as I don't know if it would cause issues if we disabled that validation too ... the main issues that I see all over the place is SONiC expects `"1234" == 1234`, and `"true" == true`

bradh352:
> Fine, then call it accordingly. Something like `LYD_PARSE_JSON_STRING_DATATYPES` and `LYPLG_TYPE_STORE_ALLOW_STRING_VALHINT`.

please look at the PR now, I reworked it.  i can rename the options/flags whatever you want, just want to make sure this looks like I'm on the right track.  I added test cases too.

bradh352:
also I added lydctx to some functions but the json ctx was already there ... i'm assuming its probably better to extract the json ctx from the lydctx in those functions?  I couldn't tell but it looked like the json ctx might get manipulated and wasn't sure of the implications of that.  just let me know if I should instead be replacing jsonctx in those functions with lydctx and then doing `lydctx->jsonctx`

michalvasko:
Of course, you only need `lydctx`, since `jsonctx` is a member. Also, please change the base branch to `devel`, we do not accept PRs into `master`. If you then rename the values, it should be fine.

bradh352:
@michalvasko
> Of course, you only need `lydctx`, since `jsonctx` is a member. Also, please change the base branch to `devel`, we do not accept PRs into `master`. If you then rename the values, it should be fine.

done with all your suggestions...

src/parser_data.h
```diff
@@ -179,7 +179,10 @@ struct ly_in;
 #define LYD_PARSE_JSON_NULL 0x4000000       /**< Allow using JSON empty value 'null' within JSON input, such nodes are
                                                  silently skipped and treated as non-existent. By default, such values
                                                  are invalid. */
-
+#define LYD_PARSE_JSON_STRING_DATATYPES 0x8000000 /*!**< By default, JSON data values are validated to be in the proper format.
```
michalvasko, line 182: Please use the common `/**<` starting tag. Also, the description could be more exact such as:
```
By default, JSON data values are expected to be in the correct format according to RFC 7951 based on their type. Using this option the validation can be softened to accept boolean and number type values enclosed in quotes.
```
No need to mention *libyang* v1, it is obsolete for a long time now.

src/parser_json.c
```diff
@@ -340,16 +340,17 @@ lydjson_get_snode(struct lyd_json_ctx *lydctx, ly_bool is_attr, const char *pref
...
  * in some circumstances the function manipulates with the context so the status is updated.
  * @param[out] type_hint_p Pointer to the variable to store the result.
  * @return LY_SUCCESS in case of success.
  * @return LY_EINVAL in case of invalid context status not referring to a value.
  */
 static LY_ERR
-lydjson_value_type_hint(struct lyjson_ctx *jsonctx, enum LYJSON_PARSER_STATUS *status_p, uint32_t *type_hint_p)
+lydjson_value_type_hint(struct lyd_json_ctx *lydctx, enum LYJSON_PARSER_STATUS *status_p, uint32_t *type_hint_p)
 {
+    struct lyjson_ctx *jsonctx = lydctx->jsonctx;
```
michalvasko, line 353: Please always leave an empty line after var declarations.

michalvasko (changes requested):
Thanks, looks fine, just minor issues.

src/parser_data.h
bradh352: the starting tag was a typo, fixed

src/parser_json.c
bradh352: fixed

michalvasko (approved):

src/plugins_types.c
```diff
@@ -685,7 +685,8 @@ lyplg_type_check_hints(uint32_t hints, const char *value, size_t value_len, LY_D
     case LY_TYPE_INT32:
         LY_CHECK_ARG_RET(NULL, base, LY_EINVAL);
 
-        if (!(hints & (LYD_VALHINT_DECNUM | LYD_VALHINT_OCTNUM | LYD_VALHINT_HEXNUM))) {
+        if (!(hints & (LYD_VALHINT_DECNUM | LYD_VALHINT_OCTNUM | LYD_VALHINT_HEXNUM)) &&
+            !(hints & LYD_VALHINT_STRING_DATATYPES)) {
```
michalvasko, line 689: Coding style issues, we use double indent when continuing the previous line.

michalvasko (changes requested):

src/plugins_types.c
bradh352: fixed

src/plugins_types.c
```diff
@@ -695,7 +696,8 @@ lyplg_type_check_hints(uint32_t hints, const char *value, size_t value_len, LY_D
     case LY_TYPE_INT64:
         LY_CHECK_ARG_RET(NULL, base, LY_EINVAL);
 
-        if (!(hints & LYD_VALHINT_NUM64)) {
+        if (!(hints & LYD_VALHINT_NUM64) &&
+            !(hints & LYD_VALHINT_STRING_DATATYPES)) {
```
michalvasko, line 700: Another indentation problem.

src/plugins_types.c
```diff
@@ -714,7 +716,8 @@ lyplg_type_check_hints(uint32_t hints, const char *value, size_t value_len, LY_D
         }
         break;
     case LY_TYPE_BOOL:
-        if (!(hints & LYD_VALHINT_BOOLEAN)) {
+        if (!(hints & LYD_VALHINT_BOOLEAN) &&
+            !(hints & LYD_VALHINT_STRING_DATATYPES)) {
```
michalvasko, line 720: Another indentation problem.

src/plugins_types.c
bradh352: Ever think about adding a .clang-format to help people with code formatting?  Its gotten pretty good these days, we use it in c-ares :)

michalvasko (changes requested):
Sorry for not being explicit, there were more instances of the indentation problem.

src/plugins_types.c
michalvasko: We tried that but strongly prefer `uncrustify`, we could not configure `clang` for our coding style. But you can use it to format the auotmatically, even directly supported by our projects, just run `make format`.

src/plugins_types.c
bradh352: fixed

michalvasko (approved):

src/plugins_types.c
bradh352: ah, good to know, didn't see that

and yeah, clang-format works great on most code styles, but if you have a deviation you have to submit a patch to clang and add a toggle for that.  Actually, I recently upstreamed one for c-ares.

PR #2351, opened 2025-02-15
schema: helpers for tracking backlinks

bradh352:
In libyang v1, each schema node had a backlinks object that would allow a caller to quickly see if the node was referenced by another node.  Since this functionality was removed, it takes quite a bit of code to accomplish the same thing, including full tree scans.

This adds these primary new helpers:
 * `lys_find_backlinks()` scan the entire schema tree (all modules) and locate any node that contains leafrefs (including in unions), possibly filtered to a single target node or ancestor thereof.
 * `lysc_node_find_lref_targets()` for a given node, return a set of leafref target nodes. This is an enhanced version of `lysc_node_lref_target()`, however this returns a set instead of a single node as it also returns targets associated with unions (which there may be more than one leafref).  This does not return recursive results, meaning if the node provided isn't a leaf or leaflist, it will not return any results.

It also adds a couple of new helpers used by the above new functions that may be found useful so they were made public:
 * `lysc_node_has_ancestor()` determine if the ancestor node exists as a parent, grandparent, etc within the node schema for the specified node.  This simply scans up the tree using `node->parent` looking for a match.
 * `lysc_type_lref_target()` similar to `lysc_node_lref_target()` except it can return the target node referenced in a leafref for a `struct lysc_type`.  This is used internally by `lysc_node_find_lref_targets()`.

This functionality was determined to be needed when porting SONiC from libyang v1 to v3.  SONiC uses libyang-python, and the API does not currently expose some of the underlying functions used by these helpers. Instead of simply modifying the libyang-python to add CFFI wrappers it was determined it would be cost prohibitive to support `lysc_module_dfs_full()` in python due to generating a Python SNode for each node in the tree, of which most nodes are not needed to be evaluated by Python.

It is not unlikely that other users may have the same need to track backlinks so adding these helpers would also benefit other users.

### DIFF
--- a/src/tree_schema.c
+++ b/src/tree_schema.c
@@ -684,6 +684,110 @@ cleanup:
     return snode;
 }
 
+
+typedef struct {
+    const struct lysc_node *match_node;
+    ly_bool match_ancestors;
+    struct ly_set *set;
+} lys_find_backlinks_t;
+
+static LY_ERR
+lys_find_backlinks_clb(struct lysc_node *node, void *data, ly_bool *dfs_continue)
+{
+    lys_find_backlinks_t *bldata = data;
+    LY_ERR ret = LY_SUCCESS;
+    struct ly_set *set = NULL;
+    size_t i;
+
+    (void)dfs_continue;
+
+    /* Not a node type we are interested in */
+    if (node->nodetype != LYS_LEAF && node->nodetype != LYS_LEAFLIST) {
+        return LY_SUCCESS;
+    }
+
+    /* Fetch leafrefs targets for comparison against our match node.  Even if we
+     * are going to throw them away, we still need a count to know if this has
+     * valid leafref targets*/
+    ret = lysc_node_find_lref_targets(node, &set);
+    if (ret == LY_ENOTFOUND) {
+        return LY_SUCCESS;
+    } else if (ret != LY_SUCCESS) {
+        goto cleanup;
+    }
+
+    /* If set contains no entries, don't add node */
+    if (set == NULL || set->count == 0) {
+        goto cleanup;
+    }
+
+    /* If we're not requiring a match of a target node, just add this node to
+     * the returned set */
+    if (bldata->match_node == NULL) {
+        ly_set_add(bldata->set, node, 1, NULL);
+        goto cleanup;
+    }
+
+    /* We are doing target matching, scan to see if this node should be added */
+    for (i=0; i<set->count; i++) {
+        if (bldata->match_ancestors) {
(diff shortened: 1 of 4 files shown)


bradh352:
I have not yet added test cases.  This is due to feedback on https://github.com/CESNET/libyang-python/pull/132 which suggested this functionality should be in the core libyang rather than implemented in libyang-python.  

I've tried to make the APIs as similar to others in libyang as possible.  I'm definitely willing to make changes, which is the primary reason I haven't added test cases, I wanted to ensure the public api was acceptable before doing so.

michalvasko:
You could have started with asking about this functionality because it has already been implemented, by a PR. Look at how [`LY_CTX_LEAFREF_LINKING`](https://netopeer.liberouter.org/doc/libyang/master/html/group__contextoptions.html#gac7e1e7fb181f0259946d3332ae600fd3) context flag works.

bradh352:
@michalvasko 

> You could have started with asking about this functionality because it has already been implemented, by a PR. Look at how [`LY_CTX_LEAFREF_LINKING`](https://netopeer.liberouter.org/doc/libyang/master/html/group__contextoptions.html#gac7e1e7fb181f0259946d3332ae600fd3) context flag works.

Its not immediately obvious to me how that context flag accomplishes the same thing.  Doesn't that strictly work on data not schema?  The routines I added were strictly operating on the schema.

bradh352:
`match_ancestors` in `lys_find_backlinks()` is a very strong requirement for what we are doing, without it we'd need to maintain external patches which is non-optimal.  I don't see the harm in this function having another argument like that, but if you want to turn it into a flags type argument instead for future possible capabilities, that's fine with me.  The only use-case in SONiC right now that utilizes it is through  libyang-python, but of course that brings up other issues since even less is exposed in python so we can't implement similar functionality there ... it really needs to be in C libyang.

Also, `lysc_node_find_lref_targets()` is needed to be public as well since it is needed to support the union usecase.  The other two functions don't really matter to me if they are public or not.

bradh352:
@michalvasko 

>  it will take considerable effort to get it to a state when it can be merged, not even talking about the coding style. I think it may even be faster if I just adjust the code but I can also just iteratively provide feedback, up to you.

I'm definitely ok if you want to take the code and adjust it directly if you think that is more efficient.  I think at this point with my last reply you understand the requirements in SONiC.  This PR is currently integrated as a PR into SONiC being reviewed, but it shouldn't be a problem to modify that SONiC PR with whatever you come up with as long as it meets the aforementioned needs.

michalvasko:
Please look at the PR #2352 I created and test it.

PR #2352, opened 2025-02-17
Leafref util functions

michalvasko:
Refs #2351

### DIFF
--- a/src/tree_schema.h
+++ b/src/tree_schema.h
@@ -1908,12 +1908,24 @@ LIBYANG_API_DECL struct lysc_must *lysc_node_musts(const struct lysc_node *node)
 LIBYANG_API_DECL struct lysc_when **lysc_node_when(const struct lysc_node *node);
 
 /**
- * @brief Get the target node of a leafref node.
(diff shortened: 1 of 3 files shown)


bradh352:
thanks, at first glance this looks basically the same so it should work.  I'll pull it into the SONiC PRs I've got so I can run the tests there to make sure all is good.  I'll report back. Thanks!

bradh352:
@michalvasko I can confirm this PR passes our test cases in SONiC.  Thanks!

bradh352:
I'm not familiar with your process, but it appears the libyang-python repo depends on commits in the libyang `master` branch only for running its tests.  Any idea when this PR (as well as PR #2344) might get merged into master so the libyang-python PRs can be evaluated (https://github.com/CESNET/libyang-python/pull/134 https://github.com/CESNET/libyang-python/pull/132)?

michalvasko:
The last release is still fairly new so it will likely be at least a few months.

bradh352:
Ok, I'll need to try to push @rjarry I guess to make sure the changes will be acceptable or if there are any requested APi changes in libyang-python when using the new APIs as I'll need to pull in PRs into SONiC as patches before they are accepted.

PR #2354, opened 2025-02-21
lyd_validate_minmax(): always return a valid data path

bradh352:
The current code never returns a data path, only a schema path for failures where the 'min' value is not reached.  Integrators may need to obtain both a LIST key as well as the leaf name to present proper error messages.

When we don't have a full valid data path, the code now will return the parent's data path with the current schema node name appended.

This fix was inspired by porting of SONiC's sonic-mgmt-common which is utilizing this feature which previously worked with libyang1.


bradh352:
guess I need to update test cases to look at the 'better' paths returned :)

michalvasko:
Fixed the path myself.

PR #2358, opened 2025-02-27
schema: add lysc_node_lref_xpath_targets() helper

### LINKED ISSUE #2357: lysc_node_lref_targets() insufficent (related to PR #2351/#2352)
So I've hit an issue that I hadn't noticed or expected and I need a solution.  I need to be able to extract the original leafref target xpath.

Right now when using the new `lysc_node_lref_targets()`, I get a set of schema node pointers which is insufficient as calling `lysc_path()` on the returned nodes effectively ignores any path key restrictions.

For a concrete example, I have a yang schema that says:
```
leaf bgp_asn {
  type leafref {
    path "/sbg:sonic-bgp-global/sbg:BGP_GLOBALS/sbg:BGP_GLOBALS_LIST[sbg:vrf_name=current()/../vrf_name]/sbg:local_asn";
  }
}
```

However when I call `lysc_path()` on the target returned from `lysc_node_lref_targets()`, I get:
```
/sonic-bgp-global:sonic-bgp-global/BGP_GLOBALS/BGP_GLOBALS_LIST/local_asn
```

I think I'm going to need another function of something like `lysc_node_lref_xpath_targets()` that instead of returning a set of target schema nodes can return a set of xpaths.  It looks like it would be easy to do, so I'll probably submit a PR  for that.

In fact, due to this unforeseen issue, likely `lysc_node_lref_targets()` doesn't serve any useful purpose to be a public function, its still needed for tracking backlinks internally for `lysc_node_lref_backlinks()`, however.  So since this hasn't made it into a public release, the PR I submit will probably move this to a static non-public symbol.

bradh352:
The lysc_node_lref_targets() helper returns a node in the schema, but if the xpath itself had added restrictions such as only matching certain list keys, then there was no way to retrieve that information.

This adds the appropriate helper, sharing as much code as possible with lysc_node_lref_targets() and also adds a test case to check for this.

Fixes #2357

### DIFF
--- a/src/tree_schema.h
+++ b/src/tree_schema.h
@@ -1928,6 +1928,18 @@ LIBYANG_API_DECL const struct lysc_node *lysc_node_lref_target(const struct lysc
 LIBYANG_API_DECL LY_ERR lysc_node_lref_targets(const struct lysc_node *node, struct ly_set **set);
 
 /**
+ * @brief Get the target xpath(s) of a leafref node or union node with leafrefs.
+ *
+ * @param[in] node Term node to use.
+ * @param[out] set Set with all the leafref target xpaths, may be empty if the node is a different type or the targets
+ * are not found.  The set is filled with NULL-terminated "const char *" strings which can be retrieved via the
+ * objs member.
+ * @return LY_SUCCESS on success.
+ * @return LY_ERR value on error.
+ */
+LIBYANG_API_DECL LY_ERR lysc_node_lref_xpath_targets(const struct lysc_node *node, struct ly_set **set);
+
+/**
  * @brief Get all the leafref (or union with leafrefs) nodes that target a specific node.
  *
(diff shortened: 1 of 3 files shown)


michalvasko:
As I mentioned in the linked issue, I would strongly prefer this helper function to be implemented in your project, has little use and hence redundant to be included in *libyang*.

bradh352:
I'm closing this one.  I ended up not using it directly in SONiC.  Given the fact that the paths returned aren't valid xpaths, it is of minimal use. For Golang I just implemented this directly as cgo code, then rewrote the xpaths to be valid.  For python a different approach is used there where it uses the original `lysc_node_lref_targets()` to look up the data paths that could be impacted, then scans all results doing a value check to limit to the actual matching nodes.  A bit more overhead ... but the old libyang1 code was actually doing something similar so its no worse.

PR #2360, opened 2025-02-28
validate: unions with leafrefs may return misleading error codes

### LINKED ISSUE #2359: non-existent leafref inheriting error-message and error-app-tag
This seems like a bug.  When a leaf is a leafref, but the validation fails because the target doesn't have a matching entry, for some reason it inherits the `error-message` and `error-app-tag` from the referenced schema.

A schema like this:
```
module sonic-port {
    namespace "http://github.com/Azure/sonic-port";
    prefix prt;
    container sonic-port {
        container PORT {
            list PORT_LIST {
                key "ifname";
                leaf ifname {
                    type string {
                        pattern "Ethernet([1-3][0-9]{3}|[1-9][0-9]{2}|[1-9][0-9]|[0-9])" {
                            error-message "Invalid interface name";
                            error-app-tag interface-name-invalid;
                        }
                    }
                }
            }
        }
    }
}
```
```
module sonic-portchannel {
    namespace "http://github.com/Azure/sonic-portchannel";
    prefix pch;
    container sonic-portchannel {
        container PORTCHANNEL {
            list PORTCHANNEL_LIST {
                key "name";
                leaf name {
                    type string {
                        pattern "PortChannel([1-3][0-9]{3}|[1-9][0-9]{2}|[1-9][0-9]|[0-9])";
                    }
                }
            }
        }
    }
}
```
```
module sonic-vlan {
    namespace "http://github.com/Azure/sonic-vlan";
    prefix svlan;
    import sonic-port {
        prefix prt;
    }
    import sonic-portchannel {
        prefix pch;
    }
    container sonic-vlan {
        container VLAN_MEMBER { 
            list VLAN_MEMBER_LIST { 
                key "name ifname";
                leaf name {
                    type string;
                }
                leaf ifname {
                    type union {
                        type leafref {
                            path "/prt:sonic-port/prt:PORT/prt:PORT_LIST/prt:ifname";
                        }
                        type leafref {
                            path "/pch:sonic-portchannel/pch:PORTCHANNEL/pch:PORTCHANNEL_LIST/pch:name";
                        }
                    }
                }
            }
        }
    }
}
```

Then with data like:
```
{  
  "sonic-vlan:sonic-vlan":{
    "VLAN_MEMBER": {
      "VLAN_MEMBER_LIST": [
          { "name":"test" }, { "ifname": "PortChannel4" }
      ]
    }
  },
  "sonic-port:sonic-port": {
    "PORT": {
      "PORT_LIST": [
        { "ifname": "Ethernet0" }
      ]
    }
  }
}
```

It might yield an error message like:
```
libyang[0]: Invalid union value "PortChannel4" - no matching subtype found:
    libyang 2 - leafref, version 1: [Error]Invalid interface name
    libyang 2 - leafref, version 1: Invalid leafref value "PortChannel4" - no target instance "/pch:sonic-portchannel/pch:PORTCHANNEL/pch:PORTCHANNEL_LIST/pch:name" with the same value
```

The first sub error has an `apptag` of ` interface-name-invalid` where the second has `instance-required`

That error message makes sense if the entry in PORT_LIST didn't match the required pattern, but when its a leafref in a union and it doesn't match the pattern for it to be eligible, that error doesn't make sense.

If the `error-message` and `error-app-tag` were not in the schema, then it would return the expected error messages indicating no target instance for each leafref in the union and properly return `instance-required` as the apptag.

I haven't fully researched this to see if it is specific to leafrefs in unions or not.

bradh352:
When a leafref is not found in a union, it should not return an error indicating the value for the leafref wouldn't match the restrictions for a target but instead just say the leafref target wasn't found.

The only data that should indicate the error message for data type restrictions is data associated with the target itself, not leafrefs.

Fixes #2359

### DIFF
--- a/src/plugins_types/union.c
+++ b/src/plugins_types/union.c
@@ -159,25 +159,39 @@ lyb_parse_union(const void *lyb_data, size_t lyb_data_len, uint32_t *type_idx, c
  * @param[in] type leafref type used to extract target path
  * @param[in] value value attempting to be stored
  * @param[in] value_len Length of value that was attempted to be stored.
+ * @return LY_ERR value. Only possible errors are LY_SUCCESS and LY_EMEM.
  */
-static void
+static LY_ERR
 union_update_lref_err(struct ly_err_item *err, const struct lysc_type *type, const void *value, size_t value_len)
 {
     const struct lysc_type_leafref *lref;
     char *valstr = NULL;
+    int msg_len;
 
     if ((err == NULL) || (type->basetype != LY_TYPE_LEAFREF)) {
-        return;
+        return LY_SUCCESS;
     }
 
     lref = (const struct lysc_type_leafref *)type;
     free(err->apptag);
     err->apptag = strdup("instance-required");
+    if (err->apptag == NULL) {
+        return LY_EMEM;
+    }
 
     free(err->msg);
+    err->msg = NULL;
     valstr = strndup((const char *)value, value_len);
-    asprintf(&err->msg, LY_ERRMSG_NOLREF_VAL, valstr, lyxp_get_expr(lref->path));
+    if (valstr == NULL) {
+        return LY_EMEM;
+    }
+
+    msg_len = asprintf(&err->msg, LY_ERRMSG_NOLREF_VAL, valstr, lyxp_get_expr(lref->path));
     free(valstr);
+    if (msg_len == -1) {
+        return LY_EMEM;
+    }
+    return LY_SUCCESS;
 }
 
 /**
@@ -200,7 +214,7 @@ union_store_type(const struct ly_ctx *ctx, struct lysc_type_union *type_u, uint3
         uint32_t options, ly_bool validate, const struct lyd_node *ctx_node, const struct lyd_node *tree,
         struct lys_glob_unres *unres, struct ly_err_item **err)
 {
-    LY_ERR rc = LY_SUCCESS;
+    LY_ERR urc, rc = LY_SUCCESS;
     struct lysc_type *type = type_u->types[type_idx];
     const void *value = NULL;
     size_t value_len = 0;
@@ -222,7 +236,10 @@ union_store_type(const struct ly_ctx *ctx, struct lysc_type_union *type_u, uint3
                 memset(&subvalue->value, 0, sizeof subvalue->value);
 
                 /* if this is a leafref, lets make sure we propagate the appropriate error, and not a type validation failure */
-                union_update_lref_err(*err, type_u->types[ti], value, value_len);
+                urc = union_update_lref_err(*err, type_u->types[ti], value, value_len);
+                if (urc != LY_SUCCESS) {
+                    rc = urc;
+                }
                 goto cleanup;
             }
 
@@ -260,7 +277,10 @@ union_store_type(const struct ly_ctx *ctx, struct lysc_type_union *type_u, uint3
         memset(&subvalue->value, 0, sizeof subvalue->value);
 
         /* if this is a leafref, lets make sure we propagate the appropriate error, and not a type validation failure */
-        union_update_lref_err(*err, type, value, value_len);
+        urc = union_update_lref_err(*err, type, value, value_len);
+        if (urc != LY_SUCCESS) {
+            rc = urc;
+        }
         goto cleanup;
     }
 
--- a/src/plugins_types/union.c
+++ b/src/plugins_types/union.c
@@ -153,14 +153,13 @@ lyb_parse_union(const void *lyb_data, size_t lyb_data_len, uint32_t *type_idx, c
 /**
  * @brief For leafref failures, ensure the appropriate error is propagated, not a type validation failure.
  *
- * @param[in] ctx_node Context node for prefix resolution.
  * @param[in,out] err Error record to be updated
  * @param[in] type leafref type used to extract target path
  * @param[in] value value attempting to be stored
  * @param[in] value_len Length of value that was attempted to be stored.
  */
 static void
-union_update_lref_err(const struct lyd_node *ctx_node, struct ly_err_item *err, const struct lysc_type *type, const void *value, size_t value_len)
+union_update_lref_err(struct ly_err_item *err, const struct lysc_type *type, const void *value, size_t value_len)
 {
     const struct lysc_type_leafref *lref;
     char *valstr = NULL;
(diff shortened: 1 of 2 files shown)


src/plugins_types/union.c
```diff
@@ -228,6 +256,11 @@ union_store_type(const struct ly_ctx *ctx, struct lysc_type_union *type_u, uint3
     if ((rc != LY_SUCCESS) && (rc != LY_EINCOMPLETE)) {
         /* clear any leftover/freed garbage */
         memset(&subvalue->value, 0, sizeof subvalue->value);
+
+        /* if this is a leafref, lets make sure we propagate the appropriate error, and not a type validation failure */
+        if (type->basetype == LY_TYPE_LEAFREF) {
```
michalvasko, line 261: Redundant check, since it is in the function.

src/plugins_types/union.c
```diff
@@ -150,6 +150,31 @@ lyb_parse_union(const void *lyb_data, size_t lyb_data_len, uint32_t *type_idx, c
...
+    }
+
+    lref = (struct lysc_type_leafref *)type;
+    free(err->apptag);
+    err->apptag = strdup("instance-required");
+
+    free(err->msg);
+
+    if (lyd_get_value(ctx_node) != NULL) {
+        valstr = strdup(lyd_get_value(ctx_node));
```
michalvasko, line 170: Only `value` with `value_len` are being stored, why would you use the value in the node?

src/plugins_types/union.c
```diff
@@ -150,6 +150,31 @@ lyb_parse_union(const void *lyb_data, size_t lyb_data_len, uint32_t *type_idx, c
     }
 }
 
+static void
+union_lref_error_rewrite(const struct lyd_node *ctx_node, struct ly_err_item *err, struct lysc_type *type, const void *value, size_t value_len)
```
michalvasko, line 154: I would prefer a bit different name although that is a small issue. Maybe `union_update_lref_err`. But the doxygen is missing and required.

michalvasko (changes requested):
Fine, after fixing the issues this can be merged.

src/plugins_types/union.c
bradh352: done

src/plugins_types/union.c
bradh352: Because without that logic, I get a test failure and a value I can't understand:
```
Could not run test: "Invalid LYB union value - no matching subtype found:
    libyang 2 - leafref, version 1: Invalid leafref value "one" - no target instance "../../a/name" with the same value.
    libyang 2 - leafref, version 1: Invalid leafref value "ins" - no target instance "../../b/name" with the same value.
" != "Invalid LYB union value - no matching subtype found:
    libyang 2 - leafref, version 1: Invalid leafref value "one" - no target instance "../../a/name" with the same value.
    libyang 2 - leafref, version 1: Invalid leafref value "one" - no target instance "../../b/name" with the same value.
"
[   LINE   ] --- /home/brad/libyang/tests/utests/types/union.c:296: error: Failure!ly_err_last(current_utest_context->ctx)
[   LINE   ] --- /home/brad/libyang/tests/utests/types/../utests.h:1350: error: Failure!Test teardown failed
```

Notice `ins` instead of `one`.

src/plugins_types/union.c
bradh352: done

src/plugins_types/union.c
michalvasko: Obviously, not a reason to write such a code. It took me a while to find the reason, it being that dynamic `value` is freed after it is stored, whether successfully or not. But then the freed memory is (always) reused for `err->apptag`, so while freed memory is used, the content is "valid" making it extremely confusing. Try this [patch](https://github.com/user-attachments/files/19052650/value_fix.patch.txt) and then you can update the code properly.

src/plugins_types/union.c
```diff
@@ -150,6 +150,40 @@ lyb_parse_union(const void *lyb_data, size_t lyb_data_len, uint32_t *type_idx, c
...
+ * @brief For leafref failures, ensure the appropriate error is propagated, not a type validation failure.
+ *
+ * @param[in] ctx_node Context node for prefix resolution.
+ * @param[in,out] err Error record to be updated
+ * @param[in] type leafref type used to extract target path
+ * @param[in] value value attempting to be stored
+ * @param[in] value_len Length of value that was attempted to be stored.
+ */
+static void
+union_update_lref_err(const struct lyd_node *ctx_node, struct ly_err_item *err, const struct lysc_type *type, const void *value, size_t value_len)
```
michalvasko, line 163: Keep 120 chars limit, break after `type`.

src/plugins_types/union.c
```diff
@@ -150,6 +150,40 @@ lyb_parse_union(const void *lyb_data, size_t lyb_data_len, uint32_t *type_idx, c
     }
 }
 
+/**
+ * @brief For leafref failures, ensure the appropriate error is propagated, not a type validation failure.
```
michalvasko, line 154: Please also mention the RFC 7950 section that defines the specific error and the error-app-tag.

src/plugins_types/union.c
```diff
@@ -150,6 +150,40 @@ lyb_parse_union(const void *lyb_data, size_t lyb_data_len, uint32_t *type_idx, c
...
+    err->apptag = strdup("instance-required");
+
+    free(err->msg);
+
+    if (lyd_get_value(ctx_node) != NULL) {
+        valstr = strdup(lyd_get_value(ctx_node));
+    } else {
+        valstr = strndup((const char *)value, value_len);
+    }
+    asprintf(&err->msg, LY_ERRMSG_NOLREF_VAL, valstr, lyxp_get_expr(lref->path));
```
michalvasko, line 183: Missing memory allocation failure checks.

michalvasko (changes requested):

src/plugins_types/union.c
bradh352: Ok, I didn't know the reason for that.  I just figured it was a junk value.  It wasn't clear ownership was being transferred.  I incorporated your patch.

src/plugins_types/union.c
bradh352: didn't know that requirement, uncrustify didn't seem to auto-fix that and there are other lines > 120 that exist in this file.  Regardless, the parameter list is shorter now after the other change so this is resolved.

src/plugins_types/union.c
bradh352: done

src/plugins_types/union.c
bradh352: done. 

As a side note, I'm not a fan of OOM error checking in general.  On linux at least, it just doesn't work since linux uses an optimistic memory allocation strategy which will return success even if there is no memory available and the kernel will just kill your application when you try to use it.  There's probably exceptions where linux bypasses this optimism on ***huge*** allocations where its relevant (which isn't relevant to libyang), but in general it adds a lot of code that does nothing.

Plus all modern memory-safe languages just abort() on OOM, and likely not everything is C these days....

I even wrote a blog post on it : https://bradhouse.dev/2023/11/how-to-handle-out-of-memory-oom-in-c/

michalvasko: I am neither but my previous boss was and so we kept adding the checks everywhere. One argument may be that our libraries are not used only on Linux but on other platforms, some running usually on a very limited HW.

michalvasko (approved):

PR #2361, opened 2025-02-28
union: remove assert in sort

bradh352:
Getting assertion in lyplg_type_sort_union().  This assertion appears invalid as it comes up in test cases in SONiC but when removed everything passes.


bradh352:
> I believe the assert is relevant and I would guess the issue is caused by leafrefs, not sure why it did not come up sooner. Try using the following code for the loop and see whether that fixes the assert.

You're right, I should have spent more time evaluating this.  Your patch works properly. All test cases in both libyang and sonic-mgmt-common pass with your change.   I've updated this PR to use your change.

michalvasko (approved):

PR #2362, opened 2025-02-28
validation: Add LYD_VALIDATE_NOEXTDEPS to bypass leafref/when/must

### LINKED ISSUE #2355: `LYD_OPT_NOEXTDEPS` no longer exists, alternatives?
The transition manual from v1 to v2 says:
> Many validation flags were removed because they became obsolete (LYD_OPT_DESTRUCT, LYD_OPT_WHENAUTODEL, LYD_OPT_NOEXTDEPS, LYD_OPT_DATA_NO_YANGLIB, LYD_OPT_VAL_DIFF) or we consider them redundant (LYD_OPT_OBSOLETE, LYD_OPT_NOSIBLINGS, LYD_OPT_DATA_ADD_YANGLIB).

In sonic-mgmt-common, it is using `LYD_OPT_NOEXTDEPS` for partial tree validations.  This validation still wants to check for things like valid data for constraints/patterns/length, min/max, mandatory, and so forth ... but does not want to check outside of the current module for leafrefs/must/when.  The code does not have the full data tree available at this point when calling lyd_validate_all(), hence the use of this flag.

Is there an alternative approach that can be used?  Or would you accept patches to reimplement a flag like `LYD_VALIDATE_NOEXTDEPS` to restore the prior behavior?

bradh352:
In libyang v1, there was an LYD_OPT_NOEXTDEPS flag.  This was removed, and this patch re-adds a flag with similar functionality.

The gNMI/RESTCONF portion of the SONiC port is currently relying on this patch:
https://github.com/sonic-net/sonic-mgmt-common/pull/157

If this isn't accepted upstream, we'll need to maintain this patch separately.

Fixes #2355

### DIFF
--- a/src/parser_data.h
+++ b/src/parser_data.h
@@ -222,7 +222,8 @@ struct ly_in;
 #define LYD_VALIDATE_NOT_FINAL 0x0020       /**< Skip final validation tasks that require for all the data nodes to
                                                  either exist or not, based on the YANG constraints. Once the data
                                                  satisfy this requirement, the final validation should be performed. */
-
+#define LYD_VALIDATE_NOEXTDEPS 0x0040       /**< Allow external dependencies (external leafrefs, instance-identifiers,
+                                                 must, and when) to not be resolved/satisfied during validation. */
 #define LYD_VALIDATE_OPTS_MASK  0x0000FFFF  /**< Mask for all the LYD_VALIDATE_* options. */
 
 /** @} datavalidationoptions */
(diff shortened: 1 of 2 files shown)


michalvasko:
We have discussed this flag and as I have said before, we removed it intentionally and we do not want to support it because it is not reliable and, what is worse, leads to bad design of applications.

bradh352:
understood, just one of those policies to always submit upstream :)  We'll maintain this patch separately in SONiC until the code that relies on this can be rearchitected.

PR #2363, opened 2025-03-02
avoid repeated tzset() calls

irfanHaslanded:
The  `tzset()` function  initializes  the tzname variable from the
TZ environment variable.

There doesn't seem to be any valid usecase of calling this repeatedly.
It is an overhead that can be avoided.

### DIFF
--- a/src/tree_data_common.c
+++ b/src/tree_data_common.c
@@ -1620,6 +1620,21 @@ ly_time_tz_offset(void)
     return ly_time_tz_offset_at(time(NULL));
 }
 
+/**
+ * @brief Call tzset() if not already called by this process.
+ */
+static void
+ly_tzset_once(void)
+{
+    static int ly_tzset_called = 0;
+
+    if (ly_tzset_called) {
+        return;
+    }
+    tzset();
+    ly_tzset_called = 1;
+}
+
 LIBYANG_API_DEF int
 ly_time_tz_offset_at(time_t time)
 {
@@ -1627,7 +1642,7 @@ ly_time_tz_offset_at(time_t time)
     int result = 0;
 
     /* init timezone */
-    tzset();
+    ly_tzset_once();
 
     /* get local and UTC time */
     localtime_r(&time, &tm_local);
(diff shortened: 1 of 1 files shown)


jktjkt:
Just remove that explicit call to `tzset()`. In both of these places where `tzset()` is invoked, the very next function that gets called is `locatime_r` which is [documented](https://pubs.opengroup.org/onlinepubs/009604499/functions/localtime.html) to also perform the `tzset()`'s side effects.

michalvasko:
From my man page:
```
       The  localtime() function converts the calendar time timep to broken-down time representation, expressed relative to the user's specified timezone.  The function also
       sets the external variables tzname, timezone, and daylight as if it called tzset(3).  The return value points to a statically allocated struct which  might  be  over‐
       written  by subsequent calls to any of the date and time functions.  The localtime_r() function does the same, but stores the data in a user-supplied struct.  It need
       not set tzname, timezone, and daylight.
```

irfanHaslanded:
> Just remove that explicit call to `tzset()`. In both of these places where `tzset()` is invoked, the very next function that gets called is `locatime_r` which is [documented](https://pubs.opengroup.org/onlinepubs/009604499/functions/localtime.html) to also perform the `tzset()`'s side effects.

Unfortunately, this doesn't work.
It causes failure in 
```sh
25: [ RUN      ] test_data_xml
25: [  ERROR   ] --- "2005-05-25T23:15:15.88888+00:00" != "2005-05-25T21:15:15.88888-02:00"
25:  1 FAILED TEST(S)
1/1 Test #25: utest_yang_types .................***Failed    0.04 sec
```

jktjkt:
> From my man page:
> ...

Oops, my bad.

irfanHaslanded:
Yes, in the attached link as well
```sh
Unlike localtime(), the reentrant version is not required to set tzname
If the reentrant version does not set tzname, it shall not set daylight and shall not set timezone.
```

michalvasko (approved):
+1 for the function name.

PR #2366, opened 2025-03-05
bugfix lyd_dup find first sibling explicitly

irfanHaslanded:
the first_dup may not always remain the first sibling. It must be explicitly found.


irfanHaslanded:
> Could you please mention details about the use-case when `first_dup` is not the first sibling? Ideally even include them as a comment in the code.

I will try and reproduce the original issue, for which this was needed.

irfanHaslanded:
It was likely due to passing a incorrectly constructed input tree to duplicate. I cannot reproduce it anymore.

PR #2378, opened 2025-04-01
context UPDATE add ietf-yang-library-augmentedby support

jeremie6wind:
Add the list of modules that augment a particular yang module. The modules and their augmentations are visible under the "/ietf-yang-library:yang-library xpath".

Link: https://datatracker.ietf.org/doc/html/draft-ietf-netconf-yang-library-augmentedby


michalvasko:
Seems fine but I would strongly prefer to wait for the YANG module RFC to be published before implementing it.

Issue #2374, opened 2025-03-31
Hash table expected behavior


steweg:
Hi,

I am currently troubleshooting one issue related to leafref_links, which uses your internal hash table implementation. I just wanted to ask about expectation of validity of returned records from the table. I am right that if some operation happens to hash_table, e.g. removal of records, the other records are untouched and pointers to them are still valid or whole table can be reallocated and as such pointers to other records are invalid and I need to check it look them again?

My current troubleshooting implicates that the pointers are invalidated during hash_table operations, but I am not sure if this is intended or it is kind of bug.

Thanks

michalvasko:
Yes, when the hash table needs to be resized, all the values are reallocated. If this is a problem for you, store pointers of the values.

PR #2379, opened 2025-04-01
Fixed invalid memory read/segfault

steweg:
This patch fixes segfaults caused by invalid memory read during leafref links removal process. It is tight hash_table implementation, as hash_table removal of record can invalidate all other record pointers due to memory reallocation process

Related to issue #2374

### DIFF
--- a/src/tree_data.c
+++ b/src/tree_data.c
@@ -3610,8 +3610,11 @@ LY_ERR
 lyd_get_or_create_leafref_links_record(const struct lyd_node_term *node, struct lyd_leafref_links_rec **record, ly_bool create)
 {
     struct ly_ht *ht;
+    LY_ERR ret = LY_SUCCESS;
     uint32_t hash;
     struct lyd_leafref_links_rec rec = {0};
+    struct lyd_leafref_links_rec *rec_p = &rec;
+    struct lyd_leafref_links_rec **rec_p2;
 
     assert(node);
     assert(record);
@@ -3626,15 +3629,23 @@ lyd_get_or_create_leafref_links_record(const struct lyd_node_term *node, struct 
     ht = LYD_CTX(node)->leafref_links_ht;
     hash = lyht_hash((const char *)&node, sizeof node);
 
-    if (lyht_find(ht, &rec, hash, (void **)record) == LY_ENOTFOUND) {
+    if (lyht_find(ht, &rec_p, hash, (void **)&rec_p2) == LY_ENOTFOUND) {
         if (create) {
-            LY_CHECK_RET(lyht_insert_no_check(ht, &rec, hash, (void **)record));
+            rec_p = calloc(1, sizeof rec);
+            rec_p->node = node;
+            LY_CHECK_ERR_RET(!rec_p, LOGMEM(LYD_CTX(node)), LY_EMEM);
+            ret = lyht_insert_no_check(ht, &rec_p, hash, (void **)&rec_p2);
+            LY_CHECK_ERR_GOTO(ret, free(rec_p), cleanup);
         } else {
             return LY_ENOTFOUND;
         }
     }
 
-    return LY_SUCCESS;
+cleanup:
+    if (!ret) {
+        *record = *rec_p2;
+    }
+    return ret;
 }
 
 LIBYANG_API_DEF LY_ERR
--- a/src/context.c
+++ b/src/context.c
@@ -272,9 +272,9 @@ ly_ctx_ht_err_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UN
 static ly_bool
 ly_ctx_ht_leafref_links_equal_cb(void *val1_p, void *val2_p, ly_bool UNUSED(mod), void *UNUSED(cb_data))
 {
-    struct lyd_leafref_links_rec *rec1 = val1_p, *rec2 = val2_p;
+    struct lyd_leafref_links_rec **rec1 = val1_p, **rec2 = val2_p;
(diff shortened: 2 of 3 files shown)


steweg:
I have not added any specific tests as I was not able to reproduce the issue using just C code. I my project I am using libyang-python, which actually triggers this and I was able to find the issue by using valgrind plugin. I am still not 100% that the proposed patch is correct solution as this behavior of hash_table is not explained in documentation nor expect by the user. I might even suggest to rather have like internal structure of pointers to the final records and if you need to reallocate, reallocate just that part but keep the returned record pointers valid for the caller.
Alternative approach is to avoid using operations with the hash_table in sequence like this:
- get 1st record associate with hash1
- while performing some operation with 1st record, find 2nd record associated with hash2 and remove 2nd record from table
- continue working with 1st record (<-- here you are already working with invalid memory)

Hope it helps to explain the problem. Feel free to propose better solution.

steweg:
I have rewritten the bugfix based on comment in the issue. I get checks error related to MSVC, but it seems not to be problem of my code, if I understand correctly the output. If there is some issue with my patch just let me know, I will fix it

src/tree_data.c
```diff
@@ -3626,15 +3629,23 @@ lyd_get_or_create_leafref_links_record(const struct lyd_node_term *node, struct
...
         } else {
             return LY_ENOTFOUND;
         }
     }
 
+    *record = *rec_p2;
     return LY_SUCCESS;
+
+cleanup:
+    return ret;
```
michalvasko, line 3648: `cleanup` label should always be performed, so please adjust this. Something like
```C
cleanup:
    if (!ret) {
        *record = *rec_p2;
    }
    return ret;
```

michalvasko (changes requested):
Code changes seem fine, just a minor issue.

But why do you actually need records not to be reallocated? Is that a *libyang* requirement or of your application and specific usage?

src/tree_data.c
steweg: Done

steweg:
> Code changes seem fine, just a minor issue.
> 
> But why do you actually need records not to be reallocated? Is that a _libyang_ requirement or of your application and specific usage?

Well as the leafref link records include eventually cyclic reference to each other. On one side leafref is having refrerence its target while target is able to see which leafrefs are pointing to it. The code always build both references and similarly it removes it from references in case of removal. And the removal is the problem as I need to make sure that record entry in memory is still valid regardless of operation done on target node. Unfortunately in case of remove it is not true as removal of reference can lead to having just an empty record with no targets or leafrefs. In that case I wanted to remove the record from good from memory as it is no-use.
So in the end I end-up exactly in scenario that I needed valid pointer to 1st record while I remove 2nd record from hash_table and continue to use 1st records for other operations. Hash table reallocation actually break the 1st record validity.

michalvasko:
> Well as the leafref link records include eventually cyclic reference to each other.

How? I see that `struct lyd_leafref_links_rec` includes only pointer and arrays of pointers to other nodes, the records themselves are not referenced.

steweg:
> > Well as the leafref link records include eventually cyclic reference to each other.
> 
> How? I see that `struct lyd_leafref_links_rec` includes only pointer and arrays of pointers to other nodes, the records themselves are not referenced.

Yes and no. The relationship originally should be as follows:
- node1 1<--->0..n target_nodes
- node1 1<--->0..n leafref_nodes

Because you asked me to not modify lyd_node_term structure, we are using hash_table instead which changes the relationship as follows:
- node1 1<--->0..1 record
- node1 1<--->0..n target_nodes
- record 1<--->0..n leafref_nodes

So the cyclic relationship is there like this:
- node1 --> record1 --> leafref_nodes[x] -> node2 --> record2 --> target_nodes[y] --> node1

Hope it helps to understand the problem

michalvasko:
So the link `node1` -> `record1` would be a problem, if it was a direct pointer. But all I found is the function `lyd_get_or_create_leafref_links_record()`, which finds the record of the node in the hash table. Meaning even if the record is moving in memory after hash table realloc, it will work because there is no direct link (pointer). So I still do not see the issue.

steweg:
> So the link `node1` -> `record1` would be a problem, if it was a direct pointer. But all I found is the function `lyd_get_or_create_leafref_links_record()`, which finds the record of the node in the hash table. Meaning even if the record is moving in memory after hash table realloc, it will work because there is no direct link (pointer). So I still do not see the issue.

Yes, it is not directly link/pointer. So let's say that behavior of hash_table is as it is and the scope of my issue just that function. So I have two solutions:
1. Rewrite the code as I did right now to use pointer value as hash_table records
2. Rewrite the code in a way that I will perform additional lookups in hash_table after the possible "reallocation" (this was my initial version of patch)

Either way it will work for my case. I would prefer to keep the current proposed solution (1). Is that ok with you? Or do you prefer different solution?

steweg:
I don't think it is good idea. The record itself includes two arrays of pointers to nodes, so if I return a copy, shall this be copied as well (a.k.a. deepcopy) or just copy array pointer (a.k.a. shallow copy). I guess you are assuming doing shallow-copy. But then when it comes to array deletion, the pointer in hash_table record can be left in invalid state after removal of final element from array. Deepcopy is even worse.

The other question would be who would release the memory of the "copy" record and when? I can imagine that it can be put on caller stack, so freeing memory will be done automatically once the record is not needed anymore.

That being said, I would rather not change it in this way. But rather use solution 1 or 2.

michalvasko (approved):

PR #2386, opened 2025-04-17
Coverity_fixes_jsc-se

jsc-se:
Fixes of issues from company internal Coverity check.

### DIFF
--- a/src/plugins_types/union.c
+++ b/src/plugins_types/union.c
@@ -350,6 +350,8 @@ union_find_type(const struct ly_ctx *ctx, struct lysc_type_union *type_u, struct
         }
         if (msg_len == -1) {
             LY_CHECK_ERR_GOTO(!errs, ret = LY_EMEM, cleanup);
+            /* for further actions in function msg_len is just 0 */
+            msg_len = 0;
         }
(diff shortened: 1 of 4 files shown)


src/schema_compile_node.c
```diff
@@ -4223,6 +4223,7 @@ lys_compile_grouping(struct lysc_ctx *ctx, struct lysp_node *pnode, struct lysp_
     len = lys_compile_grouping_pathlog(ctx, grp->parent, &path);
     if (len < 0) {
         LOGMEM(ctx->ctx);
+        free(path);
```
michalvasko, line 4226: I have tried to fix this properly with the last `devel` commit so it should not be needed anymore.

src/xpath.c
```diff
@@ -3666,7 +3666,9 @@ warn_equality_value(const struct lyxp_expr *exp, struct lyxp_set *set, uint32_t
                 type->plugin->free(set->ctx, &storage);
             }
         }
-        free(value);
+        if (value) {
+            free(value);
```
michalvasko, line 3670: You can safely call `free()` on `NULL` and we always do, the `if` is redundant.

michalvasko (changes requested):

michalvasko (approved):

PR #2391, opened 2025-05-02
ly_out add API for ly_out_printed_total

irfanHaslanded:
`ly_out_printed()` only gives the bytes printed in the last writer function call.
There are use cases where we need to know total number of bytes written as well.

So, expose this with `ly_out_printed_total()` API

### DIFF
--- a/src/out.h
+++ b/src/out.h
@@ -63,6 +63,7 @@ extern "C" {
  *
  * - ::ly_out_type()
  * - ::ly_out_printed()
+ * - ::ly_out_printed_total()
  *
  * - ::ly_out_reset()
  * - ::ly_out_free()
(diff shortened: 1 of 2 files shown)


src/out.h
```diff
@@ -291,6 +292,14 @@ LIBYANG_API_DECL LY_ERR ly_write(struct ly_out *out, const char *buf, size_t len
  */
 LIBYANG_API_DECL size_t ly_out_printed(const struct ly_out *out);
 
+/**
+ * @brief Get the number of printed bytes.
```
michalvasko, line 296: Please improve and clearly state the difference. Something like
```
Get the total number of printed bytes of this out structure.
```

michalvasko (changes requested):

PR #2394, opened 2025-05-11
tree schema FEATURE emit paths suitable for key search

snar:
Motivation: consider the structure with list hierarchy:
```
container earth {
  list country {
    key name;
    leaf name {....};
    list city {
      key name;
......
```
and you want to find cities in some country. Path built with LYSC_PATH_DATA_PATTERN (```/earth/country[name='%s']/city[name='%s']/name```) can't be used for this task because it contains predicate for city (so you must know it in advance), and path built with LYSC_PATH_DATA (```/earth/country/city/name```) will find cities in the entrie earth, not limited by country.
New LYSC_PATH_KEY_PATTERN will emit ```/earth/country[name='%s']/city/name``` here making the task doable.

### DIFF
--- a/src/tree_schema.c
+++ b/src/tree_schema.c
@@ -695,7 +695,7 @@ lysc_path_until(const struct lysc_node *node, const struct lysc_node *parent, LY
         buffer[0] = '\0';
     }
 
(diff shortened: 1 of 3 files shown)


michalvasko:
I am sorry but I do not see much value in such an addition. There are lots of other path patterns with possible use-cases and none of them supported. So I suppose if you really need such a path, you have to generate it yourself.

PR #2404, opened 2025-06-03
common BUGFIX define PATH_MAX if not available

pinotree:
`PATH_MAX` is optional in POSIX, and thus may not be provided (e.g. on GNU/Hurd). Provide a fallback definition for it, so the existing code keeps working for now; include `<limits.h>` to ensure that `PATH_MAX` is available if provided on the current OS/libc.

### DIFF
--- a/src/ly_common.h
+++ b/src/ly_common.h
@@ -16,6 +16,7 @@
 #ifndef LY_COMMON_H_
 #define LY_COMMON_H_
 
+#include <limits.h>
 #include <pthread.h>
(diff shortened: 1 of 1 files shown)


michalvasko (approved):
No problem, thanks.

michalvasko (changes requested):
`uncrustify` requires includes to be sorted alphabetically, that is why the tests failed.

michalvasko (approved):

PR #2410, opened 2025-07-08
yanglint UPDATE extension parsing and validation

Ardival:
Extension parsing and validation for both interactive and non-interactive mode. 
+example to try


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "files.associations": {
+        "typeinfo": "c"
+    }
+}
```
michalvasko, line 5: Redundant file.

tools/lint/cmd_data.c
```diff
@@ -417,6 +453,34 @@ evaluate_xpath(const struct lyd_node *tree, const char *xpath)
     return 0;
 }
 
+int
+parse_ext_string(char *extension_name, struct yl_opt *yo)
```
michalvasko, line 457: In this case modifying `extension_name` (not a good param name) should be fine but in general is not a good practice so please change to `const char *` and parse the fields without the use of `strtok()`.

tools/lint/cmd_data.c
```diff
@@ -675,11 +739,53 @@ process_data(struct ly_ctx *ctx, enum lyd_type type, uint8_t merge, LYD_FORMAT o
...
-    /* parse, validate and print data */
-    if (process_data(ctx, yo->data_type, yo->data_merge, yo->data_out_format, yo->out, yo->data_parse_options,
-            yo->data_validate_options, yo->data_print_options, &yo->data_operational, &yo->reply_rpc,
-            &yo->data_inputs, &yo->data_xpath)) {
-        return 1;
+    if (yo->data_ext) {
+        struct lyd_node *tree = NULL;
+        struct cmdline_file *input_f = NULL;
+        struct lys_module *module = NULL;
+        uint32_t idx = 0;
```
michalvasko, line 746: Initialize only variables that need to be.

tools/lint/cmd_data.c
```diff
@@ -675,11 +739,53 @@ process_data(struct ly_ctx *ctx, enum lyd_type type, uint8_t merge, LYD_FORMAT o
...
+
+        while ((module = ly_ctx_get_module_iter(ctx, &idx))) {
+            if (!strcmp(module->prefix, yo->prefix)) {
```
michalvasko, line 749: It is not safe to assume the prefix is unique, only the module name must be unique so use that instead.

tools/lint/cmd_data.c
```diff
@@ -675,11 +739,53 @@ process_data(struct ly_ctx *ctx, enum lyd_type type, uint8_t merge, LYD_FORMAT o
...
+            }
+        }
+
+        if (!module) {
+            YLMSG_E("Cannot find the \"%s\" prefix in yang modules.", yo->prefix);
+            return 1;
+        }
+
+        /* get the extension from module that user is looking for */
+        for (size_t i = 0; i < LY_ARRAY_COUNT(module->compiled->exts); ++i) {
```
michalvasko, line 760: Can be replaced with `LY_ARRAY_FOR(module->compiled->exts, i) {`.

tools/lint/cmd_data.c
```diff
@@ -675,11 +739,53 @@ process_data(struct ly_ctx *ctx, enum lyd_type type, uint8_t merge, LYD_FORMAT o
...
+
+        if (!yo->ext) {
+            YLMSG_E("Extension '%s:%s:%s' not found in module '%s'.", yo->prefix, yo->name, yo->argument, module->name);
+            return 1;
+        }
+
+        input_f = (struct cmdline_file *)yo->data_inputs.objs[0];
+
+        if (lyd_parse_ext_data(yo->ext, NULL, input_f->in, input_f->format, 0, 0, &tree)) {
+            YLMSG_E("Parsing of extension data failed.")
```
michalvasko, line 776: Missing error return.

tools/lint/cmd_data.c
```diff
@@ -675,11 +739,53 @@ process_data(struct ly_ctx *ctx, enum lyd_type type, uint8_t merge, LYD_FORMAT o
 int
 cmd_data_process(struct ly_ctx *ctx, struct yl_opt *yo)
 {
-    /* parse, validate and print data */
-    if (process_data(ctx, yo->data_type, yo->data_merge, yo->data_out_format, yo->out, yo->data_parse_options,
-            yo->data_validate_options, yo->data_print_options, &yo->data_operational, &yo->reply_rpc,
-            &yo->data_inputs, &yo->data_xpath)) {
-        return 1;
+    if (yo->data_ext) {
+        struct lyd_node *tree = NULL;
```
michalvasko, line 743: Would probably make sense to have a function for getting the extension.

tools/lint/main_ni.c
```diff
@@ -215,6 +217,10 @@ help(int shortout)
     printf("  -J, --json-null\n"
             "                Allow usage of JSON empty values ('null') within input data\n\n");
 
+    printf("  -k, --ext-id\n"
```
michalvasko, line 220: `ext-id` is okay but could be called better, `id` is not used much in YANG. Just `ext-inst` or `ext-instance` should be better, you are identifying a specific extension instance.

tools/lint/cmd.c
```diff
@@ -70,7 +70,7 @@ COMMAND commands[] = {
     },
     {
         "data", cmd_data_opt, cmd_data_dep, cmd_data_store, cmd_data_process, cmd_data_help, NULL,
-        "Load, validate and optionally print instance data", "d:ef:F:hmo:O:R:r:nt:x:"
+        "Load, validate and optionally print instance data", "d:ef:F:hmo:O:R:r:nt:x:ik:"
```
michalvasko, line 73: `i` parameter seems redundant and a data type `ext` should be used just like for non-interactive params.

michalvasko (changes requested):

PR #2413, opened 2025-07-14
yanglint UPDATE extension validation

Ardival:


tools/lint/cmd_data.c
```diff
@@ -255,6 +263,12 @@ cmd_data_opt(struct yl_opt *yo, const char *cmdline, char ***posv, int *posc)
                 return 1;
             }
             break;
+        case 'k': /* --ext-id */
```
michalvasko, line 266: Updated to `--ext-inst`.

PR #2424, opened 2025-09-13
Fixed missing leafref linking

steweg:
This patch fixes the problem of missing leafref linking in case of using leafref within union types


michalvasko (changes requested):
No problem, just minor issues.

PR #2431, opened 2025-09-28
xpath UPDATE to support usage of union types as deref targets

steweg:
This patch introduces ability to use union types, which internally uses leafrefs to be used as deref target nodes. This fullfils the last statement of RFC 7950 section 10.3.1. This behavior is also allowed within leafref paths which uses deref function in it.

This patch also fixes segfault in case of using cyclic deref paths

### DIFF
--- a/src/path.c
+++ b/src/path.c
@@ -1025,6 +1025,95 @@ cleanup:
  *
  * @param[in] ctx libyang context.
  * @param[in] cur_node Current (original context) node.
+ * @param[in] target_node The deref target schema node.
+ * @param[in] cur_type The currently evaluated type of deref target node.
+ * @param[in] top_ext Extension instance containing the definition of the data being created. It is used to find
+ * the top-level node inside the extension instance instead of a module. Note that this is the case not only if
+ * the @p ctx_node is NULL, but also if the relative path starting in @p ctx_node reaches the document root
+ * via double dots.
+ * @param[in] expr Parsed path.
+ * @param[in] oper Oper option (@ref path_oper_options).
+ * @param[in] target Target option (@ref path_target_options).
+ * @param[in] format Format of the path.
+ * @param[in] prefix_data Format-specific data for resolving any prefixes (see ::ly_resolve_prefix).
+ * @param[in] log Whether to generate log message or not
+ * @param[in,out] tok_idx Index in @p exp, is adjusted.
+ * @param[out] path Compiled path.
+ * @return LY_ERR value.
+ */
+static LY_ERR
+ly_path_compile_deref_type(const struct ly_ctx *ctx, const struct lysc_node *cur_node, const struct lysc_node *target_node,
+        const struct lysc_type *cur_type, const struct lysc_ext_instance *top_ext, const struct lyxp_expr *expr, uint16_t oper, uint16_t target,
+        LY_VALUE_FORMAT format, void *prefix_data, ly_bool log,  uint32_t *tok_idx, struct ly_path **path)
+{
+    LY_ERR ret = LY_SUCCESS;
+    struct lyxp_expr expr2;
+    struct ly_path *path2 = NULL;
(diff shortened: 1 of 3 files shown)


steweg:
Just some comments on current behavior of libyang. Current version doesn't allows to deref any type of node, just the valid ones, which contradicts the RFC 7950, which says that it should support it but return empty set.

I understand that RFC7950 doesn't explicitly allows usage of unions as deref target nodes, but it make sense in case when you have scenario when yang model allows usage of predefined set of data and refer to it using leafref, and also allows to use completely custom set, using the same kind data model as for predefined ones... on top of that there might be some data which is never predefined and therefore it can have two possible paths to reference some data. Either through predefined set or through custom. Here is a yang model to better explain this:
```
module b {
    yang-version 1.1;
    namespace "asd:asd:asd:b";
    prefix b;

    grouping common-struct {
        leaf k {type string;}
        leaf v1 {type string;}
        // ...
        leaf vX {type string;}
    }

    list predefined_data {key k; uses common-struct;}

    rpc create_action {
        input {
            leaf predefined {
                type union {
                    type leafref {path "/predefined_data/k";}
                    type string {pattern "custom";}
                }
                mandatory true;
            }
            container custom_data {
                uses common-struct;
                when "../predefined = 'custom'";
            }
            leaf additional_selected_data {
                type union {
                    type leafref {path "../custom_data/vX";}
                    type leafref {path "deref(../predefined)/../vX";}
                }
                mandatory true;
            }
        }
    }
}
```

I know that I have introduced union of leafref and string together, which could be just "optional" leafref. In my case this is not possible as empty value would automatically be consider by UI layer in my project that user wishes to create custom data, which quite a lot of data. This creates user experience to see what all details are needed if he/she wants to defined it manually, and should rather use predefined set. Once he/she changed that UI will hide not needed staff leaving just the additional selected data field to be filled in by user. So we have to introduce something like string literal value, which is not supported by YANG natively, therefore this trick with `string with pattern` and need to use `deref` with union

src/path.c
```diff
@@ -1083,53 +1171,26 @@ ly_path_compile_deref(const struct ly_ctx *ctx, const struct lysc_node *cur_node
...
-    /* compile rest of the path and append it to the path */
-    LY_CHECK_GOTO(ret = ly_path_compile_leafref(ctx, node2, top_ext, &expr2, oper, target, format, prefix_data, &path2),
-            cleanup);
     LY_CHECK_GOTO(ret = ly_path_append(ctx, path2, path), cleanup);
+    deref_leaf_node = (const struct lysc_node_leaf *)node2;
+    ret = ly_path_compile_deref_type(ctx, cur_node, node2, deref_leaf_node->type, top_ext, expr, oper, target, format, prefix_data, 1, tok_idx, path);
 
 cleanup:
     ly_path_free(path2);
+    path2 = NULL;
```
michalvasko, line 1193: Seems redundant.

src/xpath.c
```diff
@@ -3980,6 +3980,83 @@ xpath_current(struct lyxp_set **args, uint32_t arg_count, struct lyxp_set *set,
...
+/**
+ * @brief Executes deref function on specific type and value. It performs evaluation in recursive manner,
+ *        which supports usage of union types as deref targets. Returns LYXP_SET_NODE_SET with either
+ *        leafref or instance-identifier target node(s).
+ *
+ * @param[in] leaf The target deref data node
+ * @param[in] sleaf The target deref schema node
+ * @param[in] value The currently evaluated value of target deref node depending on current type
+ * @param[in] cur_type The currently evaluated type of target deref node
+ * @param[in] top_type Whether the cur_type is the top level deref target node type or not
```
michalvasko, line 3992: Would call this `log` instead because that is the effect of this flag, an error is logged.

src/path.c
steweg: Done

PR #2432, opened 2025-09-28
parsing BUGFIX invalid backward parent pointer

steweg:
This patch fixes the issue if extensions are used within subsequent refine or typedef statements


michalvasko:
Okay, then I misunderstood the issue and it is simpler than I thought. Please document the actual problem (`realloc` is causing the pointers to become invalid) and it can be merged, thanks.

michalvasko:
Yes, I meant in the code, is fine now.

PR #2436, opened 2025-10-04
validation BUGFIX segfault when using containter with when condition

steweg:
This patch fixes the segfault, which is happening when RPC output data, which uses NP-container with when condition are being validated.


steweg:
I have to say that I am not 100% sure if my "fix" is correct.

The root cause is that when validation of RPC output is being performed, it add implicit nodes for some reason (don't know why), which can have `when` condition set. That automatically adds the node into `nodes_when` for evaluation of when later on. But subsequently further validation steps adds the node again into `nodes_when`. Then later during evaluation of unresolved `nodes_when`, the code evaluate first one as false and remove the node for good. And during second evaluation hits segfault conditions.

There can be multiple ways how to fix it, but from my perspective I don't understand why the `ly_set_add` has ability to behave like `list` and is actually used like that in case of `nodes_when`. Maybe it was done due to performance optimizations or something like that, but apparently has side effects. So solution is to either change usage of `nodes_when` back to `set` behavior or adjust the logic which validates RPC output somehow so that it will not add nodes twice into `nodes_when`

I will leave it up to you to judge

PR #2441, opened 2025-10-16
parsing BUGFIX invalid backward extension parent pointer

steweg:
This patch fixes the issue if extensions are used within:
 - subsequent `enum` statements
 - subsequent `type` statements of `union` type
 - subsequent `type` statements of `typedefs`


michalvasko:
Thanks, perhaps now it works correctly for all the statements.

PR #2446, opened 2025-10-27
WIP: Extension validation for the YANG module versioning work

xorrkaz:
This is a WIP PR.  An early review would be appreciated, but I want to see if others can bang on the code a bit, too.


xorrkaz:
I _think_ I licked the CI problem.  Can you re-run the PR?  And thanks for the quick look.

michalvasko:
Yes, it works now, only some formatting issue left. I can fix that, if you want.

xorrkaz:
Should be all fixed.  Give it another shot.  I do want to give this some time to bake, though.  I want more IETFers to look at it, and the YANG modules may rev prior to an RFC number being issued.

src/plugins_exts/yang_revisions.c
```diff
@@ -0,0 +1,185 @@
...
+
+    if ((date[4] != '-') || (date[7] != '-')) {
+        return LY_EVALID;
+    }
+
+    if (sscanf(date, "%4d-%2d-%2d", &year, &month, &day) != 3) {
+        return LY_EVALID;
+    }
+
+    if ((year < 1000) || (year > 9999) || (month < 1) || (month > 12) || (day < 1) || (day > 31)) {
```
avtobiff, line 87: Should dates be verified to actually exist?

This rule would allow e.g. 2026-02-31.

RFC 7950 doesn't seem to give much guidance but points at
RFC 6991 which in turn points at RFC 3339 and ISO 8601.

xorrkaz: Good question.  I considered this at the time with the code, but I don't know that other date validation is done, so I left it simpler.  That said, a bad date has bitten YANG Catalog in the past.  Let me think it over.

michalvasko: Some internal [revision date validation](https://github.com/CESNET/libyang/blob/master/src/tree_schema_common.c#L64) is performed, naturally. While it is not a public function, you can use it in an internal plugin.

michalvasko:
Fully implemented in the `schema_comparison` branch.

PR #2448, opened 2025-10-30
doc: `lyd_parse_op` has special requirements when parsing notifications

jktjkt:
As confirmed by Michal, the parsing code will try to read the `notifications` module, if it is available, or if the parsed data from the `ietf-yang-types` module are not available. As a result, a context which has not loaded the `notifications` module *and* which was not created with `LY_CTX_SET_PRIV_PARSE` will return an error when parsing notifications.

### DIFF
--- a/src/parser_data.h
+++ b/src/parser_data.h
@@ -389,6 +389,8 @@ enum lyd_type {
  *   - @p op - must be provided, the RPC/action data tree itself will be returned here, pointing to the operation;
  *
  * - ::LYD_TYPE_NOTIF_NETCONF:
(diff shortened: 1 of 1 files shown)


michalvasko:
That is not accurate. Normal context will work fine because it keeps the parsed modules and nothing else is needed. It is specific to *sysrepo* that it frees the parsed modules to save some memory except when the `LY_CTX_SET_PRIV_PARSED` flag is used when the modules are not freed (but is mostly a side-effect of the flag).

michalvasko:
No longer relevant, the `dat-and-time` value can always be validated.

PR #2453, opened 2025-11-25
Allow build on Solaris 11.4

psumbera:


CMakeLists.txt
```diff
@@ -76,7 +76,7 @@ if(WIN32)
     set(CMAKE_C_FLAGS "/Zc:preprocessor /W3 /wd4711 /w14013 /utf-8 ${CMAKE_C_FLAGS}")
 else()
     # global C flags
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wpedantic -std=c11")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wpedantic")
```
michalvasko, line 79: Sorry, but the C standard needs to be specified because various compilers use different default standards. And before deciding the exact solution I would like to know the details of the issue. Is the compiler too old on Solaris that does not support this C revision?

CMakeModules/FindPCRE2.cmake
```diff
@@ -15,6 +15,7 @@ else()
         pcre2.h
         PATHS
         /usr/include
+        /usr/include/pcre
```
michalvasko, line 18: Not really a problem since this is a transparent change but seems strange. Is the separate directory specific only for `pcre` or all the libraries have their own directories?

src/tree_schema.c
```diff
@@ -2370,7 +2370,11 @@ lys_search_localfile_file_type(const struct dirent *file, const char *wd, struct
 
     *skip = 0;
 
+#ifdef __sun__
+    if (1) {
+#else
     if ((file->d_type == DT_UNKNOWN) || (file->d_type == DT_LNK)) {
+#endif
```
michalvasko, line 2377: Please explain because just looking at this I do not understand.

michalvasko (changes requested):
The changes need to be discussed more but also change the base branch to `devel`, we do not accept PRs into `master`. But we plan to make a new release soon when `devel` will be merged into `master` so it should not matter to you.

CMakeLists.txt
psumbera: Without this, declarations for `vasprintf` and `strndup` are missing:
```
/scratch/psumbera/libyang/src/ly_common.c:704:11: error: implicit declaration of function 'vasprintf'; did you mean 'vsprintf'? [-Wimplicit-function-declaration]
  704 |     len = vasprintf(&addition, format, fp);
      |           ^~~~~~~~~
      |           vsprintf
/scratch/psumbera/libyang/src/ly_common.c: In function 'ly_parse_int':
/scratch/psumbera/libyang/src/ly_common.c:733:11: error: implicit declaration of function 'strndup'; did you mean 'strncmp'? [-Wimplicit-function-declaration]
  733 |     str = strndup(val_str, val_len);
      |           ^~~~~~~
      |           strncmp
/scratch/psumbera/libyang/src/ly_common.c:733:9: error: assignment to 'char *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
  733 |     str = strndup(val_str, val_len);
      |         ^
/scratch/psumbera/libyang/src/ly_common.c: In function 'ly_parse_uint':
/scratch/psumbera/libyang/src/ly_common.c:773:9: error: assignment to 'char *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
  773 |     str = strndup(val_str, val_len);
      |         ^
```

Note that you can play with Solaris via https://github.com/vmactions/shell-solaris .

CMakeModules/FindPCRE2.cmake
psumbera: I think it's rather specific for `pcre`: https://github.com/oracle/solaris-userland/blob/c23e526bf3805e20d5b72018168c400b491716dc/components/pcre2/Makefile#L80

src/tree_schema.c
psumbera: `d_type` is not part of `struct dirent` on Solaris.

CMakeModules/FindPCRE2.cmake
michalvasko: Weird, but fine.

src/tree_schema.c
michalvasko: I see, then it should ideally be properly detected in by `cmake` and then handled here. I can add that and it may be faster, if you want.

CMakeLists.txt
michalvasko: Seems like there are some macro definitions missing, you can learn about those in the man pages. The VM seems nice but to have a lasting effect we would need to add it into our automatic actions, which does not seem possible.

psumbera: As for macros. This is also possibiltiy:
```
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,9 @@ if(WIN32)
 else()
     # global C flags
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Wpedantic -std=c11")
+    if(${CMAKE_SYSTEM_NAME} STREQUAL "SunOS")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D__EXTENSIONS__")
+    endif()
 endif()
 
 include_directories(${PROJECT_BINARY_DIR}/libyang ${PROJECT_SOURCE_DIR}/src ${PROJECT_SOURCE_DIR}/src/plugins_exts)
```

As for Solaris CI. It's possible: https://github.com/rust-lang/libc/blob/d58c1f4f9ae01648b342d7dead291d08655677c4/.github/workflows/ci.yaml#L278

michalvasko: Fine although the macro looks very non-standard. Better adding macros than changing the C standard. As for the CI, I may take a look at it later and add an action for Solaris in our CI.

michalvasko:
Should now be supported and running in the CI.

PR #2454, opened 2025-12-04
New gNMI

Humblesaw:


src/parser_data.h
```diff
@@ -336,6 +336,24 @@ LIBYANG_API_DECL LY_ERR lyd_parse_data_path(const struct ly_ctx *ctx, const char
...
         LYD_FORMAT format, uint32_t parse_options, uint32_t validate_options, struct lyd_node **tree);
 
+/**
+ * @brief Parse data from the input handler as a bare JSON value and connect it to the node parsed from the path.
+ *
+ * @param[in] ctx Context to connect with the tree being built here.
+ * @param[in] path Path to the node so the value in the input handler can be connected to it.
+ * @param[in] in Input handler with the value in the JSON format.
+ * @param[in] format Currently only LYD_JSON is supported.
+ * @param[in] new_val_options Options for new values, see @ref newvaloptions.
```
michalvasko: Describe what they are used for.

src/parser_json.c
```diff
@@ -1879,17 +1897,88 @@ lyd_parse_json(const struct ly_ctx *ctx, const struct lysc_ext_instance *ext, st
     /* find the operation node if it exists already */
     LY_CHECK_GOTO(rc = lyd_parser_find_operation(parent, int_opts, &lydctx->op_node), cleanup);
 
-    /* read subtree(s) */
-    do {
-        r = lydjson_subtree_r(lydctx, parent, first_p, parsed);
-        LY_DPARSER_ERR_GOTO(r, rc = r, lydctx, cleanup);
+    if (status != LYJSON_OBJECT) {
+        /* parse bare JSON value */
```
michalvasko, line 1996: Add a function for all the code in this `if` branch.

src/tree_data.c
```diff
@@ -279,6 +279,75 @@ lyd_parse_data_path(const struct ly_ctx *ctx, const char *path, LYD_FORMAT forma
...
+    struct lyd_node *new_parent = NULL, *top_level = NULL;
+    const struct lysc_node *new_node_schema = NULL;
+    struct ly_set parsed = {0};
+    ly_bool p_decremented = 0;
+
+    assert(ctx);
+    assert(path && (path[0] == '/'));
+
+    /* other formats are not supported for now */
+    assert(format == LYD_JSON);
```
michalvasko: `assert` cannot be used for checking API function arguments.

michalvasko (changes requested):

michalvasko (approved):

tests/utests/data/test_parser_json.c
```diff
@@ -69,6 +71,40 @@ setup(void **state)
 #define CHECK_LYD_STRING(IN_MODEL, PRINT_OPTION, TEXT) \
     CHECK_LYD_STRING_PARAM(IN_MODEL, TEXT, LYD_JSON, PRINT_OPTION)
 
+static void
+test_baretop_leaf(void **state)
```
jktjkt, line 75: This test is checking whether you can parse a "content of a JSON field" when its value is given out-of-band (and properly NS-qualified).

Have you added a test which checks that you can still parse a JSON like `{"al":25}`?

jktjkt (changes requested):
We talked about this on Slack -- this change apparently causes libyang to skip printing of namespace prefixes for non-top-level nodes, *even if they appear as a root node in the resulting JSON*. That's a big change which should be properly explained. A commit that says "printer json FEATURE no namespace for inner nodes" is not a proper explanation because it doesn't say **why** this change needs to happen -- it's simply saying that you're making this change. So, why is it needed, and why is it a correct change?

Now, RFC 7950 is silent about this topic, but given that *all* the JSON examples from the [RESTCONF RFC](https://datatracker.ietf.org/doc/html/rfc8040) always use namespace-qualified names even when printing subtrees, I think that keeping them namespace-qualified is a good thing. Also, this is about producing this JSON; have you checked that libyang can parse these JSONs, and is there a unit test for that?

Also, the PR currently contains two commits (which is a good thing), but the first one introduces a build regression because of its use of `LYD_PRINT_WITHSIBLINGS`. This is then fixed by the follow-up commit, which means that git bisection will not work. That's a bug.

tests/utests/data/test_parser_json.c
Humblesaw: Yes, I added it and I also added a test for printing JSON.

src/tree_data.c
```diff
@@ -279,6 +279,80 @@ lyd_parse_data_path(const struct ly_ctx *ctx, const char *path, LYD_FORMAT forma
...
+    struct lyd_node *new_parent = NULL, *top_level = NULL;
+    const struct lysc_node *new_node_schema = NULL;
+    ly_bool p_decremented = 0;
+
+    if (!ctx) {
+        return LY_EINVAL;
+    }
+    if (!path || (path[0] != '/')) {
+        return LY_EINVAL;
+    }
```
michalvasko: Use `LOGARG()` instead.

src/tree_data.c
```diff
@@ -279,6 +279,80 @@ lyd_parse_data_path(const struct ly_ctx *ctx, const char *path, LYD_FORMAT forma
...
+
+    /* other formats are not supported for now */
+    if (format != LYD_JSON) {
+        return LY_EINVAL;
```
michalvasko, line 303: Add an error message, similar to line 318.

michalvasko (changes requested):

Humblesaw:
> We talked about this on Slack -- this change apparently causes libyang to skip printing of namespace prefixes for non-top-level nodes, _even if they appear as a root node in the resulting JSON_. That's a big change which should be properly explained. A commit that says "printer json FEATURE no namespace for inner nodes" is not a proper explanation because it doesn't say **why** this change needs to happen -- it's simply saying that you're making this change. So, why is it needed, and why is it a correct change?
> 
> Now, RFC 7950 is silent about this topic, but given that _all_ the JSON examples from the [RESTCONF RFC](https://datatracker.ietf.org/doc/html/rfc8040) always use namespace-qualified names even when printing subtrees, I think that keeping them namespace-qualified is a good thing. Also, this is about producing this JSON; have you checked that libyang can parse these JSONs, and is there a unit test for that?
> 
> Also, the PR currently contains two commits (which is a good thing), but the first one introduces a build regression because of its use of `LYD_PRINT_WITHSIBLINGS`. This is then fixed by the follow-up commit, which means that git bisection will not work. That's a bug.

Ok, this is only needed for the gNMI server that I'm currently working on. I added a flag that modifies the behavior of the JSON printer so that the current functionality can remain. I also added a test for it and correctly separated the commits (thanks for the catch).

michalvasko (approved):

tests/utests/data/test_parser_json.c
jktjkt: I think that this should be said explicitly. E.g., the docs for `LYD_PRINT_JSON_NO_NESTED_PREFIX` should say that parsing such a content will likely require some out-of-band information, and that `lyd_parse_value_fragment` is suitable for this purpose.

michalvasko (approved):

tests/utests/data/test_parser_json.c
Humblesaw: Ok, I added more info on that.

PR #2463, opened 2026-01-06
Add Solaris, FreeBSD, 32-bit Linux & big endian system workflows

Roytak:


Roytak:
Should also resolve the build problems listed in https://github.com/CESNET/libyang/pull/2453

michalvasko (approved):

PR #2482, opened 2026-02-11
schema diff UPDATE support for yang-version diff

avtobiff:


src/schema_diff_tree.c
```diff
@@ -3024,6 +3076,12 @@ schema_diff_module(const struct lys_diff_s *diff, const struct lys_module *mod1,
     LY_ERR rc = LY_SUCCESS;
     uint32_t i;
 
+    /* yang-version */
+    for (i = 0; i < diff->module_changes.count; ++i) {
+        LY_CHECK_GOTO(rc = schema_diff_module_version(&diff->module_changes.changes[i], mod1, mod2, diff_list),
+                cleanup);
+    }
+
```
michalvasko: I would prefer you remove `schema_diff_module_version`, rename `schema_diff_module_text` to `schema_diff_module_substmt`, for example, and add the functionality there. This 2 same loops are a bit confusing.

src/schema_diff.h
```diff
@@ -440,6 +441,20 @@ LY_ERR schema_diff_ext_insts_change(const struct lysc_ext_instance *exts1, const
...
+ * @brief Check changes of a 'yang-version'.
+ *
+ * @param[in] yvsn1 First yang-version.
+ * @param[in] yvsn2 Second yang-version.
+ * @param[in] parent_changed Parent statement of the change.
+ * @param[in] changed Changed statement.
+ * @param[in,out] changes Changes to add to.
+ * @return LY_ERR value.
+ */
+LY_ERR schema_diff_yangversion_change(uint8_t yvsn1, uint8_t yvsn2,
```
michalvasko: You can just put this function in `schema_diff.c` (cannot be clearly decided whether it concerns compiled or parsed schema anyway) and it can then be `static`.

michalvasko (changes requested):
Thanks, looks good, just some minor refactoring needed.

modules/ietf-yang-schema-comparison@2025-10-20.yang
```diff
@@ -914,6 +918,11 @@ module ietf-yang-schema-comparison {
   grouping module-substmts {
     description
       "All non-data-definition substatements of a module.";
+    leaf yang-version {
+      type string;
```
michalvasko, line 922: Do we perhaps want to make the `type` more constrained? Considering a new YANG version would almost certainly require a new revision of this module anyway, why not make it an `enumeration` with enums `1.0` and `1.1`, what do you think?

michalvasko (commented):
Just one more discussion point.

modules/ietf-yang-schema-comparison@2025-10-20.yang
avtobiff: Maybe, but the data type for the 'yang-version' keyword
argument is string. That is why I used that type.

How is it done in libyang in general?

michalvasko: I see. I have no strong preference, just seemed like a good idea to define exact expectations.

What do you mean done in libyang? You have used the macros that we have defined for this purpose.

avtobiff: I mean how is it parsed and constrained. Tested and
see that only 1 and 1.1 are allowed values.

IMHO it's enough to follow the datatype defined for
the 'yang-version' keyword.

michalvasko (approved):

PR #2483, opened 2026-02-11
Update date for ietf-yang-schema-comparison.yang

avtobiff:


michalvasko (approved):
Thanks, would have updated it myself.

PR #2491, opened 2026-02-24
Printer tree

lePici:
TODO: squash commits to one - done


src/printer_tree.c
```diff
@@ -3847,220 +3838,168 @@ pt_print_groupings(struct pt_tree_ctx *tc)
...
+    void *node = NULL, *child;
 
     substmts = compiled ? (void *)((struct lysc_ext_instance *)ext)->substmts :
             (void *)((struct lysp_ext_instance *)ext)->substmts;
 
     LY_ARRAY_FOR(substmts, i) {
-        stmt = compiled ? ((struct lysc_ext_instance *)ext)->substmts[i].stmt :
-                ((struct lysp_ext_instance *)ext)->substmts[i].stmt;
+        if (compiled) {
+            stmt = ((struct lysc_ext_instance *)ext)->substmts[i].stmt;
```
michalvasko: You can simplify this function by using a function similarly to https://github.com/CESNET/libyang/blob/devel/src/tree_schema.c#L445.

src/printer_tree.c
```diff
@@ -3840,64 +3840,24 @@ pt_print_groupings(struct pt_tree_ctx *tc)
...
  * @brief Read extension and get schema pointer.
  * @param[in] ext Extension to read.
- * @param[in] compiled Set to 1 if @p ext is from lysc tree.
- * @return pointer to lysp or lysc node.
+ * @param[in] stmt_mask First stmt which match will be returned.
+ * @return pointer to lysp node.
  */
 static void *
-pt_ext_read_storage(void *ext, ly_bool compiled)
+pt_ext_parsed_read_storage(struct lysp_ext_instance *ext, int stmt_mask)
```
michalvasko, line 3847: This function should no longer be required, or not? Just use the functions for getting the particular storage.

lePici: Which function do you mean?
In this case I only have `lysp_ext_instance`. Functions `lyplg_ext_get_storage` and `lyplg_ext_parsed_get_storage` has parameter `lysc_ext_instance` which may not always exist, because `tree_printer` can print from a parsed tree.

michalvasko: Right, sorry.

michalvasko (approved):

PR #2495, opened 2026-03-12
context UPDATE add option to force obsolete nodes compilation

jeremie6wind:
Since version 4 of libyang, obsolete nodes are not compiled by default. To compile them, the LY_CTX_COMPILE_OBSOLETE option must be passed to ly_ctx_new.
To avoid having to add this option to all existing ly_ctx_new calls, add the FORCE_LY_CTX_COMPILE_OBSOLETE build option to compile the obsolete nodes by default.


michalvasko:
> To avoid having to add this option to all existing ly_ctx_new calls

Sorry, but adding a compile-time var to the project instead does not seem like a good idea to me.

jeremie6wind:
Hi Michal,
Thank you for your reply.
Do you have any suggestion on how to globally compile the obsolete nodes?
Best regards
Jeremie

michalvasko:
No specific one, add the option to all `ly_ctx_new()` calls or create your own wrapper that will set it, neither seems like too much work.

Issue #2507, opened 2026-04-07
Unicode blocks in pattern not recognized


manoe:
libyang supports patterns that further restrict a string leaf's set of values. The YANG RFC [1] allows the use of Unicode blocks [2] in patterns (e.g. Basic Latin, Cyrillic). libyang, due to a coding error, interprets every Unicode block pattern as Latin-1 Supplement, if multiple Unicode blocks are present in the same pattern. In the corresponding code, the variable responsible to store the Unicode block information, gets overwritten.

[1] https://datatracker.ietf.org/doc/html/rfc6020
[2] https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#dt-ccesblock

manoe:
Corresponding correction PR available at:
https://github.com/CESNET/libyang/pull/2508

michalvasko:
Fixed by your PR.

PR #2508, opened 2026-04-07
Unicode block bug

manoe:
libyang supports patterns that further restrict a string leaf's set of values. The YANG RFC [1] allows the use of Unicode blocks [2] in patterns (e.g. Basic Latin, Cyrillic). libyang, due to a coding error, interprets every Unicode block pattern as Latin-1 Supplement, if multiple Unicode blocks are present in the same pattern. In the corresponding code, the variable responsible to store the Unicode block information, gets overwritten.

The PR corrects the behavior and provides also unit tests.

Corresponding issue: https://github.com/CESNET/libyang/issues/2507

[1] https://datatracker.ietf.org/doc/html/rfc6020
[2] https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#dt-ccesblock

### DIFF
--- a/src/ly_common.c
+++ b/src/ly_common.c
@@ -4,7 +4,6 @@
  * @brief common internal definitions for libyang
  *
  * Copyright (c) 2018 - 2026 CESNET, z.s.p.o.
- * Copyright (c) 2026 Nokia
  *
  * This source code is licensed under BSD 3-Clause License (the "License").
  * You may not use this file except in compliance with the License.
--- a/src/ly_common.c
+++ b/src/ly_common.c
@@ -4,6 +4,7 @@
  * @brief common internal definitions for libyang
  *
  * Copyright (c) 2018 - 2026 CESNET, z.s.p.o.
+ * Copyright (c) 2026 Nokia
  *
  * This source code is licensed under BSD 3-Clause License (the "License").
  * You may not use this file except in compliance with the License.
@@ -813,7 +814,7 @@ ly_pat_compile_xmlschema_chblocks_xmlschema2perl(const char *pattern, char **reg
         {NULL, NULL}
     };
 
-    size_t idx, idx2, start, end;
+    size_t idx, idx2, start, end, ublock;
     char *perl_regex, *ptr;
 
     perl_regex = *regex;
@@ -849,6 +850,7 @@ ly_pat_compile_xmlschema_chblocks_xmlschema2perl(const char *pattern, char **reg
             return ly_err_new(err, LY_EVALID, 0, NULL, NULL, "Regular expression \"%s\" is not valid (\"%s\": %s).",
                     pattern, perl_regex + start + 5, "unknown block name");
         }
+        ublock = idx;
 
         /* make the space in the string and replace the block (but we cannot include brackets if it was already enclosed in them) */
         for (idx2 = 0, idx = 0; idx2 < start; ++idx2) {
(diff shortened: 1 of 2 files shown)


michalvasko (approved):

src/ly_common.c
```diff
@@ -863,10 +864,10 @@ ly_pat_compile_xmlschema_chblocks_xmlschema2perl(const char *pattern, char **reg
         if (idx) {
```
michalvasko, line 864: Better to change this to `ublock` as well?

tests/utests/types/string.c
```diff
@@ -817,6 +817,52 @@ test_data_xml(void **state)
...
+            "       pattern '[\\p{IsBasicLatin}\\p{IsLatin-1Supplement}\\p{IsLatinExtended-A}]+';"
+            "}} ");
+    UTEST_ADD_MODULE(schema, LYS_IN_YANG, NULL, NULL);
+    TEST_SUCCESS_XML("T_UB_6", "Árvíztűrő tükörfúrógép!", STRING, "Árvíztűrő tükörfúrógép!");
+
+    /* Unicode block test 7 - Unknown Unicode block */
+    schema = MODULE_CREATE_YANG("T_UB_7", "leaf port {type string { pattern '\\p{IsUnknownUnicodeBlock}+';} } ");
+    UTEST_INVALID_MODULE(schema, LYS_IN_YANG, NULL, LY_EVALID);
+    CHECK_LOG_CTX("Regular expression \"\\p{IsUnknownUnicodeBlock}+\" "
+                  "is not valid (\"UnknownUnicodeBlock}+\": unknown block name).", "/T_UB_7:port", 0);
```
michalvasko: Indentation issues such as this one, should have 8 spaces more as continuation of the previous line.

tests/utests/types/string.c
```diff
@@ -817,6 +817,52 @@ test_data_xml(void **state)
...
+
+    /* Unicode block test 8 - Unknown Unicode block with Basic Latin */
+    schema = MODULE_CREATE_YANG("T_UB_8", "leaf port {type string { "
+            "       pattern '[\\p{IsBasicLatin}\\p{IsUnknownUnicodeBlock}]+';"
+            "}} ");
+    UTEST_INVALID_MODULE(schema, LYS_IN_YANG, NULL, LY_EVALID);
+    CHECK_LOG_CTX("Regular expression \"[\\p{IsBasicLatin}\\p{IsUnknownUnicodeBlock}]+\" "
+                  "is not valid (\"UnknownUnicodeBlock}]+\": unknown block name).", "/T_UB_8:port", 0);
```
michalvasko: Another indentation issue.

michalvasko (changes requested):
Thanks, just some minor issues.

src/ly_common.c
manoe: As per my understanding, here the _idx_ variable stores the information, if the Unicode block is between [] brackets or not. I would not change that to _ublock_.

michalvasko: Okay, I think you are right.

tests/utests/types/string.c
manoe: Please check the latest commit with regards to the indentation issues.

michalvasko (approved):

michalvasko:
Please use `uncrustify` to correct the indentation and fix the tests. If you are really unable to, I can run it for you and them merge the changes manually.

PR #2509, opened 2026-04-10
plugins types BUGFIX print same IPv6 format as stored

jeremie6wind:
Hi,

The stored IPv6 address is sometimes not the same as the printed one.
For instance, if a compressed IPv6 address ::2222:1111 is stored, then a dual stack IPv6 address ::34.34.17.17 will be printed. This is due to the inet_ntop function implementation.
The aim of this patch is to store the IPv6 format and print the same format.
Store the IPv6 format with the value.
Introduce ipv6address_ip2str that converts back the IPv4 part if necessary.

Thank you
Best regards
Jeremie


src/plugins_types/ipv6_address.c
```diff
@@ -60,12 +61,13 @@ static void lyplg_type_free_ipv6_address(const struct ly_ctx *ctx, struct lyd_va
  * @param[in] ctx libyang context with dictionary.
  * @param[in,out] addr Allocated value for the address.
  * @param[out] zone Ipv6 address zone in dictionary.
+ * @param[out] dual stack format.
```
michalvasko, line 64: No description of the param.

src/plugins_types/ipv6_address.c
```diff
@@ -60,12 +61,13 @@ static void lyplg_type_free_ipv6_address(const struct ly_ctx *ctx, struct lyd_va
...
  * @param[out] err Error information on error.
  * @return LY_ERR value.
  */
 static LY_ERR
 ipv6address_str2ip(const char *value, uint32_t value_len, uint32_t options, const struct ly_ctx *ctx,
-        struct in6_addr *addr, const char **zone, struct ly_err_item **err)
+        struct in6_addr *addr, const char **zone, uint8_t *dual_stack_format, struct ly_err_item **err)
```
michalvasko, line 70: You should use `ly_bool` for boolean values and can then call the param just `dual_stack`, everywhere.

src/tree_data.h
```diff
@@ -635,14 +635,16 @@ struct lyd_value_ipv4_prefix {
  */
 struct lyd_value_ipv6_address_no_zone {
     struct in6_addr addr;   /**< IPv6 address in binary */
+    uint8_t dual_stack_format; /**< Dual stack IPv6 format or compressed format */
```
michalvasko, line 638: This is a problematic change. The header is public meaning this is an NBC change. I have just made the new release and it caused enough annoyance so too bad you have not made this change before it. Now it cannot be merged and we have not planned a new major SO version bump anytime soon.

michalvasko:
Thanks for the effort but I do not think this can be merged at all, unfortunately.

PR #2510, opened 2026-04-13
Unicode Specials block bug

manoe:
Libyang/yanglint cannot match characters of Unicode Specials block.

The function _ly_pat_compile_xmlschema_chblocks_xmlschema2perl()_ responsible to translate Yang patterns containing Unicode blocks  to PCRE2-compatible Perl-based regular expressions with character ranges assumes that each range literal is 19 char long. However, the Specials Unicode block is "special". It contains the disjoint U+FEFF character, and the range U+FFF0-U+FFFD. The original implementation only copies the first 19 chars of the literal, that is \x{FEFF}|\x{FFF0}. The expression is valid unfortunately, however only matches U+FEFF and U+FFF0.

The correction changes the _ublock2urange_ two dimensional char array to a struct array, which is populated during compilation time also with the individual literal lengths to preserve performance (assuming, that the original intention for hard-coding URANGE_LEN was to avoid _strlen_ calls).

Corresponding unit and component tests are also submitted.

### DIFF
--- a/src/ly_common.c
+++ b/src/ly_common.c
@@ -724,13 +724,13 @@ cleanup:
 static LY_ERR
 ly_pat_compile_xmlschema_chblocks_xmlschema2perl(const char *pattern, char **regex, struct ly_err_item **err)
 {
-struct ublock_s {
-    char *ublock;
-    char *urange;
-    size_t size;
-};
+    struct ublock_s {
+        char *ublock;
+        char *urange;
+        size_t size;
+    };
     struct ublock_s ublock2urange[] = {
-        {"BasicLatin", "[\\x{0000}-\\x{007F}]",19},
+        {"BasicLatin", "[\\x{0000}-\\x{007F}]", 19},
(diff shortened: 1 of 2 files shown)


michalvasko:
Okay, thanks, seems this should be fixed but no need to add the length into the struct, a few `strlen()` calls do not make any difference so please adjust it.

michalvasko (approved):

manoe:
CIFuzz keeps failing, however, based on the logs I cannot tell what went wrong. Can I get a little bit help with fuzzing?

manoe:
I've rebased my branch/PR to devel, but the CIFuzz keeps failing. Apparently, the binary exits _Failed to create context_. Could you take a look whats wrong? I would appreciate your input.

michalvasko:
Never mind the fuzz test, its failures are caused by recent changes of mine but please correctly rebase your branch to `devel`, my commits must not be part of your PR.

manoe:
Hi, I managed (with much help) to remove your commits from my PR.

michalvasko (approved):

PR #2515, opened 2026-05-12
build UPDATE avoid redefining HAVE_REGEX_H

jeremie6wind:
Avoids conflicts during build with external libraries that already define HAVE_REGEX_H.

Fixes: 0ba69b4377a8 ("build UPDATE optional regex dependency")


michalvasko (commented):
I am unsure about this change. Is it always okay to rely on some other check that the one in the library? Other possibility would be to rename all these macros and add `LY_` as their prefix, for example.

michalvasko (approved):

Issue #2516, opened 2026-05-12
Internal error when `when` depends on a node whose own `when` evaluated to false


szabolcs-szekely:
Hi,

we have a validation case where `yanglint` reports an internal error when one `when` expression depends on another data node whose own `when` condition has already evaluated to false. I tested this on `devel`, commit `2d34826`, using `yanglint -t config`.

The same state ambiguity also affects `must` expressions. A `must` depending on a `when`-guarded leaf can report:

```text
Must "..." depends on a node with a when condition, which has not been evaluated.
```

even though the referenced node's `when` was evaluated to false.

### Reproducer

`twm.yang`:

```yang
module twm {
  yang-version 1.1;
  namespace "urn:tests:twm";
  prefix twm;

  container top {
    leaf flag {
      type boolean;
      default "false";
    }

    leaf gated {
      when "../flag = 'true'";
      type string;
    }

    leaf checker {
      must "/twm:top/twm:gated = 'expected'" {
        error-message "checker requires gated=expected";
      }
      type string;
    }

    leaf dep-when {
      when "/twm:top/twm:gated = 'expected'";
      mandatory true;
      type string;
    }
  }
}
```

`data-internal-error.xml`:

```xml
<top xmlns="urn:tests:twm">
  <gated>x</gated>
</top>
```

Output on `devel`:

```text
libyang err : When condition "../flag = 'true'" not satisfied. (/twm:top/gated)
libyang err : Internal error (/libyang-devel-check/src/validation.c:1058).
YANGLINT[E]: Failed to parse input data file "data-internal-error.xml".
exit_code=1
```

`data-must.xml`:

```xml
<top xmlns="urn:tests:twm">
  <gated>x</gated>
  <checker>present</checker>
</top>
```

Output on `devel`:

```text
libyang err : When condition "../flag = 'true'" not satisfied. (/twm:top/gated)
libyang err : Must "/twm:top/twm:gated = 'expected'" depends on a node with a when condition, which has not been evaluated.
YANGLINT[E]: Failed to parse input data file "data-must.xml".
exit_code=1
```

With a local `LYD_WHEN_FALSE` correction, the second command reports the real `must` error instead:

```text
libyang err : When condition "../flag = 'true'" not satisfied. (/twm:top/gated)
libyang err : checker requires gated=expected (/twm:top/checker)
YANGLINT[E]: Failed to parse input data file "data-must.xml".
exit_code=1
```

### Why this looks wrong

[...]

PR #2528, opened 2026-06-03
Fix spurious cascading errors on when-false nodes in multi-error validation

szabolcs-szekely:
Relates to #2516.
  We want to use yanglint's multi-error mode (`LYD_VALIDATE_MULTI_ERROR`) for testing. Building on the crash fix in ff3cd623a, a `when`-false node was still indistinguishable from a not-yet-evaluated
  `when`: XPath returned `LY_EINCOMPLETE` for it, and `must`/leafref/child-`when` ran against the logically non-existent subtree, producing spurious cascading errors.
  Changes:
  - New `LYD_WHEN_FALSE` flag to mark a `when`-false node distinctly from a not-yet-evaluated one, cleared again if the `when` later resolves to true.
  - XPath treats `LYD_WHEN_FALSE` nodes as non-existent (`LY_ENOT`) instead of `LY_EINCOMPLETE`.
  - Validation skips `must`/obsolete/child/container-default checks for these nodes and removes their descendants from `node_types`/`node_when`.
  - Regression tests for the `when`/`must` cross-reference (incl. keyed-list hash lookup), the leafref-inside-`when`-false-container case, and a nested `when`-in-`when` case.

### DIFF
--- a/src/validation.c
+++ b/src/validation.c
@@ -440,7 +440,44 @@ lyd_validate_unres_when(struct lyd_node **tree, const struct lys_module *mod, st
                     /* only a warning */
                     LOGWRN(LYD_CTX(node), "When condition \"%s\" not satisfied.", disabled->cond->expr);
                 } else {
-                    /* invalid data */
+                    /* invalid data; mark the when as resolved-to-false so subsequent XPath
+                     * evaluations on this node return "no match" instead of LY_EINCOMPLETE */
+                    node->flags |= LYD_WHEN_FALSE;
+                    /* remove descendants from node_types so their leafrefs are not validated
+                     * against the when-false subtree, which would produce spurious cascading
+                     * errors (mirrors the cleanup done by lyd_validate_autodel_node_del) */
+                    if (node_types && node_types->count) {
+                        struct lyd_node *iter;
+                        uint32_t idx;
+
+                        LYD_TREE_DFS_BEGIN(node, iter) {
+                            if ((iter->schema->nodetype & LYD_NODE_TERM) &&
+                                    LYSC_GET_TYPE_PLG(((struct lysc_node_leaf *)iter->schema)->type->plugin_ref)->validate_tree &&
+                                    ly_set_contains(node_types, iter, &idx)) {
+                                ly_set_rm_index(node_types, idx, NULL);
+                            }
+                            LYD_TREE_DFS_END(node, iter);
+                        }
+                    }
+                    /* remove descendants from node_when so their own when conditions are not
+                     * evaluated; a when-false subtree is logically non-existent, so child
+                     * whens are moot and would otherwise produce spurious cascading errors */
+                    if (node_when->count > 1) {
+                        struct lyd_node *iter;
+                        uint32_t idx;
+
+                        count = node_when->count;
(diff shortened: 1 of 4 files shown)


src/validation.c
```diff
@@ -436,13 +436,52 @@ lyd_validate_unres_when(struct lyd_node **tree, const struct lys_module *mod, st
...
+                            if ((iter != node) && ly_set_contains(node_when, iter, &idx)) {
+                                ly_set_rm_index_ordered(node_when, idx, NULL);
+                            }
+                            LYD_TREE_DFS_END(node, iter);
+                        }
+                        if (count > node_when->count) {
+                            /* descendants were removed, refresh the iteration index */
+                            ly_set_contains(node_when, node, &i);
+                        }
+                    }
```
michalvasko: With this code, the function is too long, please put this into a separate function. Also, it may impact performance so why not use the new flag `LYD_WHEN_FALSE` only for `LYD_VALIDATE_MULTI_ERROR` validation?

michalvasko (changes requested):
Okay, I suppose, since it does not add too much code. But like I said before, this feature was added later and the original design did not anticipate it so it is expected it does not work perfectly. For a proper implementation, the validation would need significant refactoring, which seems not worth it.

src/validation.c
szabolcs-szekely: I separated the function and attached the `LYD_WHEN_FALSE` flag only for the `LYD_VALIDATE_MULTI_ERROR` mode with commit f78b246b76e26d46610c28674c3845dd7233871d

szabolcs-szekely:
@michalvasko what do you prefer, if I resolve your comment or you will resolve it when you are okay with it?

src/validation.c
```diff
@@ -436,13 +489,21 @@ lyd_validate_unres_when(struct lyd_node **tree, const struct lys_module *mod, st
                     /* only a warning */
                     LOGWRN(LYD_CTX(node), "When condition \"%s\" not satisfied.", disabled->cond->expr);
                 } else {
-                    /* invalid data */
+                    /* invalid data; only for multi-error validation keep the node in the tree
+                     * marked as when-false so XPath treats it as non-existent and validation can
+                     * continue without spurious cascading errors. In single-error validation this
+                     * is needless work because validation stops at this error. */
+                    if (val_opts & LYD_VALIDATE_MULTI_ERROR) {
```
michalvasko: To follow the logic of the code, moving the `if` to the `else` above would improve readability.

michalvasko (changes requested):
Just a minor improvement, should be good to go, otherwise.

michalvasko:
The changes are fine now, I have not replied sooner because of a vacation of mine.

src/validation.c
szabolcs-szekely: Corrected with de9ac4179d6c4c2acec583e93eefe69dedd8a36f commit.

szabolcs-szekely:
Rebased on top of devel and corrected all comments. Thanks

michalvasko (approved):

szabolcs-szekely:
Thank you. When do you plan to release a new version? We would like to use and integrate the correction.

michalvasko:
Ideally next week but I may not have enough time for that so maybe the week after.

PR #2529, opened 2026-06-10
Refactor utests: replace internal functions with API and clean up

HanzlikPetr:


tests/utests/basic/test_common.c
```diff
@@ -14,390 +14,362 @@
...
+test_utf8(void **state)
 {
-    char buf[5] = {0};
-    const char *str = buf;
-    unsigned int c;
-    uint32_t len;
+    int target_idx = 59;
+    struct ly_ctx *ctx = UTEST_LYCTX;
+    // array coresponds to the following YANG module:
+    // module test { namespace "urn:test"; prefix t; description "XXX"; }
```
Roytak: Use block comment.

tests/utests/types/instanceid.c
```diff
@@ -18,7 +18,13 @@
...
 /* LOCAL INCLUDE HEADERS */
 #include "libyang.h"
-#include "path.h"
+
+enum ly_path_pred_type {
+    LY_PATH_PREDTYPE_POSITION,  /**< position predicate - [2] */
+    LY_PATH_PREDTYPE_LIST,      /**< keys predicate - [key1='val1'][key2='val2']... */
+    LY_PATH_PREDTYPE_LEAFLIST,  /**< leaflist value predicate - [.='value'] */
+    LY_PATH_PREDTYPE_LIST_VAR   /**< keys predicate with variable instead of value - [key1=$USER]... */
+};
```
Roytak: Just copying the enum from path.h verbatim feels very fragile, modifying this enum in path.h would break this test. Also the actual value comparison was removed from `CHECK_LYD_VALUE_INST` and only the array sizes are used - these values are never used. A plain `int` array would serve the same purpose. Also this same enum was also copied to `union.c`. My suggestion is to either:
- add the value validation back to CHECK_LYD_VALUE_INST and `#define` these values in `utests.h` (based on a comment in `path.h`: `"/**< Predicate type (see YANG ABNF) */"`, it would seem that these values are standardized, but I am not sure about that, you may try to find out more about it)
- remove the enums from the tests completely, since the values are not checked anyways, and just pass integer arrays to CHECK_LYD_VALUE_INST

tests/utests/types/union.c
```diff
@@ -18,7 +18,13 @@
...
 /* LOCAL INCLUDE HEADERS */
 #include "libyang.h"
-#include "path.h"
+
+enum ly_path_pred_type {
+    LY_PATH_PREDTYPE_POSITION,  /**< position predicate - [2] */
+    LY_PATH_PREDTYPE_LIST,      /**< keys predicate - [key1='val1'][key2='val2']... */
+    LY_PATH_PREDTYPE_LEAFLIST,  /**< leaflist value predicate - [.='value'] */
+    LY_PATH_PREDTYPE_LIST_VAR   /**< keys predicate with variable instead of value - [key1=$USER]... */
+};
```
Roytak: Related to the comment in `instanceid.c`.

tests/utests/utests.h
```diff
@@ -232,7 +233,7 @@ struct utest_context {
     assert_non_null(NODE); \
     assert_int_equal((NODE)->basetype, TYPE); \
     CHECK_ARRAY((NODE)->exts, EXTS); \
-    assert_ptr_equal((NODE)->plugin_ref, lyplg_type_plugin_find(NULL, "", NULL, ly_data_type2str[TYPE]))
+    assert_non_null((NODE)->plugin_ref)
```
Roytak: This checks seems much weaker, would lysc_get_type_plugin work here possibly?

tests/utests/utests.h
```diff
@@ -87,6 +86,8 @@
  * function call.
  */
 
+#define UNUSED(x) UNUSED_ ## x __attribute__((__unused__))
+
```
Roytak: I see that the macro `UNUSED` is defined in `compat.h`, consider including it instead.

tests/utests/schema/test_yang.c
```diff
@@ -18,493 +18,290 @@
...
+    assert_int_equal(LY_SUCCESS, lys_parse_mem(UTEST_LYCTX, schema, LYS_IN_YANG, &mod));
+    assert_string_equal("hello", mod->dsc);
+
+    schema = "module test14 {\n"
+            "  namespace urn:test14;\n"
+            "  prefix t;\n"
+            "  description \t\n\"he\"+\'llo\';\n"
+            "}";
+    assert_int_equal(LY_SUCCESS, lys_parse_mem(UTEST_LYCTX, schema, LYS_IN_YANG, &mod));
+    assert_string_equal("hello", mod->dsc);
```
Roytak: These all seem to be testing the same thing.

tests/utests/schema/test_schema.c
```diff
@@ -1667,76 +1675,6 @@ test_extension_argument_element(void **state)
 
 }
 
-static void
-test_extension_compile(void **state)
-{
-    struct lys_module *mod;
```
Roytak, line 1673: Can this not be rewritten via public API? For example parsing a module with `md:annotation` containing substatements (or other ext like structure), then inspecting the compiled `lysc_ext_instance` and verifying the compiled substatement could possibly work.

tests/utests/schema/test_tree_schema_compile.c
```diff
@@ -1701,49 +1693,31 @@ test_type_leafref(void **state)
...
-    lyxp_expr_free(expr);
+    char yang[512];
+    size_t i;
+    const char *invalid[] = {
+        "invalid_path",
+        "..[",
+        "../",
+        "/",
+        "../../pref:id/xxx[predicate]/invalid!!!"
+    };
```
Roytak, line 1705: Is there a reason why ".." was excluded? It was tested previously.

tests/utests/types/inet_types.c
```diff
@@ -311,10 +313,11 @@ test_plugin_sort(void **state)
 
     /* ipv6-prefix */
     lysc_type = ((struct lysc_node_leaflist *)mod->compiled->data->next->next->next->next->next)->type;
-    v1 = "::C:D:E:f:a/96";
+    type = lysc_get_type_plugin(lysc_type->plugin_ref);
+    v1 = "::C:D:E:f:a/112";
     assert_int_equal(LY_SUCCESS, type->store(UTEST_LYCTX, lysc_type, v1, strlen(v1) * 8,
```
Roytak, line 318: Why were the values swapped here?

tests/utests/data/test_tree_data_sorted.c
```diff
@@ -1268,9 +1284,6 @@ test_lyds_free_metadata(void **state)
     assert_int_equal(lyd_new_term(NULL, mod, "ll", "1", 0, &first), LY_SUCCESS);
     assert_int_equal(lyd_new_term(NULL, mod, "ll", "2", 0, &node), LY_SUCCESS);
     assert_int_equal(lyd_insert_sibling(first, node, NULL), LY_SUCCESS);
-    lyds_free_metadata(first);
-    assert_null(first->meta);
-    assert_null(node->meta);
 
     lyd_free_all(first);
 }
```
Roytak: It seems this test doesn't really test anything now, consider removing/adjusting it.

tests/utests/utests.h
```diff
@@ -1004,11 +1003,6 @@ struct utest_context {
     { \
         LY_ARRAY_COUNT_TYPE arr_size = sizeof(VALUE) / sizeof(VALUE[0]); \
         assert_int_equal(arr_size, LY_ARRAY_COUNT((NODE).target)); \
-        for (LY_ARRAY_COUNT_TYPE it = 0; it < arr_size; it++) { \
-            if ((NODE).target[it].predicates) { \
-                assert_int_equal(VALUE[it], (NODE).target[it].predicates[0].type); \
-            } \
-        } \
     }
```
Roytak, line 1005: See comment in `instanceid.c`.

Roytak (commented):
Noticed a couple of possible places for improvement, but very good overall.

tests/utests/utests.h
HanzlikPetr: My specific task for this pull request was to rewrite these tests strictly using the public API and to decouple them from internal structures.

I defined the UNUSED macro here locally specifically to avoid pulling in compat.h, since it's an internal header and I was instructed to strictly respect that public API boundary.

tests/utests/schema/test_schema.c
HanzlikPetr: I actually discussed this specific test with Michal Vasko, and he pointed out that this compilation logic is already naturally triggered and tested every time a context is loaded.

tests/utests/types/inet_types.c
HanzlikPetr: The type->sort behavior changed from string comparison to numeric comparison. I swapped them so that val1 is still mathematically greater than val2 for the assertions further down.

tests/utests/basic/test_context.c
```diff
@@ -16,12 +16,20 @@
...
+    int count = 0;
+
+    while (list[count]) {
+        count++;
+    }
+    return count;
+}
 
 #ifdef _WIN32
+# define PATH_SEPARATOR ";"
```
michalvasko, line 32: Var seems to be unused.

michalvasko (changes requested):
Please rebase the current `devel` changes.

tests/utests/basic/test_context.c
jktjkt: yeah, when I added it, I was designing for its inclusion via `src/ly_common.h`.

HanzlikPetr: [look here](https://github.com/CESNET/libyang/blob/fe2591d797e592a5a159f87c3645cc1db973b6c9/tests/utests/basic/test_context.c#L702)

src/ly_common.h
```diff
@@ -850,6 +827,7 @@ LY_ERR ly_val_get_quot(const struct ly_ctx *ctx, const char *value, char *quot);
 LY_ERR ly_append_str(char **str, uint32_t *size, uint32_t *used, const char *format, ...);
 
 /**
+49954cdbe (utests basic UPDATE replace internal functions with API)
```
michalvasko: Seems like a typo.

michalvasko (changes requested):

michalvasko (approved):

PR #2530, opened 2026-06-11
plugins types UPDATE pass ipv6 structure object

jeremie6wind:
Give the object instead of the two fields to the str2ip functions.

### DIFF
--- a/src/plugins_types/ipv6_address.c
+++ b/src/plugins_types/ipv6_address.c
@@ -58,14 +58,13 @@ static void lyplg_type_free_ipv6_address(const struct ly_ctx *ctx, struct lyd_va
  * @param[in] value_len Length of @p value.
  * @param[in] options Type store callback options.
  * @param[in] ctx libyang context with dictionary.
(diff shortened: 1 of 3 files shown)


michalvasko (commented):
What is the point of this PR since it is not changing the functionality?

jeremie6wind:
Hi MIchal,
Few weeks ago I proposed the following PR to display the IPv6 format that same way it was set:
https://github.com/CESNET/libyang/pull/2509

As a reminder, this IPv6 address test would fail:
`    TEST_SUCCESS_XML("a", "l4", "::2121:3737", STRING, "::2121:3737");`

However the previous PR was NBC.

If this PR was accepted, I could have a simple unmerged patch with my dual_stack_format field.
If you have a better idea, I'd love to hear it.
Best regards.

michalvasko:
> If this PR was accepted, I could have a simple unmerged patch with my dual_stack_format field.

Fair enough, I just wanted to know the reason.

PR #2531, opened 2026-06-16
yanglint: Add commands for sample skeleton generation and IETF validation

HanzlikPetr:


tools/lint/cmd.h
```diff
@@ -276,6 +276,27 @@ int cmd_print_dep(struct yl_opt *yo, int posc);
...
+/**
+ * @copydoc cmd_add_opt
+ */
+int cmd_sample_opt(struct yl_opt *yo, const char *cmdline, char ***posv, int *posc);
+/**
+ * @copydoc cmd_add_dep
+ */
+int cmd_sample_dep(struct yl_opt *yo, int posc);
+void cmd_sample_help(void);
+/**
```
Roytak, line 293: Missing a new line.

tools/lint/cmd.h
```diff
@@ -391,4 +412,26 @@ int cmd_debug_store(struct ly_ctx **ctx, struct yl_opt *yo, const char *posv);
 int cmd_debug_setlog(struct ly_ctx *ctx, struct yl_opt *yo);
 void cmd_debug_help(void);
 
+/* cmd_ietf.c */
+/**
+ * @copydoc cmd_add_opt
+ */
+int cmd_ietf_opt(struct yl_opt *yo, const char *cmdline, char ***posv, int *posc);
+/**
+ * @copydoc cmd_add_dep
```
Roytak: Missing a new line.

tools/lint/main_ni.c
```diff
@@ -227,6 +234,9 @@ help(int shortout)
             "                Unsupported for the Release build\n\n"
 #endif
           );
+
+    printf("  -t, --ietf\n"
+            "                Enable strict IETF validation rules for the loaded schemas.\n\n");
 }
```
Roytak: Should be `-T` I guess.

tools/lint/cmd_ietf.c
```diff
@@ -0,0 +1,711 @@
...
+ * @param[in] tpdfs Array of top-level parsed `typedef` statements.
+ * @param[in] grps Linked list of top-level parsed `grouping` statements.
+ * @param[in,out] out The libyang output handler where warnings and errors are printed.
+ * @param[in,out] found_2119 Pointer to a flag tracking if RFC 2119 keywords were found anywhere in this specific file.
+ */
+static void
+check_parsed_boilerplate(const char *name, const char *filepath, const char *dsc, const char *contact, const char *org, struct lysp_revision *revs, int type, struct lysp_ext *exts,
+        struct lysp_feature *feats, struct lysp_ident *idents, struct lysp_node_augment *augments, struct lysp_tpdf *tpdfs, struct lysp_node_grp *grps, struct ly_out *out, int *found_2119)
+{
+    const char *file_name = filepath ? strrchr(filepath, '/') + 1 : name;
```
Roytak: If filepath is non-NULL, but containts no '/', `strrchr` returns NULL, and NULL + 1 is UB. Check if this can ever happen and consider adding some guards/extracting to a function, because the same pattern can be found 3 times in this file.

tools/lint/main_ni.c
```diff
@@ -818,6 +843,20 @@ main_ni(int argc, char *argv[])
                 goto cleanup;
             }
         }
+    } else if (yo.data_out_format) {
+        for (u = 0; u < yo.schema_modules.count; ++u) {
+            yo.last_one = (u + 1) == yo.schema_modules.count;
+            if ((ret = cmd_sample_exec(&ctx, &yo, ((struct lys_module *)yo.schema_modules.objs[u])->name))) {
+                goto cleanup;
+            }
+        }
```
Roytak: Sample should probably use it's own dedicated flag instead of `yo.data_out_format`, because this flag is also used by format (-f). So it seems to me that if a user passes `-f`, it will now also print a sample skeleton before processing data.

tools/lint/cmd_sample.c
```diff
@@ -0,0 +1,559 @@
...
+        ly_print(out, "/>\n");
+        break;
+    }
+}
+
+static void
+print_xml_nodes(struct ly_out *out, const struct lysc_node *node, const char *ns, int *space_count)
+{
+    const uint32_t PRINT_MASK = LYS_CONTAINER | LYS_LIST | LYS_LEAF | LYS_ANYXML | LYS_ANYDATA;
+    int nodetype;
```
Roytak, line 216: It would seem that `LYS_LEAFLIST` is missing compared to the JSON implementation, which has it.

tools/lint/cmd_sample.c
```diff
@@ -0,0 +1,559 @@
...
+}
+
+int
+cmd_sample_exec(struct ly_ctx **ctx, struct yl_opt *yo, const char *posv)
+{
+    const struct lys_module *mod;
+    char *revision, *name;
+    uint8_t out_alloc = 0;
+
+    name = strdup(posv);
```
Roytak, line 518: Check for NULL.

tools/lint/ietf.c
```diff
@@ -0,0 +1,711 @@
...
+}
+
+int
+cmd_ietf_exec(struct ly_ctx **ctx, struct yl_opt *yo, const char *posv)
+{
+    const struct lys_module *mod;
+    char *revision, *name;
+    uint8_t out_alloc = 0;
+
+    name = strdup(posv);
```
Roytak, line 1012: Missing NULL check.

tools/lint/ietf.c
```diff
@@ -0,0 +1,711 @@
...
+ *
+ * @param[in,out] out The libyang output handler where warnings and errors are printed.
+ * @param[in] data The first top-level node of the compiled data tree to be inspected.
+ * @param[in] file_name The name of the file being evaluated (used to prefix the error messages).
+ */
+static void
+check_nodes_ietf(struct ly_out *out, const struct lysc_node *data, const char *file_name)
+{
+    const struct lysc_node *root, *elem;
+    char node_path[256];
```
Roytak, line 222: Seems not to be used anywhere, so either remove it or use it.

tools/lint/cmd_sample.c
```diff
@@ -0,0 +1,559 @@
...
+ *
+ * @param[in,out] out The libyang output handler to print to.
+ * @param[in] node The compiled schema node currently being processed.
+ * @param[in] ns The XML namespace string to apply, or NULL if it is inherited.
+ * @param[in] space_count Pointer to the current indentation level (number of spaces).
+ */
+static void
+handle_xml_default(struct ly_out *out, const struct lysc_node *node, const char *ns, int *space_count)
+{
+    switch (node->nodetype) {
```
Roytak, line 149: I think that the `ns` parameter is not used and should be removed. All callers but the recursive `LYS_CHOICE` pass NULL and when it's `LYS_LEAFLIST`, then `node->module->ns` is used instead. Verify this.

tools/lint/cmd_ietf.c
```diff
@@ -0,0 +1,711 @@
...
+ */
+static int
+check_words(const char *dsc)
+{
+    regex_t regex;
+    int ret;
+
+    // The regex pattern checks for exact uppercase words surrounded by word boundaries (spaces, punctuation, or string ends)
+    const char *pattern = "(^|[^a-zA-Z])(MUST|MUST NOT|REQUIRED|SHALL|SHALL NOT|SHOULD|SHOULD NOT|RECOMMENDED|NOT RECOMMENDED|MAY|OPTIONAL)([^a-zA-Z]|$)";
+
```
Roytak: Block comment, please. Also maybe in this case just declare the variable first, and then set the value below, with the comment above.

tools/lint/cmd_sample.c
```diff
@@ -0,0 +1,559 @@
...
+ */
+
+#define _GNU_SOURCE
+
+#include "cmd.h"
+
+#include <getopt.h>
+#include <stdint.h>
+#include <string.h>
+#include <strings.h>
```
Roytak: Are there any functions used from `strings.h`?

tools/lint/cmd_ietf.c
```diff
@@ -0,0 +1,711 @@
...
+ *
+ * @param[in] text The target string to search within. If NULL, the function returns 0.
+ * @param[in] pattern The POSIX extended regular expression pattern to match against.
+ * @return 1 if the pattern is successfully found within the text.
+ * @return 0 if the pattern is not found, if a compilation error occurs, or if invalid NULL arguments are passed.
+ */
+static int
+check_regex(const char *text, const char *pattern)
+{
+    regex_t regex;
```
Roytak: Consider initializing.

tools/lint/cmd_ietf.c
```diff
@@ -0,0 +1,711 @@
...
+ */
+static void
+check_nodes_ietf(struct ly_out *out, const struct lysc_node *data, const char *file_name)
+{
+    const struct lysc_node *root, *elem;
+    char node_path[256];
+
+    LY_LIST_FOR(data, root) {
+
+        int is_mandatory = 0;
```
Roytak: Remove the extra NL, and possibly move var definition to the beginning and just set is_mandatory to 0 here everytime.

tools/lint/cmd_ietf.c
```diff
@@ -0,0 +1,711 @@
...
+/**
+ * @brief Orchestrator for IETF RFC 8407 compliance validation on a YANG module.
+
+ * @param[in] mod The libyang module structure containing both the parsed schema
+ * @param[in,out] out The libyang output handler where compliance warnings and errors are printed.
+ */
+static void
+check_ietf(const struct lys_module *mod, struct ly_out *out)
+{
+    const char *file_name = mod->filepath ? strrchr(mod->filepath, '/') + 1 : mod->name;
```
Roytak: Same as above.

tools/lint/cmd_ietf.c
```diff
@@ -0,0 +1,711 @@
...
+            mod->parsed->groupings, out, &main_found_2119);
+
+    snprintf(ns, sizeof(ns), "urn:ietf:params:xml:ns:yang:%s", mod->name);
+    if (strcmp(ns, mod->ns)) {
+        IETF_WARN(file_name, "8407", "4.9", out, "namespace value should be \"%s\"", ns);
+    }
+
+    LY_ARRAY_FOR(mod->parsed->includes, i) {
+        struct lysp_submodule *sub = mod->parsed->includes[i].submodule;
+        const char *sub_file = sub->filepath ? strrchr(sub->filepath, '/') + 1 : sub->name;
```
Roytak: Same as above with the path, also consider moving var definitions to the beginning.

tools/lint/cmd_ietf.c
```diff
@@ -0,0 +1,711 @@
...
+    LY_ARRAY_FOR(tpdfs, i) {
+        if (tpdfs[i].dsc == NULL) {
+            IETF_ERR(file_name, "8407", "4.13, 4.14", out, "statement \"typedef\" must have a \"description\" substatement");
+        } else if (check_words(tpdfs[i].dsc)) {
+            *found_2119 = 1;
+        }
+    }
+
+    LY_LIST_FOR(grps, grp) {
+        if (grp->dsc == NULL) {
```
Roytak: Some code in this PR uses the short check, i.g. `!grp->dsc` while others the long one, i.g. `grp->dsc = NULL`. You should almost always use the first one, since it's libyang idiomatic.

tools/lint/cmd_sample.c
```diff
@@ -0,0 +1,559 @@
...
+ * @brief Initialize and print the top-level XML skeleton structure.
+ *
+ * @param[in,out] out  The libyang output handler to print to.
+ * @param[in] root The root compiled schema node of the module being processed.
+ */
+static void
+print_xml_skeleton(struct ly_out *out, struct lysc_node *root)
+{
+    struct lysc_node *node = root;
+    int spaces = 2;
```
Roytak: Consider a macro instead of a random magical `2`, there are multiple occurrences.

tools/lint/main_ni.c
```diff
@@ -52,6 +52,9 @@ help(int shortout)
             "    yanglint [-f { xml | json }] <schema>... <file>...\n"
             "        Validates the YANG modeled data <file>(s) according to the <schema>(s) optionally\n"
             "        printing them in the specified format.\n\n"
+            "   yanglint -S { xml | json } <schema>...\n"
```
Roytak: Needs one more space.

Roytak (commented):
Mostly smaller things, good overall.

tools/lint/ietf.c
HanzlikPetr: use on the end of check_ietf function

tools/lint/cmd.h
```diff
@@ -276,6 +276,28 @@ int cmd_print_dep(struct yl_opt *yo, int posc);
 int cmd_print_exec(struct ly_ctx **ctx, struct yl_opt *yo, const char *posv);
 void cmd_print_help(void);
 
+/* cmd_sample.c */
+
+/**
+ * @copydoc cmd_add_opt
+ */
+int cmd_sample_opt(struct yl_opt *yo, const char *cmdline, char ***posv, int *posc);
```
michalvasko, line 285: Missing newline after this one.

tools/lint/cmd_sample.c
```diff
@@ -0,0 +1,562 @@
...
+        {NULL, 0, NULL, 0}
+    };
+
+    if ((rc = parse_cmdline(cmdline, &argc, &argv))) {
+        return rc;
+    }
+
+    /* reset getopt */
+    optind = 0;
+    while ((opt = getopt_long(argc, argv, "f:ho:", options, &opt_index)) != -1) {
```
michalvasko: Please keep the order of the chars the same as in `options` to easily keep track of them.

tools/lint/cmd_sample.c
```diff
@@ -0,0 +1,562 @@
...
+    case LYS_LIST:
+    case LYS_ANYDATA:
+    case LYS_ANYXML:
+        ly_print(out, "/>\n");
+        break;
+    }
+}
+
+static void
+print_xml_nodes(struct ly_out *out, const struct lysc_node *node, int *space_count)
```
michalvasko: Why is `space_count` an out parameter? It should be simpler to just use an `int space_count`, no need to decrease the count then. Also, missing `doxygen`.

tools/lint/cmd_sample.c
```diff
@@ -0,0 +1,562 @@
...
+        return 0;
+
+    case LY_TYPE_INT64:
+    case LY_TYPE_UINT64:
+    case LY_TYPE_DEC64:
+        return 1;
+
+    default:
+        return 1;
+    }
```
michalvasko, line 315: Based on what were the specific types selected and rest put into `default`? Best to just enumerate all the types and not use any default.

tools/lint/ietf.c
```diff
@@ -0,0 +1,649 @@
...
+check_regex(const char *text, const char *pattern)
+{
+    void *pcre2_code = NULL;
+    int ret;
+
+    if (!text || !pattern) {
+        return 0;
+    }
+
+    if (ly_pattern_compile(NULL, pattern, &pcre2_code) != 0) {
```
michalvasko: Read the docs, but I think you do not have to call `ly_pattern_compile()` if you do not want to reuse the pattern code, just pass `NULL` to `ly_pattern_match()`.

tools/lint/ietf.c
```diff
@@ -0,0 +1,649 @@
...
+check_nodes_ietf(struct ly_out *out, const struct lysc_node *data, const char *file_name)
+{
+    const struct lysc_node *root, *elem;
+    char node_path[256];
+    int is_mandatory;
+
+    LY_LIST_FOR(data, root) {
+        is_mandatory = 0;
+
+        if (root->flags & LYS_MAND_TRUE) {
```
michalvasko: This check is enough, the other ones are covered by this, the flag is set for those other cases as well.

tools/lint/ietf.c
```diff
@@ -0,0 +1,649 @@
...
+
+/**
+ * @brief Helper macro to validate locally scoped typedefs and groupings inside parsed data nodes.
+ *
+ * @param node_ptr Pointer to the parsed node structure containing the `typedefs` and `groupings` arrays.
+ * @param node_type_str A string literal representing the node type (e.g., "container", "list") for the error message.
+ * @param node_name The string name of the parent node to provide context in the error message.
+ * @param found_2119 A pointer to an integer flag that tracks if RFC 2119 keywords are used in the module/submodule.
+ */
+#define CHECK_INNER_DEFS(node_ptr, node_type_str, node_name, found_2119) \
```
michalvasko: Please refactor to a function, macro too complex.

tools/lint/ietf.h
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @brief Validate a module against the IETF requirements.
+ *
+ * @param[in,out] ctx A double pointer to the libyang context containing the loaded modules.
+ * @param[in,out] yo A pointer to the yanglint options structure, which manages global
+ * @param[in] posv The positional command-line argument provided by the user, representing the target module name (e.g., "ietf-interfaces" or "ietf-interfaces@2014-05-08").
+ * @return 0 on successful execution of the validation pipeline (note: this returns 0 even if IETF compliance warnings/errors were printed to the output).
+ * @return 1 on fatal execution errors (e.g., module not found, module failed to compile, or memory allocation failure).
+ */
+int ietf(struct ly_ctx **ctx, struct yl_opt *yo, const char *posv);
```
michalvasko: Please use clear names with the local prefix even for function in the internal API. `yanglint` uses `yl` prefix, so `yl_validate_ietf`, for example.

tools/lint/main_ni.c
```diff
@@ -507,6 +523,8 @@ process_args(int argc, char *argv[], struct yl_opt *yo, struct ly_ctx **ctx)
         {"extended-leafref",  no_argument,       NULL, 'X'},
         {"json-null",         no_argument,       NULL, 'J'},
         {"debug",             required_argument, NULL, 'G'},
+        {"sample-skeleton",   required_argument, NULL, 'S'},
+        {"ietf",              no_argument,       NULL, 1000},
```
michalvasko: Please always use a valid single-char variant for every command.

michalvasko (changes requested):

tools/lint/main_ni.c
```diff
@@ -518,7 +536,7 @@ process_args(int argc, char *argv[], struct yl_opt *yo, struct ly_ctx **ctx)
     yo->line_length = 0;
 
     opterr = 0;
-    while ((opt = getopt_long(argc, argv, "hvVQf:I:p:DF:iP:qs:neE:At:d:lL:o:O:R:myY:XJx:G:", options, &opt_index)) != -1) {
+    while ((opt = getopt_long(argc, argv, "hvVQf:I:p:DF:iP:qs:neE:At:d:lL:o:O:R:myY:XJx:G:S:", options, &opt_index)) != -1) {
```
michalvasko: Missing the `T` short argument handling.

michalvasko (changes requested):
I appreciate the RFC sections for each warning or error.

michalvasko (approved):

michalvasko:
Please mention the `--ietf` parameter in the non-interactive `yanglint` help.

michalvasko (approved):

PR #2547, opened 2026-08-05
fix warning message for YANG module revision ordering

jktjkt:
The [standard](https://datatracker.ietf.org/doc/html/rfc7950#section-7.1.9) says that:

> [...] a new [revision] SHOULD be added in front of the revisions
> sequence so that all revisions are in reverse chronological order

The actual error message was misleading, saying that an older revision was found after a new revision. That is actually a correct order. The check was correct, but the error message should have said that the older revision was found *before* a newer one.

We actually "fixed" our YANG models based on this diagnostic message, only to revert that later on.


michalvasko (approved):

michalvasko:
True, sorry about that.

PR #2557, opened 2026-08-18
printer cbor BUGFIX crash when printing (leaf-)lists nested in a list

manoe:
The CBOR printer kept a single "currently open array" pointer (cborpr_ctx.array) shared across all nesting levels, while tracking the open nodes on a proper stack (cborpr_ctx.open). When a leaf-list or list was nested inside a list, printing the inner array overwrote that pointer and then reset it to NULL, so the enclosing list subsequently pushed into a NULL array, causing a segfault in cbor_array_push().

Replace the single array pointer with a stack (cborpr_ctx.arrays) kept parallel to the open-node stack, add cbor_current_array() to fetch the innermost open array, and update cbor_print_leaf_list()/cbor_print_opaq() and their callers accordingly. The stack is freed on all exit paths of cbor_print_data().

Add a regression test (test_nested_list) covering a leaf-list nested in a list, a list nested in a list, and mixed multi-instance nesting, plus the supporting list-nested node in the cbor-test module.

### DIFF
--- a/src/printer_cbor.c
+++ b/src/printer_cbor.c
@@ -51,7 +51,8 @@ struct cborpr_ctx {
     const struct lyd_node *first_leaflist;  /**< first printed leaf-list instance, used when printing its metadata/attributes */
 
     cbor_item_t *root_map;                  /**< root CBOR map */
-    cbor_item_t *array;                     /**< currently open CBOR array for leaf-list/list instances */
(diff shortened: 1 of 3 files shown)


michalvasko (approved):

michalvasko:
Thanks, looks fine.

PR #2561, opened 2026-08-24
printer sid BUGFIX include nodes augmented into other modules

manoe:
# printer sid BUGFIX: include nodes augmented into other modules

## Summary

The RFC 9595 SID generator does not assign SIDs to data nodes that a module
contributes to **other** modules via `augment`. For any module that places its
data by augmenting another module, `lys_sid_gen()`/`lys_sid_update()` (and
`yanglint -g`) produce a `.sid` file containing only the `module` item and **no
data items**. This PR makes the generator collect those augmented-in nodes.

## Problem

`sid_collect_items()` gathered data-namespace items only from the processed
module's own compiled tree:

```c
LY_CHECK_RET((rc = lysc_module_dfs_full(module, collect_data_cb, callback_data)), rc);
```

Nodes that `module` augments into another module live in the **target** module's
compiled tree (their `node->module` still points back to the augmenting
`module`), so `lysc_module_dfs_full(module, ...)` never visits them. As a result
they receive no SID.

Minimal example — a module whose only data is an augment into another module:

```yang
module a1 {
  import b1 { prefix b1; }
  augment "/b1:cont" {
    leaf x { type string; }
  }
}
```

Before this PR, `a1`'s generated `.sid` contains a single item (`module a1`) and
**0 data items**; the augmented-in `/b1:cont/a1:x` is missing.

## Fix

Collect all nodes **defined by** the processed module, wherever they are grafted
in the compiled schema:

- `collect_data_cb()` now attributes each node to its defining module and skips
  nodes whose `node->module` differs from the module being processed. This both
  (a) lets us safely traverse other modules' trees and pick only the
  augmented-in nodes, and (b) fixes a latent mis-attribution where a module's own
  `.sid` could wrongly include nodes augmented into it by others.
- After walking the module's own tree, `sid_collect_items()` also walks every
  context module whose `augmented_by` references the processed module (guarded on
  `->compiled`), using the same callback.

libyang already implements augment target modules when the augmenting module is
implemented (`lys_precompile_augments_deviations()`), so the target trees are
present during generation. The existing choice/case handling and top-level
extension-data (`yang-data`/`structure`) traversal are unchanged.

## Backward compatibility

Output is unchanged for modules that do not augment other modules. The only
behavioral change is correct attribution: augmented-in nodes now appear in the
augmenting module's `.sid`, and are no longer (incorrectly) included in the
target module's `.sid`.

## Notes

- Deviation-introduced nodes (`deviated_by`) are a separate RFC 9595 concern and
  are intentionally out of scope here.

### DIFF
--- a/src/printer_sid.c
+++ b/src/printer_sid.c
@@ -50,6 +50,7 @@ struct sid_collect_data {
     uint64_t count;           /**< Number of items collected so far; exceeding @p max, surplus items are only counted. */
     uint64_t max;             /**< Maximum number of items (capacity of the array). */
     struct ly_ctx *ctx;       /**< libyang context used for error logging. */
+    const struct lys_module *module;  /**< Module whose nodes are collected; used to attribute augmented nodes to their defining module. */
 };
 
 /**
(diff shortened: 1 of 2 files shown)


michalvasko (approved):

michalvasko:
Thanks, seems fine. The RFC does not mention explicitly how to process augments but this should be correct.

PR #2565, opened 2026-08-26
tree data: duplicate anydata children into target context

Yeaseen:
## Summary

Fix a use-after-free in cross-context duplication of data trees containing
`anydata` XML subtrees.

This addresses `GHSA-jq97-j23h-2vqw`:
https://github.com/CESNET/libyang/security/advisories/GHSA-jq97-j23h-2vqw

## Problem

`lyd_dup_siblings_to_ctx()` is expected to duplicate a data tree into the target
context. However, when duplicating an `anydata` node that stores its value as a
child subtree, the embedded children were copied with `lyd_dup_siblings()`.

That preserves references to the source context. If the source tree and source
context are freed after successful duplication, later operations on the duplicate
can access freed memory. One affected path is XPath traversal through
`lyd_trim_xpath()`.

## Fix

Duplicate `anydata` child subtrees with `lyd_dup_siblings_to_ctx()` and
`LYD_CTX(trg)` so embedded children are copied into the same target context as
the containing `anydata` node.

Add a regression test that duplicates an `anydata` XML subtree into a second
context, destroys the original context, and then evaluates XPath on the
duplicate.

## Verification

I ran the advisory reproducer against an ASan/UBSan build with this fix applied.
The reproducer completes successfully and no use-after-free is reported.

### DIFF
--- a/tests/utests/data/test_tree_data.c
+++ b/tests/utests/data/test_tree_data.c
@@ -282,6 +282,39 @@ test_compare_diff_ctx(void **state)
 }
 
 static void
+test_dup_anydata_to_ctx(void **state)
+{
+    struct ly_ctx *ctx1 = NULL, *ctx2 = NULL;
+    struct lyd_node *tree = NULL, *dup = NULL;
+    const char *schema;
+    const char *data;
+
+    (void)state;
+
+    schema = "module c {yang-version 1.1; namespace \"urn:tests:c\"; prefix c;"
+            "container root {anydata payload;}}";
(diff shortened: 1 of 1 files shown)


michalvasko:
Already fixed in the `devel` branch but if you rebase (remove the fix), I can still merge the test.

Yeaseen:
@michalvasko Done, thanks. I rebased onto `devel`, removed the duplicate fix, and kept only the regression test.

PR #2574, opened 2026-09-18
yanglint BUGFIX const qualifier violation in cmd_load

mferreirarivier:
cmd_load_exec is called with an immutable string as posv. Cast of strchr return silently discards that const qualifier, so posv ends up modified in cmd_load regardless of it.

Duplicate an owned copy before splitting the name and the revision, the same way it is done in cmd_print.


michalvasko:
Thanks, but change the target branch to `devel`, please.

